PR #1, opened 2016-06-05
Merkle tree

c633:
Implement the Merkle prefix tree #2  and STR #3


merkletree.go
```diff
@@ -0,0 +1,275 @@
...
+	leftHash   []byte
+	rightHash  []byte
+	level      int
+}
+
+type userLeafNode struct {
+	interiorNode
+	key   string
+	value []byte
+	index []byte
```
masomel: The field for the cryptographic commitment ```commit(key || value)``` should still be included since it needs to be returned as part of the auth path.

operation.go
```diff
@@ -0,0 +1,6 @@
+package merkletree
+
+type Operation struct {
```
masomel: I'm thinking that ```Operation``` should maybe be moved outside of the merkle prefix tree entirely. Right now, it's not doing much since it's only a wrapper for a key-value pair, and all operations (registrations and changes) are being treated as inserts (which is fine). Since the CONIKS server will be handling registration and key change operations, and it will have to order these, it may make more sense to define ```Operation``` at the server level only?

c633: > Since the CONIKS server will be handling registration and key change operations, and it will have to order these, it may make more sense to define Operation at the server level only?

Yes. I already fixed this in my latest commit.

c633:
temporarily close this PR.

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+			}
+			position++
+		default:
+			return ErrInvalidTree
+		}
+	}
+	return nil
+}
+
+func (m *MerkleTree) RecomputeHash() {
```
masomel: Is this wrapper function really needed?

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+}
+
+func (m *MerkleTree) digest(input []byte) []byte {
+	h := m.hash.Hash
+	defer h.Reset()
+	h.Write(input)
+	return h.Sum(nil)
+}
+
+// Node manipulation functions
```
masomel: This is rather minor, but this should help make your code even more readable: There's some inconsistent naming in the node manipulation functions. For example, you have ```hashInteriorNode()``` but then only ```serialize()```.

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+	key        string
+	value      []byte
+	index      []byte
+	commitment []byte
+}
+
+type MerkleNode interface {
+	Value() []byte
+}
+
```
masomel: A function that should belong in this module is the a generator of auth paths for proofs of inclusion/absence these are general features of Merkle trees.

merkletree.go
c633: `hashInteriorNode()` is a function and `serialize()` is a method of `interiorNode` struct. So I would name these functions/methods as current naming. 
These may have `serialize()` method of `userLeafNode` struct in the future. Also I cannot turn the `hashxxx()` functions into methods since there is no `emptyNode` struct. 

Anw, I will try to have a better naming convention here after refactoring phase.

merkletree.go
c633: No, it will be removed in the refactoring phase.

kv.go
```diff
@@ -0,0 +1,111 @@
...
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+package merkletree
```
Liamsi: I think it would make sense to put this in a separate package (for instance db, storage, or kvstorage)

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+	input = append(input, node.leftHash...)
+	input = append(input, node.rightHash...)
+	return input
+}
+
+func leafNodeCommitment(m *MerkleTree, key string, value []byte) []byte {
+	commit := append([]byte{}, m.salt...)
+	commit = append(commit, key...)
+	commit = append(commit, value...)
+	return m.digest(commit)
```
Liamsi: That's not so important as it only could be a optimization: But is there a reason for not using the hash functions `Write` (and Sum at the end) instead of copying all bytes into a new byte array?

str.go
```diff
@@ -0,0 +1,87 @@
...
+	sig         []byte
+	prev        *SignedTreeRoot
+}
+
+func (m *MerkleTree) generateSTR(ep int64, prevEp int64, prevHash []byte) *SignedTreeRoot {
+	bytesPreSig := getSTRBytesForSig(m, ep, prevEp, prevHash)
+	sig := m.scheme.Sign(bytesPreSig)
+
+	return &SignedTreeRoot{
+		treeRoot:    &(*m.root),
```
Liamsi: Why do you dereference and immediately reference here again? Shouldn't that give you the same adress at the end? (like here: https://play.golang.org/p/wt-JIa-5w_)

str.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (m *MerkleTree) generateNextSTR(ep int64) *SignedTreeRoot {
+	currentSTR := getCurrentSTR()
+	prevEpoch := currentSTR.epoch
+	prevStrHash := m.digest(serializeSTR(*currentSTR))
+	bytesPreSig := getSTRBytesForSig(m, ep, prevEpoch, prevStrHash)
+
+	sig := m.scheme.Sign(bytesPreSig)
+	return &SignedTreeRoot{
+		treeRoot:    &(*m.root),
```
Liamsi: Same as https://github.com/coniks-sys/libmerkleprefixtree-go/pull/1/files#r66312105

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+	return nil
+}
+
+func (m *MerkleTree) digest(input []byte) []byte {
+	h := m.hash.Hash
+	defer h.Reset()
+	h.Write(input)
+	return h.Sum(nil)
+}
```
Liamsi: related to https://github.com/coniks-sys/libmerkleprefixtree-go/pull/1/files#r66309558:
you could change this to something similar to: https://play.golang.org/p/IxBaAVdi35

str.go
c633: Thanks! My bad :(

history.go
```diff
@@ -0,0 +1,59 @@
+package merkletree
```
masomel: Signed tree roots and the history should probably have their own package.

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+		index:      index,
+		commitment: leafNodeCommitment(m, key, value),
+	}
+
+	return m.insertNode(index, &toAdd)
+}
+
+// Private Index calculation function
+// would be replaced with Ismail's VRF implementation
+func (m *MerkleTree) computePrivateIndex(key string) []byte {
```
masomel: I would put this in a separate package for all things crypto

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+				append(prefixBits, true), currentNodeI.rightChild)
+		}
+		return hashInteriorNode(m, currentNodeI)
+	case *userLeafNode:
+		return hashLeafNode(m, node.(*userLeafNode))
+	}
+	return nil
+}
+
+func (m *MerkleTree) digest(ms ...[]byte) []byte {
```
masomel: Same as above, I would put this in a separate crypto package

str.go
```diff
@@ -0,0 +1,91 @@
+package merkletree
```
masomel: Same as with the history, I would put this in a separate package together with the history

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+type Scheme interface {
+	Sign([]byte) []byte
+	Verify(publicKey []byte, message, sig []byte) bool
+}
+
+type HashFunction struct {
+	HashSizeByte int
+	Hash         hash.Hash
+	HashId       crypto.Hash
+}
```
masomel: I would put ```Scheme``` and ```HashFunc``` in a separate module for all things crypto.

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+
+type userLeafNode struct {
+	interiorNode
+	key        string
+	value      []byte
+	index      []byte
+	commitment []byte
+}
+
```
masomel: I think it may make sense to put all node-related definitions and functions in their own separate node module.

merkletree.go
masomel: I see your reasoning with the method vs function naming conventions, but turning functions into methods or vice versa to have consistent naming isn't what I was thinking. I made of comment on this in a later commit, I think what would solve this may be refactoring the node types into their own module(s) for clarity. 

I think given the requirements for auth path generation, it would make sense to create an ```emptyNode`` type.

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+func (node *interiorNode) serialize() []byte {
+	input := make([]byte, 0, len(node.leftHash)+len(node.rightHash))
+	input = append(input, node.leftHash...)
+	input = append(input, node.rightHash...)
+	return input
+}
+
+func leafNodeCommitment(m *MerkleTree, key string, value []byte) []byte {
```
masomel: There's a ```userLeafNode``` type, so this should really be declared as a method on that type. Then, once there are one or more node modules, this method should be moved there.

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+	return m.digest(
+		[]byte(m.salt),
+		[]byte(key),
+		[]byte(value))
+}
+
+func hashInteriorNode(m *MerkleTree, node *interiorNode) []byte {
```
masomel: Same as with ```leafNodeCommitment```.

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+	return m.digest(node.serialize())
+}
+
+func hashLeafNode(m *MerkleTree, node *userLeafNode) []byte {
```
masomel: Same as with ```leafNodeCommitment```.

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+	return m.digest(
+		[]byte{LeafIdentifier},         // K_leaf
+		[]byte(m.treeNonce),            // K_n
+		[]byte(node.index),             // i
+		[]byte(intToBytes(node.level)), // l
+		[]byte(node.commitment),        // commit(key|| value)
+	)
+}
+
+func hashEmptyNode(m *MerkleTree, prefixBits []bool) []byte {
```
masomel: As mentioned earlier, if you add an ```emptyNode``` type to support auth paths, this should probably be declared as separate ```serialize()``` and ```hash()``` methods on that type.

merkletree.go
```diff
@@ -0,0 +1,344 @@
...
+	return m.digest(
+		[]byte{EmptyBranchIdentifier},       // K_empty
+		[]byte(m.treeNonce),                 // K_n
+		[]byte(toBytes(prefixBits)),         // i
+		[]byte(intToBytes(len(prefixBits))), // l
+	)
+}
+
+// tree clone methods
```
masomel: Move these ```clone()``` methods to a node module or their corresponding type modules.

str.go
```diff
@@ -0,0 +1,91 @@
...
+	strBytes = append(strBytes, str.treeRoot.serialize()...)   // root
+	strBytes = append(strBytes, longToBytes(str.epoch)...)     // epoch
+	strBytes = append(strBytes, longToBytes(str.prevEpoch)...) // previous epoch
+	strBytes = append(strBytes, str.prevStrHash...)            // previous hash
+	strBytes = append(strBytes, str.sig...)                    // signature
+
+	return strBytes
+}
+
+func serializePolicies(m *MerkleTree) []byte {
```
masomel: Since this function exists, you can probably add a policy field to the STR struct.

.travis.yml
```diff
@@ -0,0 +1,15 @@
...
+        only:
+                - merkle-tree
+
+notifications:
+  # Email notifications are disabled to not annoy anybody.
+  # See http://about.travis-ci.org/docs/user/build-configuration/ to learn more
+  # about configuring notification recipients and more.
+  email: false
+
+    Status API Training Shop Blog About
```
arlolra: ?

c633:
I decided to remove db module for the sake of simplicity. Thus, tree/STR storage will be implemented outside this library. All look-up operations now are in-memory only.

.travis.yml
c633: cc @masomel

merkletree.go
```diff
@@ -0,0 +1,242 @@
...
+	if nodePointer == nil {
+		return nil, nil, ErrInvalidTree
+	}
+	switch nodePointer.(type) {
+	case *userLeafNode:
+		if nodePointer.(*userLeafNode).key == key {
+			proof = append(proof, nodePointer.(*userLeafNode))
+			return nodePointer.(*userLeafNode), proof, nil
+		} else {
+			// TODO: what should we return here?
```
c633: @masomel : in case the lookup prefix is same as the prefix of an index (i.e. an user leaf node) in the tree, but they are actually different, what should we return for proof of absence?

.travis.yml
masomel: That's probably a copy-paste error. It didn't affect the Travis build, though...

merkletree.go
masomel: @c633 What do you mean by "they are actually different"?

masomel:
> I decided to remove db module for the sake of simplicity. Thus, tree/STR storage will be implemented outside this library. All look-up operations now are in-memory only. 

This sounds good for now. As I mentioned in my email, I still think we will want to add hooks for a storage backend later to improve scalability. I'll add this as an issue.

merkletree.go
masomel: If we get to this else case, it would mean that the exact index isn't in the tree. I think the right thing to do here would be to return the existing leaf node as a proof of absence, but without the key and value to preserve privacy. Since the hash of the leaf node doesn't include the ley and value directly, it should still be able to verify the auth path.

merkletree.go
```diff
@@ -228,117 +215,28 @@ func (m *MerkleTree) hashNode(prefixBits []bool, node interface{}) []byte {
...
-func (node *interiorNode) serialize() []byte {
-	input := make([]byte, 0, len(node.leftHash)+len(node.rightHash))
-	input = append(input, node.leftHash...)
-	input = append(input, node.rightHash...)
-	return input
-}
-
-func leafNodeCommitment(m *MerkleTree, key string, value []byte) []byte {
-	return m.digest(
+func commit(m *MerkleTree, key string, value []byte) []byte {
```
masomel: I would move ```commit()``` to the crypto module for consistency.

merkletree.go
```diff
@@ -80,12 +93,13 @@ func lookUp(key string, str *SignedTreeRoot) (MerkleNode, []MerkleNode, error) {
...
+		proof = append(proof, nodePointer.(*userLeafNode))
 		if nodePointer.(*userLeafNode).key == key {
-			proof = append(proof, nodePointer.(*userLeafNode))
 			return nodePointer.(*userLeafNode), proof, nil
-		} else {
-			// TODO: what should we return here?
 		}
+		// reached a different leaf with a matching prefix
+		// return nil and a auth path including the leaf node
+		return nil, proof, nil
```
masomel: Since the proof includes the different leaf, the verifier could still see all the contents of the leaf, including the key and the value, right? Or am I missing something? Shouldn't there be a step here that "removes" the ```key``` and the ```value``` from the leaf in the proof to preserve privacy?

masomel:
Looking good! The only major missing piece is #5 and making sure proofs of absence are privacy-preserving.

merkletree.go
c633: The verifier could see only exported contents through `LookUpProofNode` interface (https://github.com/coniks-sys/libmerkleprefixtree-go/blob/0cc51a16ed2f2a23023043ee3b72e7861f861657/node.go#L30). Currently, the exported content is the hash value of the node. I'm not sure what else should be exported but it could be extended in the future.

masomel: Ah I see now, ok. Thanks for clarifying! I think all fields included in the hash value of the leaf should be exported to allow the verifier to recompute the hash value. But I agree, this extension can be done in the future.

masomel:
Looks like @Liamsi will create a separate PR for addressing #5, so unless others still have comments, I think this PR is ready to be closed :)

.travis.yml
masomel: @c633 Thanks for fixing this!

c633:
@arlolra : do you have any comments on this PR?

arlolra:
I will look at it today.

Liamsi:
A rather general comment/question not necessarily directly related to this PR: Wouldn't it make sense to further split this in sub-packages instead of files only? I could think of the following sketched directory structure:
```
libmerkleprefixtree-go/
├── ...
├── crypto/
│   ├── crypto.go
│   ├── crypto.go
│   ├── extra25519/
│   ├── vrf/
│   ├── vrf.go
│   ├── vrf_test.go
│   └── vrf_test.go
└── internal/
      └── util.go
```
The `...` part contains all files not explicitly listed. The internal package contains all functions needed in several sub-packages but which we wouldn't explicitly want to export (because they aren't part of the API or because thy may change in the future). Find more details on internal packages [here](https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit). Maybe that's an overkill but as we will add some more files it could make sense. What do you think?

util.go
```diff
@@ -0,0 +1,47 @@
...
+	for i := 0; i < len(bits); i++ {
+		if bits[i] {
+			bs[i/8] |= (1 << 7) >> uint(i%8)
+		}
+	}
+	return bs
+}
+
+// copy from dename authors
+func ToBits(bs []byte) []bool {
```
Liamsi: Function currently doesn't seem to be used.

crypto.go
```diff
@@ -0,0 +1,39 @@
...
+	"golang.org/x/crypto/ed25519"
+	"golang.org/x/crypto/sha3"
+)
+
+const (
+	HashSizeByte   = 32
+	HashID         = "SHAKE128"
+	PrivateKeySize = ed25519.PrivateKeySize
+	PublicKeySize  = ed25519.PublicKeySize
+	SignatureSize  = ed25519.SignatureSize
```
Liamsi: These 3 constants currently don't seem to be used.

masomel:
I agree, and don't think it would be overkill. I think sub-packages would definitely help further organize our code and API.

c633:
I also agree. Thanks for pointing out internal packages.

crypto/crypto.go
```diff
@@ -0,0 +1,36 @@
...
+	"golang.org/x/crypto/sha3"
+)
+
+const (
+	HashSizeByte = 32
+	HashID       = "SHAKE128"
+)
+
+func Digest(ms ...[]byte) []byte {
+	h := sha3.NewShake256()
```
arlolra: Shouldn't this be `sha3.NewShake128()`, so that it uses exactly what it says on the tin? Doesn't seem like it affects the security either way, just the amount of internal state consumed.

crypto/crypto.go
```diff
@@ -0,0 +1,36 @@
...
+	for _, m := range ms {
+		h.Write(m)
+	}
+	var ret [HashSizeByte]byte
+	h.Read(ret[:])
+	return ret[:]
```
arlolra: I think this is clearer as,

	ret := make([]byte, HashSizeByte)
	h.Read(ret)
	return ret

.travis.yml
```diff
@@ -2,15 +2,12 @@ sudo: false
 
 language: go
 
-go:
-        - 1.4
-        - 1.5
-        - 1.6
+branches:
+        only:
+                - merkle-tree
```
arlolra: Why only this branch?

crypto/crypto.go
```diff
@@ -0,0 +1,36 @@
...
+func Verify(publicKey []byte, message, sig []byte) bool {
+	return ed25519.Verify(publicKey, message, sig)
+}
+
+func Commitment(salt []byte, key string, value []byte) []byte {
+	return Digest(
+		[]byte(salt),
+		[]byte(key),
+		[]byte(value))
+}
```
arlolra: This seems redundant and only has one use site. I'd remove it.

crypto/crypto.go
```diff
@@ -11,11 +11,11 @@ const (
 )
 
 func Digest(ms ...[]byte) []byte {
-	h := sha3.NewShake256()
+	h := sha3.NewShake128()
```
arlolra: You need to update the tests too

crypto/crypto.go
```diff
@@ -11,11 +11,11 @@ const (
...
 	for _, m := range ms {
 		h.Write(m)
 	}
-	var ret [HashSizeByte]byte
+	ret := make([]byte, HashSizeByte)
 	h.Read(ret[:])
```
arlolra: Why the slices?

`ret[:]` can just be `ret` in both these cases.

crypto/crypto.go
c633: I didn't get your meaning here. Can you make it clearer, please?

arlolra: I was suggesting you remove `Commitment` altogether, and just do `Digest(salt, []byte(key), value)` in the one place it's used.

.travis.yml
masomel: Travis was failing on master since it doesn't have any go files yet.

node.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+type interiorNode struct {
+	parent     *interiorNode
+	leftChild  MerkleNode
+	rightChild MerkleNode
+	leftHash   []byte
+	rightHash  []byte
+	level      int
+}
```
arlolra: Making `interiorNode` the base is kind of confusing in that leafs and empty nodes don't have children, right?

I'd suggest making a type

    type Node struct {
      level int
      parent *node
    }

that they all share.  Or, even making `emptyNode` the base, but then taking care to always use that as the default in your switches.

node.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+	return node.parent.rightHash
+}
+
+func (node *emptyNode) GetHash() []byte {
+	if node.parent.leftChild == node {
+		return node.parent.leftHash
+	}
+	return node.parent.rightHash
+}
```
arlolra: These three are the same and should probably be a method of the base struct.

history.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/coniks-sys/libmerkleprefixtree-go/crypto"
+)
+
+var (
+	ErrHistoryExisted = errors.New("[merkletree] trying to override existing history")
+	ErrBadEpoch       = errors.New("[merkletree] next epoch's STR has bad epoch ")
+)
+
+var currentSTR *SignedTreeRoot
```
arlolra: It would be nice to eliminate this global and create a `History` struct, like

    type History struct {
       m *MerkleTree
       currentSTR *SignedTreeRoot
       epochInterval int64
    } 

then, instead of `InitHistory`, you can have,

    func NewHistory(m *MerkleTree, startEp, epInterval int64) *History {
       h := new(History)
       h.m = m
       h.epochInterval = epInterval
       h.currentSTR = m.generateSTR(startEp, 0, make([]byte, crypto.HashSizeByte))
       return h
    }

something like that. Then the rest of these methods can have the signature `func (h *History)`. How does that sound?

str.go
```diff
@@ -0,0 +1,78 @@
...
+		policies:    m.policies.Serialize(),
+		prev:        nil,
+	}
+}
+
+func (m *MerkleTree) generateNextSTR(ep int64) *SignedTreeRoot {
+	currentSTR := getCurrentSTR()
+	prevEpoch := currentSTR.epoch
+	prevStrHash := crypto.Digest(serializeSTR(*currentSTR))
+	bytesPreSig := getSTRBytesForSig(m, ep, prevEpoch, prevStrHash)
```
arlolra: From this point on, it looks like the same thing as `generateSTR`. Can it use that?

merkletree.go
```diff
@@ -0,0 +1,270 @@
...
+	str := getCurrentSTR()
+	return lookUp(key, str)
+}
+
+func LookUpInEpoch(key string, ep int64) (MerkleNode, []LookUpProofNode, error) {
+	str := GetSTR(ep)
+	return lookUp(key, str)
+}
+
+func lookUp(key string, str *SignedTreeRoot) (MerkleNode, []LookUpProofNode, error) {
```
arlolra: It seems like `lookUp` should be a method of a `MerkleTree`, no? The only thing you use from the `str` is the `treeRoot`.

Then, `LookUp` and `LookUpInEpoch` would be methods of the new `History` struct, which stores the cloned `MerkleTree`s.

history.go
c633: Yes. It sounds good. I will fix all these issues in next commit

c633:
I fixed those issues based on @arlolra's comments in 92f23eaf94d939cfd603d8431ef861f8d0c38931 and 6a0db276b2fa8e417f8d83a8d676dea76cd012e4

internal/util.go
```diff
@@ -0,0 +1,38 @@
...
+	return buf
+}
+
+func IntToBytes(num int) []byte {
+	buf := make([]byte, 4)
+	binary.LittleEndian.PutUint32(buf, uint32(num))
+	return buf
+}
+
+// In each byte, the bits are ordered MSB to LSB
```
arlolra: This is copied from coname and should be attributed.
https://github.com/yahoo/coname/blob/master/merkle.go#L151

Does it mean we need to distribute their license?

Also, why not use the converse `ToBits` as well?
(the answer to this may become clear as I continue to review)

masomel:
> This is copied from coname and should be attributed.
...
Does it mean we need to distribute their license?

Totally agree. I'm not sure we need to distribute their license directly. What I've done (and also seen in other big projects) is to include a statement similar to this:

"This license applies to all parts of [our project] except the following:
- The [copied] library, located in [path in our project tree]
This library is copyrighted by [library authors]. See license: [license location]."

merkletree_test.go
```diff
@@ -0,0 +1,312 @@
...
+
+var treeNonce = []byte("TREE NONCE")
+var salt = []byte("salt")
+
+var (
+	pk, sk []byte
+)
+
+func init() {
+	pk, sk, _ = ed25519.GenerateKey(rand.Reader)
```
arlolra: Does this belong in the crypto package? `crypto.GenerateKey` or some such.

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+
+func (h *History) LookUp(key string) (MerkleNode, []LookUpProofNode, error) {
+	str := h.currentSTR
+	return lookUp(key, str)
+}
+
+func (h *History) LookUpInEpoch(key string, ep int64) (MerkleNode, []LookUpProofNode, error) {
+	str := h.GetSTR(ep)
+	return lookUp(key, str)
+}
```
arlolra: These two methods belong in history.go

merkletree_test.go
```diff
@@ -0,0 +1,312 @@
...
+}
+
+func TestOneEntry(t *testing.T) {
+	m := InitMerkleTree(&DefaultPolicies{}, treeNonce, salt, pk, sk)
+
+	history := NewHistory(m, 1, 1)
```
arlolra: It'd be nice to be able to test MerkleTree on its own in merkletree_test.go
and then have separate tests for History in history_test.go

merkletree_test.go
masomel: ```init()``` is unused. I think it can be removed.

Liamsi: `init` autom. get called before everything else in this file: https://golang.org/doc/effective_go.html#init
But I also think it couldn't harm to move the key-generation to specific tests instead of in an init function.

arlolra: I think `init` is called by default by golang test suite.  The variable it's initializing `pk` and `sk` are used below at least.

The reason I'm asking this to be moved to the crypto pkg is because we are otherwise abstracting away the choices of crypto primitives.

masomel: Oh I see. I'm still learning Go, sorry.

internal/util.go
arlolra: > Would something like this work? Either way, I would place any internal code that's been copied from coname in its own file with the appropriate copyright notice at the top.

Yup, that seems fine to me.

README.md
```diff
@@ -1,2 +1,43 @@
...
 A Merkle prefix tree implementation in Golang
+
+This library currently uses the `SHAKE128` ShakeHash with output size of 32 bytes. 
+
+The signature scheme is `Ed25519` signature algorithm.
+
+### Usage
+Initiate the tree & history
+```
+m := merkletree.InitMerkleTree(nonce, salt, publicKey, privateKey)
```
arlolra: This would benefit from seeing `public, private, _ = crypto.GenerateKey()` and describing `salt` and `nonce`.

crypto/crypto.go
```diff
@@ -0,0 +1,29 @@
...
+	return ret
+}
+
+func Sign(privateKey []byte, message []byte) []byte {
+	return ed25519.Sign(privateKey, message)
+}
+
+func Verify(publicKey []byte, message, sig []byte) bool {
+	return ed25519.Verify(publicKey, message, sig)
+}
```
arlolra: Once you've added a `GenerateKey`, and you also add a crypto_test.go and exercise all these methods.

.travis.yml
arlolra: Travis has a `Build only if .travis.yml is present` setting, but I see you've already pushed one to master. No bother.

str.go
```diff
@@ -0,0 +1,69 @@
...
+		epoch:       ep,
+		prevEpoch:   prevEp,
+		prevStrHash: prevHash,
+		sig:         sig,
+		policies:    m.policies.Serialize(),
+		prev:        nil,
+	}
+}
+
+func (m *MerkleTree) generateNextSTR(cur *SignedTreeRoot, ep int64) *SignedTreeRoot {
```
arlolra: This look like it should be a method on a `SignedTreeRoot`.

str.go
```diff
@@ -0,0 +1,69 @@
...
+	treeRoot    *interiorNode
+	epoch       int64
+	prevEpoch   int64
+	prevStrHash []byte
+	sig         []byte
+	policies    []byte
+	prev        *SignedTreeRoot
+}
+
+func (m *MerkleTree) generateSTR(ep int64, prevEp int64, prevHash []byte) *SignedTreeRoot {
```
arlolra: This probably should be a method of a `MerkleTree`, rather a constructor like `NewSTR`.

merkletree_test.go
arlolra: > But I also think it couldn't harm to move the key-generation to specific tests instead of in an init function.

Yes, key generation needs its own test in crypto_test.go, but the need for initialization of a public/private keypair here points to the fact that the Merkle tree constructor probably shouldn't need those, and indeed doesn't really use them. I'll note that above.

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+const (
+	EmptyBranchIdentifier = 'E'
+	LeafIdentifier        = 'L'
+)
+
+type MerkleTree struct {
+	treeNonce []byte
+	salt      []byte
+	privKey   []byte
+	pubKey    []byte
```
arlolra: The `privKey` and `pubKey` aren't used in the Merkle tree construction. They probably belong closer to the STR. I'd remove them from here.

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+	return nil, nil, ErrInvalidTree
+}
+
+func (m *MerkleTree) Set(key string, value []byte) error {
+	index := computePrivateIndex(key)
+	toAdd := userLeafNode{
+		key:        key,
+		value:      value,
+		index:      index,
+		commitment: crypto.Digest(m.salt, []byte(key), value),
```
arlolra: Can `treeNonce` be used here instead of `salt`?

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+		return nil, proof, nil
+	}
+	return nil, nil, ErrInvalidTree
+}
+
+func (m *MerkleTree) Set(key string, value []byte) error {
+	index := computePrivateIndex(key)
+	toAdd := userLeafNode{
+		key:        key,
+		value:      value,
```
arlolra: Why are `key` and `value` stored here? I thought the whole point of the commitment was to preserve the privacy of this information.

masomel: The server does need to store them in order to provide the opening to the commitment when a client is verifying an authentication path (proof of inclusion). For example, if Alice's client needs to verify that Alice's key hasn't changed unexpectedly, it needs the opening to the commitment. A proof of absence doesn't include ```key``` and ```value``` for privacy reasons.

arlolra: The intro to §3 says,

> To make the directory privacy-preserving, CONIKS employs two cryptographic primitives. First, a private index is computed for each username via a verifiable unpredictable function. Each user’s keys are stored at the associated private index rather than his or her username (or a hash of it). This prevents the data structure from leaking information about usernames.

which sounds like `key` should not be stored here.

> Second, to ensure that it is not possible to test if a users’ key data is equal to some known value even given this user’s lookup index, a cryptographic commitment to each user’s key data is stored at the private index, rather than the public keys themselves.

which sounds like `value` should not be stored here.

What am I missing?

I guess you're referring to §3.2,

> To prove inclusion of the full binding, the server provides an opening of the commitment in addition to the authentication path.

masomel: Right, I'm referring to §3.2.

arlolra: Ok, but what am I missing in the above quotes?

masomel: You're not missing anything in those quotes. I think the confusion stems from what ```key``` and ```value``` represent. ```key``` should be the private index and ```value``` should contain the commitment.

arlolra: But clearly `index` on L124 contains the index and `commitment` contains the commitment. `key` is the username and `value` is the user's public key, as far as I can tell. Hence my original comment above.

arlolra: So was I right in saying that `key` / `value` should be removed?

masomel: Yeah, I realized this when I went back to check the code. 

I should say that isn't really our main goal to protect the user's privacy against the key server, but only against other malicious parties who wish to obtain information about users by querying the key directory. (We do have a protocol for also keeping the ```value``` private from the server, though it's not the default, so I'm not sure if it should be built into the tree design). So the main argument for keeping the ```key``` and the ```value``` is to support lookups. A client looking up Alice's name for the first time should be able to obtain Alice's key from the authentication path (and will need it to verify the commitment). The server may also need access to all ```key``s in the cases in which it changes its VRF key to re-randomize the order of the leaves (this is discussed briefly in §6 of the paper, though we may not need to worry about it right now if we're not supporting this feature).

masomel: I should add, actually, that the ```value``` should really only appear as a blob of data that the client knows how to interpret based on a ```allowsPublicVisibility``` flag (discussed in §4.3.1).

arlolra: Ok, so this sentence hinges on `allowsPublicVisibility === false`,

> Second, to ensure that it is not possible to test if a users’ key data is equal to some known value even given this user’s lookup index, a cryptographic commitment to each user’s key data is stored at the private index, rather than the public keys themselves.

Thanks for clearing all this up for me. I'll move on in my review now.

To sum up, I think we should drop `key`.

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+	return m.insertNode(index, &toAdd)
+}
+
+// Private Index calculation function
+// would be replaced with Ismail's VRF implementation
+func computePrivateIndex(key string) []byte {
+	return crypto.Digest([]byte(key))
+}
+
+func (m *MerkleTree) insertNode(key []byte, toAdd *userLeafNode) error {
```
arlolra: Please be consistent and call this `index` instead of `key`.

merkletree.go
masomel: No problem!

merkletree.go
arlolra: Also, why even pass `key` here if it's part of `toAdd`?

internal/util.go
arlolra: > Also, why not use the converse ToBits as well?
> (the answer to this may become clear as I continue to review)

Ok, so the answer is that coname computes the entire path ahead of time with `ToBits`, but you're doing in on demand with `GetNthBit`. That's fine.

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+			parent: nil,
+			level:  0,
+		},
+		leftChild:  leftBranch,
+		rightChild: rightBranch,
+		leftHash:   nil,
+		rightHash:  nil,
+	}
+	leftBranch.parent = root
+	rightBranch.parent = root
```
arlolra: This whole pattern of creating an `interiorNode` is repeated in `insertNode`.  It could benefit from a `NewInteriorNode` constructor that takes a `parent` and `level`.

c633:
Thanks for all comments so far. I will fix them today.

crypto/crypto.go
c633: Actually it just a wrapper for `ed25519` official pkg, so I think we don't need any tests for our `crypto` pkg, do we?

merkletree.go
c633: I should use a consistent naming here. The `key` parameter is a byte array and the `key` of `toAdd` is string.

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+		toAdd.level++
+		switch nodePointer.(type) {
+		case *userLeafNode:
+			// reached a "bottom" of the tree.
+			// add a new interior node and push the previous leaf down
+			// then continue insertion
+			currentNodeUL := nodePointer.(*userLeafNode)
+			if currentNodeUL.parent == nil {
+				return ErrInvalidTree
+			}
```
arlolra: This assertion is suspect. It seems like this is more than an error if it were true, but a corruption of the tree itself. Do we expect to recover from it? I'd prefer to see stronger guarantees where nodes are created if this is a real concern.

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+					parent: currentNodeUL.parent,
+					level:  currentNodeUL.level,
+				},
+				leftChild:  leftBranch,
+				rightChild: rightBranch,
+			}
+			leftBranch.parent = &newInteriorNode
+			rightBranch.parent = &newInteriorNode
+
+			if currentNodeUL.key == toAdd.key {
```
arlolra: It seems like comparing indexes would be preferable.

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+				// replace the value
+				currentNodeUL.value = toAdd.value
+				return nil
+			}
+
+			currentNodeKey := computePrivateIndex(currentNodeUL.key)
```
arlolra: Is this expected to be different from `currentNodeKey.index` already?

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+			}
+			currentNodeUL.level++
+			currentNodeUL.parent = &newInteriorNode
+			if newInteriorNode.parent.(*interiorNode).leftChild == nodePointer {
+				newInteriorNode.parent.(*interiorNode).leftChild = &newInteriorNode
+			} else {
+				newInteriorNode.parent.(*interiorNode).rightChild = &newInteriorNode
+			}
+			nodePointer = &newInteriorNode
+			toAdd.level--
```
arlolra: I think this is wrong. `toAdd.level` should be the same as the new `currentNodeUL.level`, and decrementing it here makes it the same as `position`.

To make this all clearer, please remove all these increment operators.

Let's change `position` to `depth`, and then for all the places set `.level = depth + 1`. This also deserves, so please add one.

crypto/crypto.go
arlolra: Right, I'm not asking you to provide test vectors for ed25519.

I'd just like to see a test that generates a keypair, then uses it sign and verify something. Just something that exercises the abstract interface so that swapping out the internals is straightforward.

merkletree.go
arlolra: > I should use a consistent naming here. The key parameter is a byte array and the key of toAdd is string. 

Yes, the different types are the result of them not being the same data.

>  I made a separate insertNode since I thought we might have some other methods such as update or things like that in the future. I think it's still fine now.

Fair enough. That's fine.

internal/util.go
```diff
@@ -0,0 +1,38 @@
...
+func ToBytes(bits []bool) []byte {
+	bs := make([]byte, (len(bits)+7)/8)
+	for i := 0; i < len(bits); i++ {
+		if bits[i] {
+			bs[i/8] |= (1 << 7) >> uint(i%8)
+		}
+	}
+	return bs
+}
```
arlolra: A small util_test.go file would be appreciated.

merkletree.go
arlolra: Oh, sorry, nope I was confused. You're decrementing because you don't break here and instead go around the loop again the with interior node. Sorry.

But I do think the above suggestions make things clearer.

internal/util.go
c633: Yup, I did write some tests for those methods but after all those tests passed, I deleted them all. My bad :(

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+	str := h.currentSTR
+	return lookUp(key, str)
+}
+
+func (h *History) LookUpInEpoch(key string, ep int64) (MerkleNode, []LookUpProofNode, error) {
+	str := h.GetSTR(ep)
+	return lookUp(key, str)
+}
+
+func lookUp(key string, str *SignedTreeRoot) (MerkleNode, []LookUpProofNode, error) {
```
arlolra: This should be the converse of `Set`,

    func (m *MerkleTree) Get(key string) MerkleNode

If you want to keep an internal `lookupNode`, similar to `insertNode`, that's fine.

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+			nodePointer = nodePointer.(*interiorNode).rightChild
+		} else {
+			nodePointer = nodePointer.(*interiorNode).leftChild
+		}
+		position++
+	}
+
+	if nodePointer == nil {
+		return nil, nil, ErrInvalidTree
+	}
```
arlolra: This seems more like a case to panic than an error. We should always reach a leaf (empty or otherwise).

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+	case *userLeafNode:
+		proof = append(proof, nodePointer.(*userLeafNode))
+		if nodePointer.(*userLeafNode).key == key {
+			return nodePointer.(*userLeafNode), proof, nil
+		}
+		// reached a different leaf with a matching prefix
+		// return nil and a auth path including the leaf node
+		return nil, proof, nil
+	case *emptyNode:
+		proof = append(proof, nodePointer.(*emptyNode))
```
arlolra: Proof seems conflated in this method.

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+					currentNodeI.leftChild = toAdd
+					toAdd.parent = currentNodeI
+					break insertLoop
+				} else {
+					nodePointer = currentNodeI.leftChild
+				}
+			}
+			position++
+		default:
+			return ErrInvalidTree
```
arlolra: Should probably just panic here and stop returning errors altogether in this method.

merkletree.go
masomel: I'm also not sure how real of a concern this is. It's true that if this assertion is true, the tree has somehow become corrupted, and I agree that it's not the most graceful solution to return an error in this case. But I really know what a reasonable recovery state would be; I suppose you could roll back the tree to the last-known valid version, but I feel like this could somehow be exploited to DoS the server. It seems to me that in practice, a solution would be to recover from a persistent data store or a replica, and I'm thinking it's not unreasonable to leave this decision up to the user.

masomel: This being said, I think there are a few basic checks that should be added to ensure the tree isn't entirely malformed, e.g. checking that the root is of type ```interiorNode``` before entering ```insertLoop```.

merkletree.go
masomel: I think at this point it is since the check for equality happens in the if statement above. But as you said in your previous note, the check in that if statement should compare indices, not the keys.

merkletree.go
```diff
@@ -0,0 +1,274 @@
...
+				},
+				leftChild:  leftBranch,
+				rightChild: rightBranch,
+			}
+			leftBranch.parent = &newInteriorNode
+			rightBranch.parent = &newInteriorNode
+
+			if currentNodeUL.key == toAdd.key {
+				// replace the value
+				currentNodeUL.value = toAdd.value
```
masomel: Shouldn't the commitment be replaced at this point as well?

merkletree.go
masomel: It might make more sense to call ```proof``` something like "path" or "auth path".

merkletree.go
arlolra: Failure to meet those assertions means there's a problem with our implementation and that the program should panic. `Get` and `Set` should not be returning errors.

merkletree.go
arlolra: But we don't need to return the path, right. Since we're returning a node, you can walk it back from the `parent` property.

merkletree.go
masomel: Whether we return an error or a panic, the user still has to make a decision about how to handle these situations.

arlolra: No, I think that if they hit a panic, they need to file a bug in our issue tracker. That's the distinction I'm trying to make.

merkletree.go
masomel: Ah yes, that's true.

merkletree.go
arlolra: I mean, I think an error should be reserved for places where you think it's reasonable to expect a failure to occur, like making a request across a network.

Here, we're dealing with an in memory store where an error should never be reached (as far as I can tell). Any error means a problem with our logic in the implementation.

Is that fair?

masomel: I see. Though isn't is possible for the tree to become corrupted due to external factors like an attacker who has compromised the server? I guess the fact that this attack is possible would still be considered a bug...

masomel: > Any error means a problem with our logic in the implementation.

I'm not sure this will be true in 100% of the cases. But I see your point about errors vs bugs.

merkletree.go
c633: > Is this expected to be different from currentNodeUL.index already?

No, it should be the same. I removed this redundant code.
cc @masomel

merkletree.go
c633: > But we don't need to return the path, right. Since we're returning a node, you can walk it back from the parent property.
Yes. But the `MerkleNode` exposes some methods that are not expected in auth path. The `MerkleNode` is currently used to read the value of the node return by `Get` whereas the `LookUpProofNode` (I renamed to `ProofNode` in the latest commit) is expected to return some fields (e.g. the hash of the node currently) to preserve the privacy.

arlolra: I think the idea was that `Get` would return a `MerkleNode`.  And then you'd have another function,

```
func GenerateProof(mn MerkleNode) []ProofNode {
  var proof []ProofNode
  for mn.parent !== nil {
    proof = preprend(...)
    mn = mn.parent
  }
  return proof
}
```

Does that work?

merkletree.go
masomel: I'm still not sure about this. I'll consult with Joe on this one.

.travis.yml
masomel: Yeah, I realized that I probably shouldn't have pushed it to master first. Oh well.

merkletree.go
arlolra: I did this in ced1af3a049fa25b82d7131b53a3fe0e5a51dd2d

merkletree.go
arlolra: I cleaned up a bit of the confusion about this in 65012f7a30bcf2dd375a12871730831a5d4aaac0

merkletree.go
masomel: I think it makes sense to separate the ```Get``` from the proof generation logic, especially if the lookup returns an ```EmptyNode``` whenever the lookup failed, as long as there is still a lookup function that returns the proof node. As we're using the Merkle tree as an authenticated dictionary, the lookups should always return the path along with the value [[1 ] Section 3](http://dl.acm.org/citation.cfm?doid=2019599.2019602).

merkletree_test.go
arlolra: I made this final separation in f127162919a4db8b72e27e310335f3268f224052

arlolra:
Ok, I'm going to merge some of this now.

arlolra:
Let's see if 14956a902257ac75facb7080a8ea48f44d41ce89 works first though :)

c633:
This is the travis log of the previous check: https://s3.amazonaws.com/archive.travis-ci.org/jobs/139557242/log.txt (before 14956a902257ac75facb7080a8ea48f44d41ce89)

Actually Travis did run all subpackage tests

```
=== RUN TestHistoryHashChain-2
--- PASS: TestHistoryHashChain-2 (0.01s)
=== RUN TestOneEntry-2
--- PASS: TestOneEntry-2 (0.00s)
=== RUN TestTwoEntries-2
--- PASS: TestTwoEntries-2 (0.00s)
=== RUN TestThreeEntries-2
--- PASS: TestThreeEntries-2 (0.00s)
=== RUN TestInsertExistedKey-2
--- PASS: TestInsertExistedKey-2 (0.00s)
=== RUN TestTreeClone-2
--- PASS: TestTreeClone-2 (0.00s)
PASS
ok  	github.com/coniks-sys/libmerkleprefixtree-go	0.022s
=== RUN TestSignVerify-2
--- PASS: TestSignVerify-2 (0.00s)
PASS
ok  	github.com/coniks-sys/libmerkleprefixtree-go/crypto	0.017s
=== RUN TestBitsBytesConvert-2
--- PASS: TestBitsBytesConvert-2 (0.00s)
PASS
ok  	github.com/coniks-sys/libmerkleprefixtree-go/internal	0.002s
```

So I guess you want to run subpackage tests first?

arlolra:
Oh, no, I just didn't realize the default for go was `go test -v ./...`

https://docs.travis-ci.com/user/languages/go#Default-Test-Script

My bad.  Sorry for the noise.

merkletree.go
masomel: @jcb82 confirmed my suspicions:
> ... you want a different nonce per-user so that you can selectively reveal, the tree nonce is for the whole tree.

It looks to me that ```salt``` is also a tree-wide value. This needs to be changed.

arlolra:
Ok, most of this is merged in 57c98319a33afc32cc67f6ac494b8ce2dc2615d5.

I'll open up a pull with the rest for review.

arlolra:
Ok, #8 is open for business.  Let's talk about the `salt` there.

Issue #4, opened 2016-06-12
Add hooks for storage backend


masomel:
Right now, trees are kept in memory, and we're assuming that the user will set up her own storage backend and save the tree. To improve scalability, we should provide hooks for a few popular storage backend. We may also want to provide a mechanism for serializing trees into files on disk for this purpose.

c633:
As stated in Marcela's email:
> the persistent storage has multiple purposes. For reliability in case of a server failure. For scalability, which solves the issue of multiple large trees not fitting into memory (eventually, I assume the server will be memory bound), and allows for multiple distributed servers as well. And possibly even for security. 

My thought on this issue is I would like to implement a db interface (like the `kv` interface of `coname`) for supporting key-value backends (e.g., `leveldb`), as I did before. All the old trees would be stored in the db. There also has some predefined parameters (maybe from the tree construction arguments), e.g., the number of tree would be stored in the db, presumably older trees could be safely removed. Then we would provide a mechanism for looking up in these old trees by reconstructing the trees from db. This way, it also raises another question: how well it will scale if we store the tree entirely to the db? 

What do you think? @masomel @arlolra @Liamsi

masomel:
> My thought on this issue is I would like to implement a db interface (like the kv interface of coname) for supporting key-value backends (e.g., leveldb), as I did before. 

I think this could work. Then the idea would be that the user would implement the interface for their specific key-value backend? Doesn't this conflict with the notion that we would provide a mechanism for reconstructing the tree from the KV store? The reason being that in order to reconstruct the tree from a KV store you would have to make certain assumptions about how the tree data is stored in it, or am I missing something?

> This way, it also raises another question: how well it will scale if we store the tree entirely to the db? 

You could consider storing the diffs of trees only to save space and not save redundant copies of data that hasn't changed across epochs. The reconstruction might be a bit trickier then because you have to reassign the right node pointers.

Another option that may save more space would be to only store the leaf nodes (and you could even go as far as only storing the changed leaf nodes). But then you have to reconstruct the entire tree from the leaves doing a bunch of insertions. The question then becomes: is it more efficient to do many pointer reassignments or many insert operations?

c633:
[...]

PR #8, opened 2016-06-22
STR for review

arlolra:
This is split off from #1, still needing review by all.

/cc @c633 @masomel @Liamsi

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,30 @@ This library currently uses the `SHAKE128` ShakeHash with output size of 32 byte
 The signature scheme is `Ed25519` signature algorithm.
 
 ### Usage
-Initiate the tree
+Initiate the history hash chain (the persistent authenticated dictionary)
 ```
-// using DefaultPolicies as current policy of the tree
-m := merkletree.InitMerkleTree(&DefaultPolicies{}, "nonce", "salt")
-// insert existing data (e.g., from db)
-m.Set(key, value)
+// generate private key for STR signing
+signKey := crypto.GenerateKey()
+
+// init STR history chain with maximum length is len
+// using DefaultPolicies as current policy
+pad := NewPAD(NewPolicies(epochInterval), signKey, len)
+```
+
+Update tree in each epoch
+```
+// insert new data
+pad.Set(key, value)
 ...
+// update STR history chain
+// pass nil if the policies doesn't change
+pad.Update(nil)
+```
+
+Look-up
+
+`LookUp(key)` and `LookUpInEpoch(key, counter)` return a `MerkleNode` instance and an `AuthenticationPath` for proofs of inclusion/absence.
+A proof of absence also includes an empty leaf node in the returned auth path.
 
 ### TODO
 Some methods/functions should be exported in the future when the library is being used in real applications.
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ signKey := crypto.GenerateKey()
 
 // init STR history chain with maximum length is len
 // using DefaultPolicies as current policy
-pad := NewPAD(NewPolicies(epochInterval), signKey, len)
+pad := NewPAD(NewPolicies(epochDeadline), signKey, len)
 ```
 
 Update tree in each epoch
@@ -30,7 +30,7 @@ pad.Update(nil)
 
 Look-up
 
-`LookUp(key)` and `LookUpInEpoch(key, counter)` return a `MerkleNode` instance and an `AuthenticationPath` for proofs of inclusion/absence.
+`LookUp(key)` and `LookUpInEpoch(key, epoch)` return a `MerkleNode` instance and an `AuthenticationPath` for proofs of inclusion/absence.
 A proof of absence also includes an empty leaf node in the returned auth path.
 
 ### TODO
--- a/crypto/crypto.go
+++ b/crypto/crypto.go
@@ -8,14 +8,12 @@ import (
 )
 
 const (
-	HashSizeByte = 32
-	HashID       = "SHAKE128"
+	HashSizeByte   = 32
+	HashID         = "SHAKE128"
+	PrivateKeySize = 64
 )
 
-type KeyPair struct {
-	PrivateKey []byte
-	PublicKey  []byte
-}
+type SigningKey ed25519.PrivateKey
 
 func Digest(ms ...[]byte) []byte {
 	h := sha3.NewShake128()
@@ -27,18 +25,19 @@ func Digest(ms ...[]byte) []byte {
 	return ret
 }
 
-func GenerateKey() KeyPair {
-	pk, sk, _ := ed25519.GenerateKey(rand.Reader)
-	return KeyPair{
-		PrivateKey: sk,
-		PublicKey:  pk,
+func GenerateKey() (SigningKey, error) {
+	_, sk, err := ed25519.GenerateKey(rand.Reader)
+	return SigningKey(sk), err
+}
+
+func Sign(key SigningKey, message []byte) []byte {
+	return ed25519.Sign(ed25519.PrivateKey(key), message)
+}
+
+func Verify(key SigningKey, message, sig []byte) bool {
+	pk, ok := ed25519.PrivateKey(key).Public().(ed25519.PublicKey)
+	if !ok {
+		return false
 	}
-}
-
-func Sign(key KeyPair, message []byte) []byte {
-	return ed25519.Sign(key.PrivateKey, message)
-}
-
-func Verify(key KeyPair, message, sig []byte) bool {
-	return ed25519.Verify(key.PublicKey, message, sig)
+	return ed25519.Verify(pk, message, sig)
 }
--- a/crypto/crypto_test.go
+++ b/crypto/crypto_test.go
@@ -4,7 +4,10 @@ import "testing"
 
 // copied from official crypto.ed25519 tests
 func TestSignVerify(t *testing.T) {
-	key := GenerateKey()
+	key, err := GenerateKey()
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	message := []byte("test message")
 	sig := Sign(key, message)
--- a/merkletree.go
+++ b/merkletree.go
@@ -37,7 +37,7 @@ func NewMerkleTree() (*MerkleTree, error) {
 	return m, nil
 }
 
-func (m *MerkleTree) Get(key string) (MerkleNode, *AuthenticationPath) {
+func (m *MerkleTree) Get(key string) *AuthenticationPath {
 	lookupIndex := computePrivateIndex(key)
 	lookupIndexBits := util.ToBits(lookupIndex)
 	depth := 0
@@ -76,20 +76,18 @@ func (m *MerkleTree) Get(key string) (MerkleNode, *AuthenticationPath) {
 	}
 	switch nodePointer.(type) {
 	case *userLeafNode:
-		authPath.index = nodePointer.(*userLeafNode).index
-		authPath.level = nodePointer.(*userLeafNode).level
-		authPath.leaf = nodePointer.(*userLeafNode)
+		pNode := nodePointer.(*userLeafNode).Clone(nil).(*userLeafNode)
+		authPath.leaf = pNode
 		if bytes.Equal(nodePointer.(*userLeafNode).index, lookupIndex) {
-			return nodePointer.(*userLeafNode), authPath
+			return authPath
 		}
 		// reached a different leaf with a matching prefix
-		// return nil and a auth path including the leaf node
-		return nil, authPath
+		// return a auth path including the leaf node
+		pNode.value = nil
+		return authPath
 	case *emptyNode:
-		authPath.index = nodePointer.(*emptyNode).index
-		authPath.level = nodePointer.(*emptyNode).level
-		authPath.leaf = nodePointer.(*emptyNode)
-		return nil, authPath
+		authPath.leaf = nodePointer.(*emptyNode).Clone(nil).(*emptyNode)
+		return authPath
 	}
 	panic(ErrorInvalidTree)
 }
--- a/merkletree.go
+++ b/merkletree.go
@@ -10,7 +10,7 @@ import (
 )
 
 var (
-	ErrInvalidTree = errors.New("[merkletree] invalid tree")
+	ErrorInvalidTree = errors.New("[merkletree] invalid tree")
 )
 
 const (
@@ -19,28 +19,35 @@ const (
 )
 
 type MerkleTree struct {
-	treeNonce []byte
-	policies  Policies
-	root      *interiorNode
+	nonce []byte
+	root  *interiorNode
+	hash  []byte
 }
 
-func InitMerkleTree(policies Policies, treeNonce []byte) *MerkleTree {
-	root := NewInteriorNode(nil, 0)
-
-	m := &MerkleTree{
-		treeNonce: treeNonce,
-		policies:  policies,
-		root:      root,
+func NewMerkleTree() (*MerkleTree, error) {
+	root := NewInteriorNode(nil, 0, []bool{})
+	nonce := make([]byte, crypto.HashSizeByte)
+	if _, err := rand.Read(nonce); err != nil {
+		return nil, err
 	}
-	return m
+	m := &MerkleTree{
+		nonce: nonce,
+		root:  root,
+	}
+	return m, nil
 }
 
-func (m *MerkleTree) Get(key string) (MerkleNode, []ProofNode) {
+func (m *MerkleTree) Get(key string) (MerkleNode, *AuthenticationPath) {
 	lookupIndex := computePrivateIndex(key)
+	lookupIndexBits := util.ToBits(lookupIndex)
 	depth := 0
 	var nodePointer interface{}
 	nodePointer = m.root
-	var proof []ProofNode
+
+	authPath := &AuthenticationPath{
+		treeNonce:   m.nonce,
+		lookupIndex: lookupIndex,
+	}
 
 	for {
 		if _, ok := nodePointer.(*userLeafNode); ok {
@@ -51,33 +58,40 @@ func (m *MerkleTree) Get(key string) (MerkleNode, []ProofNode) {
 			// reached to an empty branch
 			break
 		}
-		proof = append(proof, nodePointer.(*interiorNode))
-		direction := util.GetNthBit(lookupIndex, depth)
+		direction := lookupIndexBits[depth]
 		if direction {
+			authPath.prunedHashes = append(authPath.prunedHashes,
+				nodePointer.(*interiorNode).leftHash)
 			nodePointer = nodePointer.(*interiorNode).rightChild
 		} else {
+			authPath.prunedHashes = append(authPath.prunedHashes,
+				nodePointer.(*interiorNode).rightHash)
 			nodePointer = nodePointer.(*interiorNode).leftChild
 		}
 		depth++
 	}
 
 	if nodePointer == nil {
-		panic(ErrInvalidTree)
+		panic(ErrorInvalidTree)
 	}
 	switch nodePointer.(type) {
 	case *userLeafNode:
-		proof = append(proof, nodePointer.(*userLeafNode))
-		if nodePointer.(*userLeafNode).key == key {
-			return nodePointer.(*userLeafNode), proof
+		authPath.index = nodePointer.(*userLeafNode).index
+		authPath.level = nodePointer.(*userLeafNode).level
+		authPath.leaf = nodePointer.(*userLeafNode)
+		if bytes.Equal(nodePointer.(*userLeafNode).index, lookupIndex) {
+			return nodePointer.(*userLeafNode), authPath
 		}
 		// reached a different leaf with a matching prefix
 		// return nil and a auth path including the leaf node
-		return nil, proof
+		return nil, authPath
 	case *emptyNode:
-		proof = append(proof, nodePointer.(*emptyNode))
-		return nil, proof
+		authPath.index = nodePointer.(*emptyNode).index
+		authPath.level = nodePointer.(*emptyNode).level
+		authPath.leaf = nodePointer.(*emptyNode)
+		return nil, authPath
 	}
-	panic(ErrInvalidTree)
+	panic(ErrorInvalidTree)
 }
 
 func (m *MerkleTree) Set(key string, value []byte) error {
@@ -91,7 +105,7 @@ func (m *MerkleTree) Set(key string, value []byte) error {
 
 	toAdd := userLeafNode{
 		key:        key,
-		value:      value,
+		value:      append([]byte{}, value...), // make a copy of value
 		index:      index,
 		salt:       salt,
 		commitment: crypto.Digest(salt, []byte(key), value),
@@ -107,7 +121,8 @@ func computePrivateIndex(key string) []byte {
 	return crypto.Digest([]byte(key))
 }
 
-func (m *MerkleTree) insertNode(key []byte, toAdd *userLeafNode) {
+func (m *MerkleTree) insertNode(index []byte, toAdd *userLeafNode) {
+	indexBits := util.ToBits(index)
 	depth := 0
 	var nodePointer interface{}
 	nodePointer = m.root
@@ -121,7 +136,7 @@ insertLoop:
 			// then continue insertion
 			currentNodeUL := nodePointer.(*userLeafNode)
 			if currentNodeUL.parent == nil {
-				panic(ErrInvalidTree)
+				panic(ErrorInvalidTree)
 			}
 
 			if bytes.Equal(currentNodeUL.index, toAdd.index) {
@@ -132,7 +147,7 @@ insertLoop:
 				return
 			}
 
-			newInteriorNode := NewInteriorNode(currentNodeUL.parent, depth)
+			newInteriorNode := NewInteriorNode(currentNodeUL.parent, depth, indexBits[:depth])
 
 			direction := util.GetNthBit(currentNodeUL.index, depth)
 			if direction {
@@ -150,9 +165,9 @@ insertLoop:
 			nodePointer = newInteriorNode
 		case *interiorNode:
 			currentNodeI := nodePointer.(*interiorNode)
-			direction := util.GetNthBit(key, depth)
-
+			direction := indexBits[depth]
 			if direction { // go right
+				currentNodeI.rightHash = nil
 				if currentNodeI.rightChild.isEmpty() {
 					currentNodeI.rightChild = toAdd
 					toAdd.level = depth + 1
@@ -162,6 +177,7 @@ insertLoop:
 					nodePointer = currentNodeI.rightChild
 				}
 			} else { // go left
+				currentNodeI.leftHash = nil
 				if currentNodeI.leftChild.isEmpty() {
 					currentNodeI.leftChild = toAdd
 					toAdd.level = depth + 1
@@ -173,16 +189,22 @@ insertLoop:
 			}
 			depth += 1
 		default:
-			panic(ErrInvalidTree)
+			panic(ErrorInvalidTree)
 		}
 	}
-	return
+}
+
+func (m *MerkleTree) recomputeHash() {
+	m.hash = m.root.Hash(m)
+}
+
+func (m *MerkleTree) GetHash() []byte {
+	return m.hash
 }
 
 func (m *MerkleTree) Clone() *MerkleTree {
 	return &MerkleTree{
-		treeNonce: m.treeNonce,
-		policies:  m.policies,
-		root:      m.root.clone(nil),
+		nonce: m.nonce,
+		root:  m.root.Clone(nil).(*interiorNode),
 	}
 }
--- a/merkletree_test.go
+++ b/merkletree_test.go
@@ -41,19 +41,19 @@ func TestOneEntry(t *testing.T) {
 			"get", m.root.rightHash)
 	}
 
-	r, _ := m.Get(key)
-	if r == nil {
+	r := m.Get(key)
+	if r.Leaf().Value() == nil {
 		t.Error("Cannot find value of key:", key)
 		return
 	}
-	v := r.Value()
+	v := r.Leaf().Value()
 	if !bytes.Equal(v, val) {
 		t.Errorf("Value mismatch %v / %v", v, val)
 	}
 
 	// Check leaf node hash
 	h.Reset()
-	h.Write(r.(*userLeafNode).salt)
+	h.Write(r.Leaf().(*userLeafNode).salt)
 	h.Write([]byte(key))
 	h.Write(val)
 	h.Read(commit[:])
@@ -72,8 +72,8 @@ func TestOneEntry(t *testing.T) {
 			"get", m.root.leftHash)
 	}
 
-	r, _ = m.Get("abc")
-	if r != nil {
+	r = m.Get("abc")
+	if r.Leaf().Value() != nil {
 		t.Error("Invalid look-up operation:", key)
 		return
 	}
@@ -97,22 +97,22 @@ func TestTwoEntries(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	n1, _ := m.Get(key1)
-	if n1 == nil {
+	ap1 := m.Get(key1)
+	if ap1.Leaf().Value() == nil {
 		t.Error("Cannot find key:", key1)
 		return
 	}
 
-	n2, _ := m.Get(key2)
-	if n2 == nil {
+	ap2 := m.Get(key2)
+	if ap2.Leaf().Value() == nil {
 		t.Error("Cannot find key:", key2)
 		return
 	}
 
-	if !bytes.Equal(n1.Value(), []byte("value1")) {
+	if !bytes.Equal(ap1.Leaf().Value(), []byte("value1")) {
 		t.Error(key1, "value mismatch")
 	}
-	if !bytes.Equal(n2.Value(), []byte("value2")) {
+	if !bytes.Equal(ap2.Leaf().Value(), []byte("value2")) {
 		t.Error(key2, "value mismatch")
 	}
 }
@@ -140,66 +140,43 @@ func TestThreeEntries(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	n1, _ := m.Get(key1)
-	if n1 == nil {
+	ap1 := m.Get(key1)
+	if ap1.Leaf().Value() == nil {
 		t.Error("Cannot find key:", key1)
 		return
 	}
-	n2, _ := m.Get(key2)
-	if n2 == nil {
+	ap2 := m.Get(key2)
+	if ap2.Leaf().Value() == nil {
 		t.Error("Cannot find key:", key2)
 		return
 	}
-	n3, _ := m.Get(key3)
-	if n3 == nil {
+	ap3 := m.Get(key3)
+	if ap3.Leaf().Value() == nil {
 		t.Error("Cannot find key:", key3)
 		return
 	}
 
-	// since the first bit of n2 index is false and the one of n1 & n3 are true
-	if reflect.ValueOf(m.root.leftChild).Pointer() !=
-		reflect.ValueOf(n2.(*userLeafNode)).Pointer() {
-		t.Error("Malformed tree insertion")
-	}
-	if n2.(*userLeafNode).level != 1 {
+	// since the first bit of ap2 index is false and the one of ap1 & ap3 are true
+	if ap2.Leaf().Level() != 1 {
 		t.Error("Malformed tree insertion")
 	}
 
 	// since n1 and n3 share first 2 bits
-	if n1.(*userLeafNode).level != 3 {
+	if ap1.Leaf().Level() != 3 {
 		t.Error("Malformed tree insertion")
 	}
-	if n3.(*userLeafNode).level != 3 {
-		t.Error("Malformed tree insertion")
-	}
-	// n1 and n3 should have same parent
-	if reflect.ValueOf(n1.(*userLeafNode).parent).Pointer() !=
-		reflect.ValueOf(n3.(*userLeafNode).parent).Pointer() {
-		t.Error("Malformed tree insertion")
-	}
-	if reflect.ValueOf(n1.(*userLeafNode).parent.(*interiorNode).leftChild).Pointer() !=
-		reflect.ValueOf(n3).Pointer() {
-		t.Error("Malformed tree insertion")
-	}
-	if reflect.ValueOf(n3.(*userLeafNode).parent.(*interiorNode).rightChild).Pointer() !=
-		reflect.ValueOf(n1).Pointer() {
+	if ap3.Leaf().Level() != 3 {
 		t.Error("Malformed tree insertion")
 	}
 
-	if !bytes.Equal(n1.Value(), []byte("value1")) {
+	if !bytes.Equal(ap1.Leaf().Value(), []byte("value1")) {
 		t.Error(key1, "value mismatch")
 	}
-	if !bytes.Equal(n2.Value(), []byte("value2")) {
+	if !bytes.Equal(ap2.Leaf().Value(), []byte("value2")) {
 		t.Error(key2, "value mismatch")
 	}
-	if !bytes.Equal(n3.Value(), []byte("value3")) {
+	if !bytes.Equal(ap3.Leaf().Value(), []byte("value3")) {
 		t.Error(key3, "value mismatch")
-	}
-
-	// check index of empty node on the tree
-	n := n3.(*userLeafNode).parent.(*interiorNode).parent.(*interiorNode).rightChild.(*emptyNode)
-	if !bytes.Equal(n.index, []byte{192}) {
-		t.Error("Malformed tree insertion")
 	}
 }
 
@@ -221,18 +198,18 @@ func TestInsertExistedKey(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	val, _ := m.Get(key1)
-	if val == nil {
+	ap := m.Get(key1)
+	if ap.Leaf().Value() == nil {
 		t.Error("Cannot find key:", key1)
 		return
 	}
 
-	if !bytes.Equal(val.Value(), []byte("new value")) {
+	if !bytes.Equal(ap.Leaf().Value(), []byte("new value")) {
 		t.Error(key1, "value mismatch\n")
 	}
 
-	if !bytes.Equal(val.Value(), val2) {
-		t.Errorf("Value mismatch %v / %v", val.Value(), val2)
+	if !bytes.Equal(ap.Leaf().Value(), val2) {
+		t.Errorf("Value mismatch %v / %v", ap.Leaf().Value(), val2)
 	}
 
 	val3 := []byte("new value 2")
@@ -240,14 +217,14 @@ func TestInsertExistedKey(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	val, _ = m.Get(key1)
-	if val == nil {
+	ap = m.Get(key1)
+	if ap.Leaf().Value() == nil {
 		t.Error("Cannot find key:", key1)
 		return
 	}
 
-	if !bytes.Equal(val.Value(), val3) {
-		t.Errorf("Value mismatch %v / %v", val.Value(), val3)
+	if !bytes.Equal(ap.Leaf().Value(), val3) {
+		t.Errorf("Value mismatch %v / %v", ap.Leaf().Value(), val3)
 	}
 }
 
@@ -285,21 +262,21 @@ func TestTreeClone(t *testing.T) {
 	}
 
 	// lookup
-	r, _ := m2.Get(key1)
-	if r == nil {
+	ap := m2.Get(key1)
+	if ap.Leaf().Value() == nil {
 		t.Error("Cannot find key:", key1)
 		return
 	}
-	if !bytes.Equal(r.Value(), []byte("value1")) {
+	if !bytes.Equal(ap.Leaf().Value(), []byte("value1")) {
 		t.Error(key1, "value mismatch\n")
 	}
 
-	r, _ = m2.Get(key2)
-	if r == nil {
+	ap = m2.Get(key2)
+	if ap.Leaf().Value() == nil {
 		t.Error("Cannot find key:", key2)
 		return
 	}
-	if !bytes.Equal(r.Value(), []byte("value2")) {
+	if !bytes.Equal(ap.Leaf().Value(), []byte("value2")) {
 		t.Error(key2, "value mismatch\n")
 	}
 }
--- a/merkletree_test.go
+++ b/merkletree_test.go
@@ -4,12 +4,86 @@ import (
 	"bytes"
 	"reflect"
 	"testing"
+
+	"github.com/coniks-sys/libmerkleprefixtree-go/internal"
+	"golang.org/x/crypto/sha3"
 )
 
-var treeNonce = []byte("TREE NONCE")
+func TestOneEntry(t *testing.T) {
+	m, err := NewMerkleTree()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var commit [32]byte
+	var expect [32]byte
+
+	key := "key"
+	val := []byte("value")
+
+	if err := m.Set(key, val); err != nil {
+		t.Fatal(err)
+	}
+	m.recomputeHash()
+
+	index := computePrivateIndex(key)
+
+	// Check empty node hash
+	h := sha3.NewShake128()
+	h.Write([]byte{EmptyBranchIdentifier})
+	h.Write(m.nonce)
+	h.Write(util.ToBytes([]bool{true}))
+	h.Write(util.IntToBytes(1))
+	h.Read(expect[:])
+	if !bytes.Equal(m.root.rightHash, expect[:]) {
+		t.Error("Wrong righ hash!",
+			"expected", expect,
+			"get", m.root.rightHash)
+	}
+
+	r, _ := m.Get(key)
+	if r == nil {
+		t.Error("Cannot find value of key:", key)
+		return
+	}
+	v := r.Value()
+	if !bytes.Equal(v, val) {
+		t.Errorf("Value mismatch %v / %v", v, val)
+	}
+
+	// Check leaf node hash
+	h.Reset()
+	h.Write(r.(*userLeafNode).salt)
+	h.Write([]byte(key))
+	h.Write(val)
+	h.Read(commit[:])
+
+	h.Reset()
+	h.Write([]byte{LeafIdentifier})
+	h.Write(m.nonce)
+	h.Write(index)
+	h.Write(util.IntToBytes(1))
+	h.Write(commit[:])
+	h.Read(expect[:])
+
+	if !bytes.Equal(m.root.leftHash, expect[:]) {
+		t.Error("Wrong left hash!",
+			"expected", expect,
+			"get", m.root.leftHash)
+	}
+
+	r, _ = m.Get("abc")
+	if r != nil {
+		t.Error("Invalid look-up operation:", key)
+		return
+	}
+}
 
 func TestTwoEntries(t *testing.T) {
-	m := InitMerkleTree(&DefaultPolicies{}, treeNonce)
+	m, err := NewMerkleTree()
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	key1 := "key1"
 	val1 := []byte("value1")
@@ -44,7 +118,10 @@ func TestTwoEntries(t *testing.T) {
 }
 
 func TestThreeEntries(t *testing.T) {
-	m := InitMerkleTree(&DefaultPolicies{}, treeNonce)
+	m, err := NewMerkleTree()
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	key1 := "key1"
 	val1 := []byte("value1")
@@ -118,10 +195,19 @@ func TestThreeEntries(t *testing.T) {
 	if !bytes.Equal(n3.Value(), []byte("value3")) {
 		t.Error(key3, "value mismatch")
 	}
+
+	// check index of empty node on the tree
+	n := n3.(*userLeafNode).parent.(*interiorNode).parent.(*interiorNode).rightChild.(*emptyNode)
+	if !bytes.Equal(n.index, []byte{192}) {
+		t.Error("Malformed tree insertion")
+	}
 }
 
 func TestInsertExistedKey(t *testing.T) {
-	m := InitMerkleTree(&DefaultPolicies{}, treeNonce)
+	m, err := NewMerkleTree()
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	key1 := "key"
 	val1 := append([]byte(nil), "value"...)
@@ -144,6 +230,25 @@ func TestInsertExistedKey(t *testing.T) {
 	if !bytes.Equal(val.Value(), []byte("new value")) {
 		t.Error(key1, "value mismatch\n")
 	}
+
+	if !bytes.Equal(val.Value(), val2) {
+		t.Errorf("Value mismatch %v / %v", val.Value(), val2)
+	}
+
+	val3 := []byte("new value 2")
+	if err := m.Set(key1, val3); err != nil {
+		t.Fatal(err)
+	}
+
+	val, _ = m.Get(key1)
+	if val == nil {
+		t.Error("Cannot find key:", key1)
+		return
+	}
+
+	if !bytes.Equal(val.Value(), val3) {
+		t.Errorf("Value mismatch %v / %v", val.Value(), val3)
+	}
 }
 
 func TestTreeClone(t *testing.T) {
@@ -152,16 +257,31 @@ func TestTreeClone(t *testing.T) {
 	key2 := "key2"
 	val2 := []byte("value2")
 
-	m1 := InitMerkleTree(&DefaultPolicies{}, treeNonce)
+	m1, err := NewMerkleTree()
+	if err != nil {
+		t.Fatal(err)
+	}
 	if err := m1.Set(key1, val1); err != nil {
 		t.Fatal(err)
 	}
+	m1.recomputeHash()
 
 	// clone new tree and insert new value
 	m2 := m1.Clone()
 
 	if err := m2.Set(key2, val2); err != nil {
 		t.Fatal(err)
+	}
+	m2.recomputeHash()
+
+	// tree hash
+	// right branch hash value is still the same
+	if bytes.Equal(m1.root.leftHash, m2.root.leftHash) {
+		t.Fatal("Bad clone")
+	}
+	if reflect.ValueOf(m1.root.leftHash).Pointer() == reflect.ValueOf(m2.root.leftHash).Pointer() ||
+		reflect.ValueOf(m1.root.rightHash).Pointer() == reflect.ValueOf(m2.root.rightHash).Pointer() {
+		t.Fatal("Bad clone")
 	}
 
 	// lookup
--- a/node.go
+++ b/node.go
@@ -1,5 +1,10 @@
 package merkletree
 
+import (
+	"github.com/coniks-sys/libmerkleprefixtree-go/crypto"
+	"github.com/coniks-sys/libmerkleprefixtree-go/internal"
+)
+
 type node struct {
 	parent MerkleNode
 	level  int
@@ -9,6 +14,8 @@ type interiorNode struct {
 	node
 	leftChild  MerkleNode
 	rightChild MerkleNode
+	leftHash   []byte
+	rightHash  []byte
 }
 
 type userLeafNode struct {
@@ -22,19 +29,26 @@ type userLeafNode struct {
 
 type emptyNode struct {
 	node
+	index []byte
 }
 
-func NewInteriorNode(parent MerkleNode, level int) *interiorNode {
+func NewInteriorNode(parent MerkleNode, level int, prefixBits []bool) *interiorNode {
+	prefixLeft := append([]bool(nil), prefixBits...)
+	prefixLeft = append(prefixLeft, false)
+	prefixRight := append([]bool(nil), prefixBits...)
+	prefixRight = append(prefixRight, true)
 	leftBranch := &emptyNode{
 		node: node{
 			level: level + 1,
 		},
+		index: util.ToBytes(prefixLeft),
 	}
 
 	rightBranch := &emptyNode{
 		node: node{
 			level: level + 1,
 		},
+		index: util.ToBytes(prefixRight),
 	}
 	newNode := &interiorNode{
 		node: node{
@@ -43,6 +57,8 @@ func NewInteriorNode(parent MerkleNode, level int) *interiorNode {
 		},
 		leftChild:  leftBranch,
 		rightChild: rightBranch,
+		leftHash:   nil,
+		rightHash:  nil,
 	}
 	leftBranch.parent = newNode
 	rightBranch.parent = newNode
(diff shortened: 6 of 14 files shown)


merkletree.go
```diff
@@ -83,11 +83,17 @@ func (m *MerkleTree) Get(key string) (MerkleNode, []ProofNode) {
 
 func (m *MerkleTree) Set(key string, value []byte) {
 	index := computePrivateIndex(key)
+
+	// generate random salt per user
+	salt := make([]byte, crypto.MaxSaltSize)
+	_, _ = io.ReadFull(rand.Reader, salt)
```
Liamsi: I'm not sure what could go wrong here, but I would still read and catch the error here (*or* check if all bytes were read).
And a matter of style, but this could be changed to:
```
salt := make([]byte, c)
_, err := rand.Read(b)
```
`Read` uses  `ReadFull` internally: https://golang.org/src/crypto/rand/rand.go?s=714:752#L11

Liamsi: Actually, in very rare cases an error could occur if there isn't enough entropy (at least under linux): https://github.com/golang/go/blob/master/src/internal/syscall/unix/getrandom_linux.go#L27

arlolra: Yeah, this seems like a legitimate reason for returning an error from `Set`.

merkletree.go
```diff
@@ -123,7 +130,7 @@ insertLoop:
 				return
```
arlolra: Need to update the `salt` here too. Maybe it'd be worth swapping in the whole `toAdd`, and only updating the `parent` and `level`, to avoid not missing any fields in the future.

masomel: Yup, replacing the node with ```toAdd`` makes sense to me.

c633: fixed in https://github.com/coniks-sys/libmerkleprefixtree-go/blob/30123ed70d6ca2c67ff07e005a939fe92357f225/merkletree.go#L133

node.go
```diff
@@ -87,12 +97,69 @@ func (node *emptyNode) isEmpty() bool {
 
 var _ ProofNode = (*node)(nil)
 
+func (node *node) GetHash() []byte {
```
masomel: Why is ```node``` implementing the ```ProofNode``` interface? Shouldn't ```interiorNode```, ```emptyNode``` and ```userLeafNode``` be implementing it?

node.go
```diff
@@ -87,12 +97,69 @@ func (node *emptyNode) isEmpty() bool {
...
+		}
+	default:
+		return nil
+	}
+
+	return parent.rightHash
+}
+
+func (node *interiorNode) GetHash() []byte {
+	return node.hash()
```
masomel: If ```interiorNode``` is implementing ```ProofNode```, I think it should have the functionality of ```node```'s ```GetHash```, no? After all, the ```interiorNode```s in the proof should only expose the hash of the pruned subtree.

node.go
```diff
@@ -87,12 +97,69 @@ func (node *emptyNode) isEmpty() bool {
...
 
+func (node *node) GetHash() []byte {
+	parent, ok := node.parent.(*interiorNode)
+	if !ok {
+		return nil
+	}
+	switch parent.leftChild.(type) {
+	case *emptyNode:
+		if &parent.leftChild.(*emptyNode).node == node {
+			return parent.leftHash
```
masomel: The ```ProofNode``` should only expose the hashes of the pruned subtrees. This seems to be doing the opposite. I believe this should return ```parent.rightHash``` if the node on the path is the ```leftChild```. The same goes for the ```userLeafNode``` case.

node.go
```diff
@@ -87,12 +97,69 @@ func (node *emptyNode) isEmpty() bool {
...
+		}
+	case *userLeafNode:
+		if &parent.leftChild.(*userLeafNode).node == node {
+			return parent.leftHash
+		}
+	default:
+		return nil
+	}
+
+	return parent.rightHash
```
masomel: Don't forget to change this to ```leftHash``` if you're changing the return values in the switch statement.

node.go
```diff
@@ -55,6 +64,7 @@ type MerkleNode interface {
 }
 
 type ProofNode interface {
```
masomel: Users need to recompute the hash of the proof node, so they should need access to all the data included in the respective leaf hashes such as the ```commitment```, the ```treeNonce``` etc.

crypto/crypto.go
```diff
@@ -27,18 +26,23 @@ func Digest(ms ...[]byte) []byte {
...
 }
 
-func GenerateKey() KeyPair {
-	pk, sk, _ := ed25519.GenerateKey(rand.Reader)
-	return KeyPair{
-		PrivateKey: sk,
-		PublicKey:  pk,
-	}
+func GenerateKey() SigningKey {
+	_, sk, _ := ed25519.GenerateKey(rand.Reader)
```
Liamsi: Similar to some prev. comment, I think it makes sense that we return/handle the error here (similarly `io.ReadFull` is called on `crypto.rand` internally and in rare cases, e.g. not enough entropy available, the key generation could fail)

proof_test.go
```diff
@@ -0,0 +1,120 @@
...
+			[]byte{EmptyBranchIdentifier},       // K_empty
+			[]byte(ap.TreeNonce()),              // K_n
+			[]byte(ap.Index()),                  // i
+			[]byte(util.IntToBytes(ap.Level())), // l
+		)
+	}
+	return
+}
+
+func computeProofHash(ap *AuthenticationPath, depth int) (leftHash, rightHash []byte) {
```
masomel: This seems to be recomputing the proof hash from the root to the leaf (as depth is increasing in the recursion), but it should be from the leaf up to the root. This way, the client can compare the recomputed root from the auth path with the root in the STR for the same epoch.

c633: the flow of this function is mostly the same as `RecomputeHash()` method of `MerkleTree`. It returns the hash of left and right sub tree which then are compared to the value of `leftHash` and `rightHash` of the `root` node. (https://github.com/coniks-sys/libmerkleprefixtree-go/blob/1a48b4084a8821d21c25510c6192e97dd2083dd8/proof_test.go#L94)

history.go
```diff
@@ -65,3 +65,11 @@ func (h *History) NextEpoch() int64 {
 func (h *History) MerkleTree() *MerkleTree {
 	return h.tree
 }
+
+func (h *History) NewTemporaryBinding(index, value []byte) []byte {
```
masomel: I'm thinking it might be better to have a separate TB struct, because the server needs to keep track of all temporary bindings issued during a given epoch, and return a TB whenever a client makes a lookup for a name which may not be in the tree yet but which has been registered/changed. 

I think the functionality of keeping track can probably be implemented in the server itself, but I think it would facilitate handling TBs if they had their own struct.

proof_test.go
masomel: Ah, my bad for not noticing that other compare. Wouldn't it be more convenient to recompute the hash one more level up and just compare the root hash?

c633: > Wouldn't it be more convenient to recompute the hash one more level up and just compare the root hash?

Yes, I also think about it. It looks like the `MerkleTree` struct should have another field namely `treeHash` which is the hash of the root. What do you think?

masomel: > the flow of this function is mostly the same as ``RecomputeHash()`` method of ``MerkleTree``.

I thought it would have been more intuitive to actually start the function at the deepest level since you are recomputing the hashes from the leaf tree up. It wasn't immediately clear that this is what ```computeProofHash()``` is doing (unlike in ```RecomputeHash()``` it's immediately clear how you're traversing the tree). But I like your implementation, too.

masomel: > Yes, I also think about it. It looks like the ``MerkleTree`` struct should have another field namely ``treeHash`` which is the hash of the root. What do you think?

I think an additional field in ``MerkleTree`` to hold the root hash is a good idea.

proof_test.go
```diff
@@ -52,6 +52,11 @@ func computeProofHash(ap *AuthenticationPath, depth int) (leftHash, rightHash []
 	return
 }
 
+func authPathHash(ap *AuthenticationPath) []byte {
```
masomel: +1

crypto/crypto.go
```diff
@@ -27,18 +26,23 @@ func Digest(ms ...[]byte) []byte {
...
+	return ed25519.PrivateKey(s).Public()
 }
 
-func Verify(key KeyPair, message, sig []byte) bool {
-	return ed25519.Verify(key.PublicKey, message, sig)
+func Sign(key SigningKey, message []byte) []byte {
+	return ed25519.Sign(ed25519.PrivateKey(key), message)
+}
+
+func Verify(key SigningKey, message, sig []byte) bool {
```
arlolra: I would understand the need for the `public()` method if you were passing in a `crypto.PublicKey` here, but otherwise it's probably unnecessary.  The internals of `Verify()` are already implementation specific.

Also, why the local typedef for `SigningKey` but the imported `crypto.PublicKey`? I imagine disambiguating `crypto` in other parts of the codebase (should it be used) may be cumbersome, so probably best to avoid that if possible.

crypto/crypto.go
```diff
@@ -1,21 +1,20 @@
...
 	"golang.org/x/crypto/ed25519"
 	"golang.org/x/crypto/sha3"
 )
 
 const (
-	HashSizeByte = 32
-	HashID       = "SHAKE128"
+	HashSizeByte   = 32
+	HashID         = "SHAKE128"
+	PrivateKeySize = 64
```
arlolra: `PrivateKeySize` isn't being used yet, is it?

crypto/crypto_test.go
```diff
@@ -4,7 +4,7 @@ import "testing"
 
 // copied from official crypto.ed25519 tests
 func TestSignVerify(t *testing.T) {
-	key := GenerateKey()
+	key, _ := GenerateKey()
```
arlolra: Let's not ignore the error.

history.go
```diff
@@ -0,0 +1,75 @@
+package merkletree
+
+import (
+	"errors"
+
+	"github.com/coniks-sys/libmerkleprefixtree-go/crypto"
+)
+
+var (
+	ErrHistoryExisted = errors.New("[merkletree] trying to override existing history")
```
arlolra: `ErrHistoryExisted` isn't being used.

history_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+import (
+	"bytes"
+	"reflect"
+	"testing"
+
+	"github.com/coniks-sys/libmerkleprefixtree-go/crypto"
+)
+
+var signKey, _ = crypto.GenerateKey()
```
arlolra: Let's put this in an `init()` and not ignore the error.

history_test.go
```diff
@@ -0,0 +1,127 @@
...
+	key1 := "key"
+	val1 := []byte("value")
+
+	key2 := "key2"
+	val2 := []byte("value2")
+
+	key3 := "key3"
+	val3 := []byte("value3")
+
+	m1 := InitMerkleTree(&DefaultPolicies{}, treeNonce)
```
arlolra: It's kind of confusing that `treeNonce` is declared and defined in another `_test.go` file.

history.go
```diff
@@ -0,0 +1,75 @@
...
+	ErrBadEpoch       = errors.New("[merkletree] next epoch's STR has bad epoch ")
+)
+
+type History struct {
+	tree          *MerkleTree
+	currentSTR    *SignedTreeRoot
+	epochInterval int64
+}
```
arlolra: Maybe we can just call these `tree`, `str`, and `epoch`?

history.go
```diff
@@ -0,0 +1,75 @@
...
+
+func NewHistory(m *MerkleTree, key crypto.SigningKey, startEp, epInterval int64) *History {
+	h := new(History)
+	h.tree = m
+	h.epochInterval = epInterval
+	h.currentSTR = newSTR(m, startEp, 0, make([]byte, crypto.HashSizeByte), key)
+	return h
+}
+
+func (h *History) UpdateHistory(m *MerkleTree, nextEp int64) error {
```
arlolra: Why does `UpdateHistory` take any arguments?

I imagine that it would do something like,
```
func (h *History) UpdateHistory() {
  h.tree.RecomputeHash()
  clone := h.tree.clone()
  nextEp := h.currentSTR.epoch + h.epochInterval
  h.currentSTR = h.currentSTR.generateNextSTR(clone, nextEp)
}
```

README.md
```diff
@@ -8,13 +8,38 @@ This library currently uses the `SHAKE128` ShakeHash with output size of 32 byte
...
+```
+// clone the tree
+m = m.Clone()
+// insert new data
+m.Set(key, value)
+...
+// recompute tree hash
+m.RecomputeHash()
+// update STR history chain
+history.UpdateHistory(m, nextEpoch)
```
arlolra: I think a lot of this should be buried in `UpdateHistory`; see the comment below.  Is there any reason to manually clone and recompute the hash between epochs?

README.md
```diff
@@ -8,13 +8,38 @@ This library currently uses the `SHAKE128` ShakeHash with output size of 32 byte
...
+```
+
+Look-up
+
+`Get(key)` and `GetInEpoch(key, epoch)` return a `MerkleNode` instance and an `AuthenticationPath` for proofs of inclusion/absence.
```
arlolra: I'm assuming these are the ones from `History` and are therefore called `LookUp` and `LookUpInEpoch`.

str.go
```diff
@@ -0,0 +1,72 @@
...
+	epoch       int64
+	prevEpoch   int64
+	prevStrHash []byte
+	sig         []byte
+	policies    []byte
+	prev        *SignedTreeRoot
+	key         crypto.SigningKey
+}
+
+func newSTR(m *MerkleTree, ep, prevEp int64, prevHash []byte,
```
arlolra: Capital, `NewSTR`, like the others, please.

history.go
```diff
@@ -0,0 +1,75 @@
...
+func (h *History) NextEpoch() int64 {
+	if h.currentSTR == nil {
+		return h.epochInterval
+	}
+	return h.epochInterval + h.currentSTR.epoch
+}
+
+func (h *History) MerkleTree() *MerkleTree {
+	return h.tree
+}
```
arlolra: This method isn't being used.

history.go
masomel: ```epoch``` and ```epochInterval``` are different things. ```epoch``` is the actual time at which the next history is updated, whereas ```epochInterval``` is the frequency at which the history is updated. Might be worth renaming the latter to ```updateFrequency``` or something like that.

arlolra: That's why I hedged with my, "Or not ..."

Feel free to drop this comment.  Concise doesn't always mean clear (or, correct).

history.go
```diff
@@ -0,0 +1,75 @@
...
+	return nil
+}
+
+func (h *History) LookUp(key string) (MerkleNode, *AuthenticationPath) {
+	str := h.currentSTR
+	return str.tree.Get(key)
+}
+
+func (h *History) LookUpInEpoch(key string, ep int64) (MerkleNode, *AuthenticationPath) {
+	str := h.GetSTR(ep)
```
arlolra: As it stands, `GetSTR` can return `nil`.  That should be handled.

history.go
```diff
@@ -0,0 +1,75 @@
...
+	return str.tree.Get(key)
+}
+
+func (h *History) GetSTR(ep int64) *SignedTreeRoot {
+	pointer := h.currentSTR
+	for pointer.epoch > ep && pointer != nil {
```
arlolra: The order of these checks seems off.  Wouldn't you want to check if it's `nil` before dereferencing?

Also, what happens if `GetSTR` is given an `ep` that's greater than the current epoch?  Does it make sense to return the current str, as it's doing? Or, should we err? Similarly, if it's less than the starting epoch (since that's a parameter to the `NewHistory`).

Those are bugs, but I think returning `nil` is ok.  `LookUpInEpoch` can be the one to return an error if it cannot find the str for the requested epoch.

history.go
```diff
@@ -0,0 +1,75 @@
...
+	return h.tree
+}
+
+func (h *History) NewTemporaryBinding(index, value []byte) []byte {
+	str := h.currentSTR
+	tb := getSTRBytesForSig(str.tree, str.epoch, str.prevEpoch, str.prevStrHash)
+	tb = append(tb, index...)
+	tb = append(tb, value...)
+	return crypto.Sign(str.key, tb)
+}
```
arlolra: I think you should remove this from this PR.  We can review it with the rest of #13 when that happens.  Let's try to keep the PRs as small as possible to ease the cognitive burden.

history.go
masomel: I don't really disagree with you, though. It might make sense to rename ``currentSTR`` to ``str`` if what the ```History``` struct represents is a link or snapshot (or block if we want to go there) in the server's history. And to make this more clear, I might even be inclined to rename ``History`` to something more accurate. What do you think?

README.md
```diff
@@ -8,13 +8,38 @@ This library currently uses the `SHAKE128` ShakeHash with output size of 32 byte
 The signature scheme is `Ed25519` signature algorithm.
 
 ### Usage
-Initiate the tree
+Initiate the tree & history
 ```
 // using DefaultPolicies as current policy of the tree
 m := merkletree.InitMerkleTree(&DefaultPolicies{}, "nonce", "salt")
```
arlolra: "salt" is no longer here.

history.go
arlolra: Yeah, that's probably not the right term since the history is actually stored in the linked list of `SignedTreeRoot`s.

Sorry, I'm kind of spinning on the API we want to expose as this all becomes more familiar to me.

Is the right name `PAD` then? (being a persistent authenticated dictionary, which allows queries against older versions of the structure, to quote that paper).  The developer would instantiate a new `PAD`, which would in turn construct the Merkle tree as part of its internals.

masomel: > Is the right name ``PAD`` then? (being a persistent authenticated dictionary, which allows queries against older versions of the structure, to quote that paper). The developer would instantiate a new ``PAD``, which would in turn construct the Merkle tree as part of its internals.

Yeah, I think ``PAD`` is probably the right name for our Merkle tree + STR hash chain construction. But I'm not sure if PAD is the right name for what ``History`` represents right now. Going through the PAD paper, they seem to use the term "snapshot" for each version of the tree (see sections 2 and 3.2). But in essence, the STR already *is* a snapshot of the Merkle tree, plus some additional data, so I actually don't think we need a separate struct for ``History``. We can still form a linked list of STRs if that's the structure we prefer, though the hash chain doesn't technically require a linked list (i.e., we could just store each STR in an array, much like the "snapshot array" described in the PAD paper).

masomel: To circle back to the original question, ``epochInterval`` could be included in the policy, for example.

crypto/crypto.go
c633: It would be used outside the library. An example is when the developer reads the signing key from a file and he wants to validate the key size before using it.

history.go
c633: In the beginning of each epochs, the tasks would be:

1. Clone new tree from current tree (since we have to keep the state of the tree in each epoch)

2. Insert new data(s)

3.  Recompute tree hash

4. Update History chain

That is the reason why the developers must manually clone and recompute the hash. It may be cumbersome but I haven't found a better way yet.

README.md
c633: Ahhh, forgot to update the `README`

history.go
c633: It's is used outside the library. The developer would need to call the `Clone` method of the tree so he needs a method to access the tree.

history.go
c633: > Also, what happens if GetSTR is given an ep that's greater than the current epoch? Does it make sense to return the current str, as it's doing? Or, should we err? Similarly, if it's less than the starting epoch (since that's a parameter to the NewHistory).

Yes, I also think about it. May @masomel know how to handle these situations? The current flow is adopted from the reference implementation.

history.go
c633: Yes.

history.go
c633: > To circle back to the original question, epochInterval could be included in the policy, for example.

Yes, I added the `EpochInterval` to the `DefaultPolicy` structure.

I also renamed the `currentSTR` to `str`.

history.go
arlolra: The discussion @masomel and I were having above was hinting at the better way though.  Think about the end of each epoch as taking a snapshot of the current state of the tree.  In your numbering, the order would be: 2, 2, 2, 2, then 3, 1, 4. That's how I redefined `UpdateHistory` above.

The developer then doesn't need to touch anything else besides the initial tree it constructed and calling update

Does that help?

(An aside, does the developer need really need to manually call update? Or should a timer be set that invokes it automatically at every epoch interval?)

crypto/crypto.go
arlolra: Ok

history.go
arlolra: I answered my own questions with,

> I think returning nil is ok. LookUpInEpoch can be the one to return an error if it cannot find the str for the requested epoch.

history.go
c633: > (An aside, does the developer need really need to manually call update? Or should a timer be set that invokes it automatically at every epoch interval?)
 
The developer needs to setup the timer manually (https://github.com/coniks-sys/libconiks-server-go/blob/8d8e64535d4647ca336a7531f50229e2336e09b6/coniks/server.go#L141).

arlolra: > The developer needs to setup the timer manually

I was asking if that's something we should be doing for them though.

c633: > The discussion @masomel and I were having above was hinting at the better way though. Think about the end of each epoch as taking a snapshot of the current state of the tree. In your numbering, the order would be: 2, 2, 2, 2, then 3, 1, 4. That's how I redefined UpdateHistory above.

This way, new data would be inserted to the tree at the end of epoch. This is different from what is proposed in the paper (section 4.1.1)

It also rises another issue: we have 2 different STRs of each epoch, right? One from the beginning of the epoch to before the end of the epoch, and the other is at the end of the epoch. Did I missed something here?

c633: > I was asking if that's something we should be doing for them though.

Yes. I mean there are some extra works need to be done by the developer before/after updating the tree. Of course we can setup the timer as a part of the lib but I think it would have no advantages here. I need to think more about it.

arlolra: > This way, new data would be inserted to the tree at the end of epoch, not the next version of the tree. This is different from what is proposed in the paper (section 4.1.1)

> It also rises another issue: we have 2 different STRs of each epoch, right? One from the beginning of the epoch to before the end of the epoch, and the other is at the end of the epoch. Did I missed something here?

I may be reading this wrong, but as I understand it, an epoch is discrete. The current epoch is really the most recent snapshot that's been captured.  Remember, in between epochs (including the first) we hand out temporary bindings, which aren't yet included in the str (but are find to put in the tree since they won't be included in the current epoch until the next snapshot occurs).

> Yes. I mean there are some extra works need to be done by the developer before/after updating the tree. Of course we can setup the timer as a part of the lib but I think it would have no advantages here. I need to think more about it.

Ok, good. I was really asking because I don't know either. You make a good point.

c633: > I may be reading this wrong, but as I understand it, an epoch is discrete. The current epoch is really the most recent snapshot that's been captured. Remember, in between epochs (including the first) we hand out temporary bindings, which aren't yet included in the str.

Yes. That's right. But how's about my question: 
> we have 2 different STRs of each epoch, right?

I'm asking it because I have some concerns related to the monitoring protocol.

arlolra: > we have 2 different STRs of each epoch, right?

I don't think so. Let me try to explain and see if you agree.

At `t == 0`, we don't have any STRs, we've just handed out a few temporary bindings. We can set them in the tree, but since we haven't signed the root yet, any lookups just come from the temporary bindings.

At `t == 1`, we take a snapshot (clone) of the tree as it is and sign the root. We shift that STR in an array. We continue to hand out temporary bindings and set them in the tree. Now, when a lookup comes, we take `array[0]` and lookup in there. If we don't find it, we search our temporary bindings as a fallback.

At `t == 2`, we repeat the same as `1`.

Since the STRs always contain a clone of the tree, we're free to update the tree in between epochs. An epoch as a snapshot of the state at a current time. That's what I mean by it being discreet.

Is that helpful?

arlolra: Note, I may be wrong in saying that lookups fallback to searching the temporary bindings. Need to check the paper on that.

c633: > At t == 0, we don't have any STRs.

I don't think so. As illustration in the paper (figure 2), there is a STR for `t==0`. In the implementation, we also take a STR in the beginning (https://github.com/coniks-sys/libmerkleprefixtree-go/blob/0f821615420293b4c098ea9ccf93450950006c3d/history.go#L22), so does the reference implementation (https://github.com/coniks-sys/coniks-ref-implementation/blob/master/coniks_server/src/org/coniks/coniks_server/ServerHistory.java#L68)

> Note, I may be wrong in saying that lookups fallback to searching the temporary bindings

I think you are right at this point. @masomel also mentioned it in the temporary binding email thread.

arlolra: > I don't think so. As illustration in the paper (figure 2), there is a STR for t==0.

Ok, let's shift my number so that `t == 0` is now `t == -1` :)

The salient point is that the epochs are discreet. Did that resonate with you?

arlolra: > I think you are right at this point.

Sounds like it.

> 2) to allow immediate use of the key without having to wait for the next epoch. So if Alice registers (alice, pk) with the key server, the server will return a corresponding TB along with a proof of absence. At the same time, if Bob wants to send Alice a message before the next epoch, the key server also returns the TB for (alice, pk) and the proof of absence.

c633: > At t == 1, we take a snapshot (clone) of the tree as it is and sign the root.

So this STR is the STR for the tree at `t == 1`. The temporary bindings (namely `TB_0`) which is set in the tree at `t == 1` (namely `tree_1`) would belong to the tree at `t == 0` (namely `tree_0`), since we inserted them into the tree before cloning the tree. 

As a result, the `STR[0]` of `tree_0` (or `array[0]` as you called) is not computed based on `TB_0`, whereas the `STR[1]` of `tree_1` would be computed based on `TB_0`. 

Is that right?

Btw, thank you for your patience so far 😭

c633: > we have 2 different STRs of each epoch, right?

No. I'm wrong.

Liamsi:
Sorry for being slow on this review. I'll have a look into it today.

README.md
```diff
@@ -8,13 +8,38 @@ This library currently uses the `SHAKE128` ShakeHash with output size of 32 byte
...
+m := merkletree.InitMerkleTree(&DefaultPolicies{EpochInterval:60}, "nonce")
 // insert existing data (e.g., from db)
 m.Set(key, value)
 ...
+// recompute tree hash
+m.RecomputeHash()
+// generate private key for STR signing
+signKey := crypto.GenerateKey()
+// init STR history chain
+history := NewHistory(m, signKey, startupEpoch, epochInterval)
```
Liamsi: The signature of `NewHistory` changed (added `size`)

crypto/crypto.go
```diff
@@ -27,18 +25,19 @@ func Digest(ms ...[]byte) []byte {
...
 
-func Sign(key KeyPair, message []byte) []byte {
-	return ed25519.Sign(key.PrivateKey, message)
+func Sign(key SigningKey, message []byte) []byte {
+	return ed25519.Sign(ed25519.PrivateKey(key), message)
 }
 
-func Verify(key KeyPair, message, sig []byte) bool {
-	return ed25519.Verify(key.PublicKey, message, sig)
+func Verify(key SigningKey, message, sig []byte) bool {
```
Liamsi: Shouldn't the `Verify`directly take the public key instead?

crypto/crypto.go
```diff
@@ -27,18 +25,19 @@ func Digest(ms ...[]byte) []byte {
...
+	pk, ok := ed25519.PrivateKey(key).Public().(ed25519.PublicKey)
+	if !ok {
+		return false
+	}
+	return ed25519.Verify(pk, message, sig)
```
Liamsi: OK, I see. This Verify function is just a wrapper around `ed25519.Verify` which everyone who only has the public key would use.

history.go
```diff
@@ -0,0 +1,88 @@
...
+
+var (
+	ErrBadEpoch           = errors.New("[merkletree] Next epoch's STR has bad epoch")
+	ErrSTRNotFound        = errors.New("[merkletree] STR not found")
+	ErrMalformedHashChain = errors.New("[merkletree] Malformed hash chain")
+)
+
+type History struct {
+	currentSTR    *SignedTreeRoot
+	hashChain     map[int64]*SignedTreeRoot
```
Liamsi: Not sure: Wouldn't it be nice to define a type which aliases `int64` and call it `epoch` (or sth. similar to `type Epoch int`) . So everyone sees the map definition, directly recognizes, that this is in fact a map from epochs to signed tree roots?

history.go
```diff
@@ -0,0 +1,88 @@
...
+	epochs        []int64 // slice of epochs in hashChain
+	epochInterval int64
+}
+
+func NewHistory(m *MerkleTree, key crypto.SigningKey, startEp, epInterval, size int64) *History {
```
Liamsi: I think, it could make sense to add a comment, what `size` actually is or give a slightly more specific name.

history.go
c633: Great!

crypto/crypto.go
c633: This would be removed from the lib, I think.

merkletree_test.go
```diff
@@ -4,10 +4,80 @@ import (
...
+		t.Error("Cannot find value of key:", key)
+		return
+	}
+	v := r.Value()
+	if !bytes.Equal(v, val) {
+		t.Errorf("Value mismatch %v / %v", v, val)
+	}
+
+	// Check leaf node hash
+	h = sha3.NewShake128()
```
Liamsi: This is just test code: but you could do `h.Reset()` instead of creating a new instance.

history.go
```diff
@@ -7,37 +7,60 @@ import (
...
 
 var (
-	ErrBadEpoch    = errors.New("[merkletree] Next epoch's STR has bad epoch")
-	ErrSTRNotFound = errors.New("[merkletree] STR not found")
+	ErrBadEpoch           = errors.New("[merkletree] Next epoch's STR has bad epoch")
+	ErrSTRNotFound        = errors.New("[merkletree] STR not found")
+	ErrMalformedHashChain = errors.New("[merkletree] Malformed hash chain")
 )
 
+type Epoch int64
```
masomel: +1

history.go
```diff
@@ -7,37 +7,60 @@ import (
...
+
 type History struct {
-	str           *SignedTreeRoot
-	epochInterval int64
+	currentSTR    *SignedTreeRoot
+	hashChain     map[Epoch]*SignedTreeRoot
+	epochs        []Epoch // slice of epochs in hashChain
+	epochInterval Epoch
```
masomel: If this is already in the STR's policy, do we also need it to be a field in the ``History`` struct?

history.go
masomel: I'm very sorry for chiming in very late here. I would like to clarify a few more things.

> At t == 0, we don't have any STRs, we've just handed out a few temporary bindings. We can set them in the tree, but since we haven't signed the root yet, any lookups just come from the temporary bindings.

Whether we start counting epochs at t=0 or t=-1, the initial epoch *does* need an STR, even if the tree is empty, since we need to return a proof of absence along with any TB.

>> At t == 1, we take a snapshot (clone) of the tree as it is and sign the root.
> So this STR is the STR for the tree at ``t == 1``. The temporary bindings (namely ``TB_0``) which is set in the tree at ``t == 1`` (namely ``tree_1``) would belong to the tree at ``t == 0`` (namely ``tree_0``), since we inserted them into the tree before cloning the tree.
> 
> As a result, the ``STR[0]`` of ``tree_0`` (or ``array[0]`` as you called) is not computed based on ``TB_0``, whereas the ``STR[1]`` of ``tree_1`` would be computed based on ``TB_0``. 
> Is that right?

Yeah, this is right. But even if we included all ``TB_0``s in a batch at the end of t=0 like you had originally suggested (and as we do in the reference implementation), since the initial epoch t=0 has an STR,  the next STR for t=1 and ``tree_1`` would be computed based on all ``TB_0``s.

c633: > But even if we included all TB_0s in a batch at the end of t=0 like you had originally suggested (and as we do in the reference implementation), since the initial epoch t=0 has an STR, the next STR for t=1 and tree_1 would be computed based on all TB_0s.

Is this wrong? According to the paper:
> The binding is signed by the identity provider, creating a non-repudiable promise to add this data to the next version of the tree.

> The client then needs to wait for the next epoch and ensure that the provider has kept its promise of inserting Alice’s binding into its key directory by the next epoch.

So as I understand, the STR of epoch `t` and `tree_t` would be computed based on all `TB_(t-1)`s (i.e., `STR[1]` of `tree_1` should be computed based on all `TB_0`s), no?

history.go
c633: Yes. They has the same value. But I think it is quicker to have this field in the `History` struct, instead of referencing to the value stored in the STR's policy (which is actually the value stored in `Policies` struct in `MerkleTree`)

history.go
```diff
@@ -7,37 +7,60 @@ import (
...
+	ErrMalformedHashChain = errors.New("[merkletree] Malformed hash chain")
 )
 
+type Epoch int64
+
 type History struct {
-	str           *SignedTreeRoot
-	epochInterval int64
+	currentSTR    *SignedTreeRoot
+	hashChain     map[Epoch]*SignedTreeRoot
```
masomel: I think having an array of ``snapshots`` might be clearer than calling the entire array ``hashChain``.

history.go
arlolra: Does the `MerkleTree` need policies? Or should that belong to `History`?

history.go
```diff
@@ -7,37 +7,60 @@ import (
...
+	epochs        []Epoch // slice of epochs in hashChain
```
masomel: ``epochs`` represents the range of the hash chain that is in memory, right? I might rename this to ``loadedEpochs`` or something like that.

history.go
```diff
@@ -46,24 +69,24 @@ func (h *History) LookUpInEpoch(key string, ep int64) (MerkleNode, *Authenticati
 	return n, ap, nil
 }
 
-func (h *History) GetSTR(ep int64) *SignedTreeRoot {
-	pointer := h.str
-	for pointer.epoch > ep && pointer != nil {
-		if pointer.prev == nil {
+func (h *History) GetSTR(ep Epoch) *SignedTreeRoot {
+	pointer := h.hashChain[h.currentSTR.epoch]
+	for pointer.epoch > ep && h.hashChain[pointer.epoch] != nil {
```
masomel: Since we have the ``epochs`` array, wouldn't it be easier to loop through that to find the right STR pointer?

masomel: And if we reach the end of the ``epochs`` range, I think we need to eventually look through our persistent storage to see if we have an appropriate STR.

history.go
c633: OK.

history.go
c633: It is actually an array of keys of `hashChain` map (that's why I commented in this line).

history.go
c633: Hmm, I think I will do it in #4 implementation.

c633: For now, I think it' safe to make an assumption that the hash chain is long enough that we don't need to load the older STRs from persistent storage.

c633: > Since we have the epochs array, wouldn't it be easier to loop through that to find the right STR pointer?

We could do something like this:

```
func (h *History) GetSTR(ep Epoch) *SignedTreeRoot {
	if ep < h.epochs[0] {
		return nil
	}
	for i := 0; i < len(h.epochs)-1; i++ {
		if h.epochs[i] <= ep && ep < h.epochs[i+1] {
			return h.hashChain[h.epochs[i]]
		}
	}
	return h.hashChain[h.currentSTR.epoch]
}
```
but I think the current implementation looks more "natural" and is easier to understand. What do you think?

history.go
c633: @masomel : Is the policies immutable? Or could the developers change it at some epochs?

README.md
Liamsi: Is this still relevant? I also think it would make sense that the caller shouldn't need to take care about calling `RecomputeHash`. But maybe you already agreed that this is the way to go?

history.go
Liamsi: Question: The keys are already stored in the map. Is there anything which couldn't be (easily) done without additionally storing the keys in an array?

README.md
c633: We're still discussing about it. You can join it here: https://github.com/coniks-sys/libmerkleprefixtree-go/pull/8#discussion_r69050418

history.go
c633: I use this array to delete older values stored in the map, since the order of the keys in golang map is randomized. (https://github.com/coniks-sys/libmerkleprefixtree-go/blob/496b4a6c345678de2a7fcec3f35b0e91c4fa6da9/history.go#L49)

masomel: Yes, I realized that ``epochs`` stores the range of keys of the map (I should have been more specific, sorry), and that you use it to keep track of which elements in the map are currently loaded in memory. This is why I was suggesting the name change.

history.go
masomel: The current implementation does more naturally traverse the hash chain, this is true. But I think taking advantage of having the ``epochs`` structure is a bit cleaner since there is one less level of indirection trying to get at the STR's epoch. On the other hand, using ``epochs`` as the key into ``hashChain`` will probably require us to confirm that ``str.epoch == h.epochs[i]`` once the appropriate STR in the chain has been selected.

masomel: I'm wondering if such a check would also be needed for the current implementation. I think using ``epochs`` could be a bit cleaner, but either implementation works for me. Maybe others have stronger opinions one way or the other.

history.go
masomel: > I think it is quicker when accessing this field in the History struct, instead of referencing to the value stored in the STR's policy.

The reason why I prefer to not have a value stored in two places is because we then always have to ensure that they are consistent or synchronized with each other. If the policy is accessible to the ``History`` struct via the STR, even if it's a bit slower, I think it's safer and anybody reading the code knows exactly where the value came from.

masomel: > Does the ``MerkleTree`` need policies? Or should that belong to ``History``?

We put the policy in the STR (not the tree) for two reasons. Many policies are tree-specific (e.g. the key for computing the VRF), and because we want the server to commit (not cryptographically) to the policies by including them in the STR.

From the paper, section 3.3: "P may include, for example, the key K_VRF used to generate private indices, an expected time the next epoch will be published, as well as the cryptographic algorithms in use, protocol version numbers, and so forth."

I think this also leads us to answer @c633's question 
> Is the policies immutable? Or could the developers change it at some epochs?

No, the policies aren't immutable, the developers can change them between epochs. Protocol version numbers are a great example of something that could change. The server could also change it's VRF public key between epochs (this helps re-randomize the order of leaves), or change its epoch interval.

history.go
masomel: > So as I understand, the STR of epoch t and tree_t would be computed based on all TB_(t-1)s (i.e., STR[1] of tree_1 should be computed based on all TB_0s), no?

Yes, this is right. This is exactly what I was saying, too.

history.go
c633: Yeah, I got it.

history.go
c633: Sorry, I might completely misunderstanding something here. 

The changes @arlolra suggesting is: insert new data to the tree, recompute the tree hash, clone the tree and generate new STR for new tree. 

So, at each epoch, the developer would do something like this:
```
func UpdateDirectory(h *History) {
    h.MerkleTree().Set(key, value)
    ....
    h.UpdateHistory()
}
```

Assume at epoch `t`, Alice send a registration request, then Alice's key would be inserted to the tree at epoch `t+1`. It means the Alice's key should not be found in the tree at epoch `t` but at epoch `t+1`, right?

But if we do as above, Alice's key would be found in the tree at epoch `t` (since we inserted to the tree before cloning it). 

Do I have any mistakes?

history.go
c633: >No, the policies aren't immutable, the developers can change them between epochs. Protocol version numbers are a great example of something that could change. The server could also change it's VRF public key between epochs (this helps re-randomize the order of leaves), or change its epoch interval.

Thanks!

history.go
masomel: > Assume at epoch t, Alice send a registration request, then Alice's key would be inserted to the tree at epoch t+1. It means the Alice's key should not be found in the tree at epoch t but at epoch t+1, right?

Correct.

> But if we do as above, Alice's key would be found in the tree at epoch t (since we inserted to the tree before cloning it). 
> 
> Do I have any mistakes? 

I think you're right. For @arlolra's suggestion to work in the same way we think tree updates should work, the inserts into the tree that occur between epochs need to be done on an internal copy of the tree while lookups need to be done on ``curSTR.tree``. Otherwise, the server has to generate an STR each time a new leaf is inserted into ``curSTR.tree`` for lookups to work as intended. I think it's more work to keep track of a shadow copy of the tree between epochs than to clone the tree first and then insert all changes.

masomel: Actually, another way to think about the suggested changes is to have the history be ahead of the current epoch. Then it would be safe to clone the tree after the inserts. In other words, after ``STR_t`` is created and ``tree_t``, we create the new snapshot for ``t+1`` by cloning ``tree_t`` into ``tree_t+1``. Then it is safe to insert new leaves into ``tree_t+1`` as long as lookups are done in ``tree_t`` until ``STR_t+1`` is created.

Thus, ``UpdateHistory()`` could look something like this:
``
func (s *Snapshot) UpdateHistory() (nextSnapshot *Snapshot){
  s.tree.RecomputeHash()
  nextEp := s.currentSTR.epoch + s.epochInterval
  s.currentSTR = s.currentSTR.generateNextSTR(s.tree, nextEp)
 nextSnapshot = NewSnapshot(s.tree.clone())
 nextSnapshot.prev = s
 return
}
``
I know this isn't exactly true to the implementation of ``History`` and STRs but I hope you get the idea. What do you think?

c633: Should `nextEp` be a parameter, which is exactly the time when `UpdateHistory` is called? Or it could be computed like this `nextEp := h.nextSTR.epoch + h.nextSTR.P.epochInterval` ? 
I think the former is more proper way?

history.go
c633: Since policies are epoch-specific, I think `UpdateHistory()` should have `Policies` as a parameter. I updated the code in a2ff2042fd80b36947392cccb6a49f33e328b458

arlolra:
Something else worth considering here is renaming `History` to `PAD` (or, similar) since that's looking like the main entrypoint to the library and closer to what it really represents.  PAD being an acronym for persistent authenticated dictionary, as mention in the paper Marcela referenced.

https://dl.acm.org/citation.cfm?doid=2019599.2019602

c633:
I updated few things in 566fa07fea02ef45149c919e127204ee8b32960c

* Auto generate tree nonce instead of passing it into the function

* Rename InitMerkleTree to NewMerkleTree for consistency

* Move SigningKey from STR to History struct

* Refine History APIs (credit to @arlolra)

arlolra:
@c633 How do you feel about the cleanup proposed in 937e2470eddb77ac54b595911fa42208eef0038f?  I'd prefer to keep methods closer to the struct.

c633:
It looks a lot better.

history.go
```diff
@@ -54,11 +68,15 @@ func (h *History) UpdateHistory(policies Policies, m *MerkleTree, nextEp Epoch)
...
 		h.loadedEpochs = append(h.loadedEpochs[:0], h.loadedEpochs[n:]...)
 	}
-	h.snapshots[nextEp] = nextStr
-	h.loadedEpochs = append(h.loadedEpochs, nextEp)
+
+	h.updateInternal(policies, nextEp)
 	return nil
 }
 
+func (h *History) Set(key string, value []byte) error {
```
masomel: It definitely it makes sense to have a wrapper around ``tree.Set()`` to abstract away the tree, but now that I think about it, isn't the semantics we want more like ``directory.Set()``? Similarly, below, wouldn't we want more something like ``directory.Lookup()``? If what we want is to abstract away the underlying PAD (including the history), shouldn't we only expose an interface to a directory at the server level?

str.go
```diff
@@ -70,3 +73,45 @@ func (str *SignedTreeRoot) serialize() []byte {
 
 	return strBytes
 }
+
+func (str *SignedTreeRoot) MarshalJSON() ([]byte, error) {
```
masomel: I think we should move all of the JSON marshaling functions to the server. I think we should keep this library only for the data structure.

history.go
masomel: I also just realized that a directory lookup also needs to return an STR, so I think this is another reason why we should move these wrappers into the server.

c633: What do you mean "the server level"? Is it our libconiks-server library? 
I think it's fine to have these wrappers in this library. We can think these wrappers as APIs of the library, and in the libconiks-server library, we should have something like `directory.LookUp` and `directory.Set()`. Is that what you mean?

masomel: > What do you mean "the server level"? Is it our libconiks-server library? 

Yes, I mean the libconiks-server library. 

> I think it's fine to have these wrappers in this library. We can think these wrappers as APIs of the library, and in the libconiks-server library, we should have something like ``directory.LookUp`` and ``directory.Set()``. Is that what you mean? 

I don't really know if ``History`` makes the most sense as the interface, though. I've been thinking of ``History`` more as the structure that gives developers access to older versions of the Merkle tree. Maybe this is coming back to an earlier discussion we had about the naming of ``History``. I don't see the problem in exposing the tree setter and getter methods to the server directly since we're exposing many other aspects of the tree anyway (e.g. the ``MerkleNode`` from the auth path).

arlolra: On my way out, but I wanted to leave a pointer to this comment, in case it's relevant here.  https://github.com/coniks-sys/libmerkleprefixtree-go/pull/8#issuecomment-231272272

masomel: Ah, I missed this comment for some reason. Thanks for pointing it out. I agree, PAD is probably the right name.

c633: I renamed to `PAD` in f1b8ca09ffa48dbc8fad13e964377c00bab4c4d6

merkletree.go
```diff
@@ -111,6 +125,7 @@ func (m *MerkleTree) insertNode(key []byte, toAdd *userLeafNode) {
 	depth := 0
```
arlolra: The argument name should be `index`, not `key`.

It might be nice to do `indexBits := util.ToBits(index)`, as above, rather than `GetNthBit`, for consistency.

merkletree.go
```diff
@@ -66,16 +76,20 @@ func (m *MerkleTree) Get(key string) (MerkleNode, []ProofNode) {
 	}
 	switch nodePointer.(type) {
 	case *userLeafNode:
-		proof = append(proof, nodePointer.(*userLeafNode))
+		authPath.index = nodePointer.(*userLeafNode).index
+		authPath.level = nodePointer.(*userLeafNode).level
+		authPath.leaf = nodePointer.(*userLeafNode)
 		if nodePointer.(*userLeafNode).key == key {
```
arlolra: Comparing indices would be preferable.

merkletree.go
```diff
@@ -132,7 +147,7 @@ insertLoop:
 				return
 			}
 
-			newInteriorNode := NewInteriorNode(currentNodeUL.parent, depth)
+			newInteriorNode := NewInteriorNode(currentNodeUL.parent, depth, prefixBits)
```
arlolra: You can then just use a slice of `indexBits` for this.

merkletree.go
c633: Ohh, I remember I fixed this in a previous commit. Thanks for pointing it out.

merkletree.go
```diff
@@ -175,13 +194,44 @@ insertLoop:
...
+}
+
+func (m *MerkleTree) recomputeHash() {
+	if m.root.leftHash == nil {
+		m.root.leftHash = m.hashNode(m.root.leftChild)
+	}
+	if m.root.rightHash == nil {
+		m.root.rightHash = m.hashNode(m.root.rightChild)
+	}
+	m.treeHash = m.root.hash()
```
arlolra: This can just be written as,

```
func (m *MerkleTree) recomputeHash() {
  m.treeHash = m.hashNode(m.root)
}
```

merkletree.go
```diff
@@ -175,13 +194,44 @@ insertLoop:
...
+func (m *MerkleTree) hashNode(node MerkleNode) []byte {
+	switch node.(type) {
+	case *interiorNode:
+		currentNodeI := node.(*interiorNode)
+		if currentNodeI.leftHash == nil {
+			currentNodeI.leftHash = m.hashNode(currentNodeI.leftChild)
+		}
+		if currentNodeI.rightHash == nil {
+			currentNodeI.rightHash = m.hashNode(currentNodeI.rightChild)
+		}
```
arlolra: Is there any reason not to move,

```
+		if currentNodeI.leftHash == nil {
 +			currentNodeI.leftHash = m.hashNode(currentNodeI.leftChild)
 +		}
 +		if currentNodeI.rightHash == nil {
 +			currentNodeI.rightHash = m.hashNode(currentNodeI.rightChild)
 +		}
```

into `func (n *interiorNode) hash() []byte {`?

In which case, this all simplifies down to `node.hash(m)` and the `hashNode()` function is not needed.

c633: Great!

str.go
```diff
@@ -0,0 +1,74 @@
...
+	bytesPreSig := str.getSTRBytesForSig()
+	str.sig = crypto.Sign(key, bytesPreSig)
+	return str
+}
+
+func (str *SignedTreeRoot) getSTRBytesForSig() []byte {
+	var strBytes []byte
+	strBytes = append(strBytes, util.LongToBytes(str.counter)...)     // t - counter number
+	strBytes = append(strBytes, util.LongToBytes(str.prevCounter)...) // t_prev - previous counter number
+	strBytes = append(strBytes, str.tree.root.serialize()...)         // root
```
arlolra: Shouldn't this be `treeHash`?

str.go
```diff
@@ -0,0 +1,74 @@
...
+	strBytes = append(strBytes, str.prevStrHash...)                   // previous STR hash
+	strBytes = append(strBytes, str.policies.Serialize()...)          // P
+	return strBytes
+}
+
+func (str *SignedTreeRoot) serialize() []byte {
+	var strBytes []byte
+	strBytes = append(strBytes, str.tree.root.serialize()...)         // root
```
arlolra: Same here.

str.go
```diff
@@ -0,0 +1,74 @@
...
+	strBytes = append(strBytes, util.LongToBytes(str.counter)...)     // counter
+	strBytes = append(strBytes, util.LongToBytes(str.prevCounter)...) // previous counter
+	strBytes = append(strBytes, str.prevStrHash...)                   // previous hash
+	strBytes = append(strBytes, str.sig...)                           // signature
+	return strBytes
+}
+
+func (str *SignedTreeRoot) Root() []byte {
+	return str.tree.root.serialize()
```
arlolra: `str.tree.GetHash`?

proof_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"bytes"
+	"testing"
+
+	"github.com/coniks-sys/libmerkleprefixtree-go/crypto"
+	"github.com/coniks-sys/libmerkleprefixtree-go/internal"
+)
+
+func computeLeafHash(ap *AuthenticationPath) (leafHash []byte) {
+	leaf := ap.Leaf()
+	if leaf.Commitment() != nil {
```
arlolra: This comparison makes me a little nervous.  Maybe you could implement `isEmpty` on `ProofNode` as well.

merkletree.go
c633: > In which case, this all simplifies down to node.hash(m) and the hashNode() function is not needed.

I think we still need the `hashNode()` function for type switch (https://github.com/coniks-sys/libmerkleprefixtree-go/blob/a28393a165dd7441544cc4f9c00afb4fc8a1a090/merkletree.go#L205)

arlolra: Ok, if so, move it to node.go as a standalone function, since it isn't a method of `MerkleTree`.

str.go
c633: I don't think so, according to the reference implementation: 
https://github.com/coniks-sys/coniks-ref-implementation/blob/master/coniks_server/src/org/coniks/coniks_server/ServerUtils.java#L364

The reason for including the `treeHash` field into the `MerkleTree` is discussed here:
https://github.com/coniks-sys/libmerkleprefixtree-go/pull/8#discussion_r68999957

But now I'm thinking that I should remove the `treeHash` field and move the `GetHash` method to the test source file to avoid confusing.

proof_test.go
c633: Yes

str.go
c633: I removed it in 41b68eea926dcebbbe599263b01d2e357f3ce1d0

str.go
arlolra: Hmm, I wonder why that is ...

The paper doesn't seem to specify that `root = h.child.0 || h.child.1`, so it's a bit unclear.  To me, the root node is just another interior node, and `root` in the STR computation should be the `H()` of that.  Let's let @masomel weigh in.

node.go
```diff
@@ -64,69 +66,39 @@ func NewInteriorNode(parent MerkleNode, level int, prefixBits []bool) *interiorN
...
+func hashNode(node MerkleNode, m *MerkleTree) []byte {
+	switch node.(type) {
+	case *interiorNode:
+		return node.(*interiorNode).hash(m)
+	case *userLeafNode:
+		return node.(*userLeafNode).hash(m)
+	case *emptyNode:
+		return node.(*emptyNode).hash(m)
+	}
+	return nil
```
arlolra: panic, instead of returning nil

str.go
c633: Yes. Same question from me.

masomel: Yeah, the root node is just an interior node (maybe we don't make this explicit in the paper? If not, then our bad). So root should also be ``h.child.0 || h.child.1``.

What we do in the reference implementation is just a vestige of an old root node format which contained many other fields (now in the STR). Sorry about this confusion! I think using ``treeHash`` in the STR makes perfect sense. And I should go back and fix this in the reference implementation, too.

merkletree.go
```diff
@@ -121,11 +121,11 @@ func computePrivateIndex(key string) []byte {
 	return crypto.Digest([]byte(key))
```
masomel: We can probably start using the VRF, now that we merged in #12.

arlolra: But please make that a separate pull, no?

arlolra:
@c633 What do you think of db4b36ceb5bd14c243ad89bfae4da1f71e8e0a7f?

merkletree.go
masomel: It should just be a one-liner, right? Do we need a separate PR for that?

str.go
```diff
@@ -13,14 +19,17 @@ import (
 // STR should be final
 type SignedTreeRoot struct {
 	tree        *MerkleTree
-	counter     int64
-	prevCounter int64
+	counter     uint64
```
masomel: I would still call these ``epoch`` and ``prevEpoch`` for clarity and consistency with the spec.

masomel:
Not exactly sure how I ended up with so many merge conflicts. I'm just going to reset HEAD back to 22c39af.

masomel:
Ok, fixed. Sorry for the noise, everyone.

str.go
```diff
@@ -0,0 +1,83 @@
...
+	var strBytes []byte
+	strBytes = append(strBytes, str.tree.hash...)                      // root
+	strBytes = append(strBytes, util.ULongToBytes(str.counter)...)     // counter
+	strBytes = append(strBytes, util.ULongToBytes(str.prevCounter)...) // previous counter
+	strBytes = append(strBytes, str.prevStrHash...)                    // previous hash
+	strBytes = append(strBytes, str.sig...)                            // signature
+	return strBytes
+}
+
+func (str *SignedTreeRoot) Root() []byte {
```
arlolra: Does this still have a use?

policy.go
```diff
@@ -5,6 +5,8 @@ import (
 	"github.com/coniks-sys/libmerkleprefixtree-go/internal"
 )
 
+type Epoch int64
```
masomel: I think naming the field in the policy ``Epoch`` is misleading now. Epochs should be the discreet points at which snapshots are taken.  I think ``timestamp`` or ``deadline`` or something similar makes more sense for what we're indicating in the policy now.

str.go
masomel: If we're just going to use ``tree.hash`` in the STR probably not.

c633: No. My mistake.

str.go
c633: Hmm, does it make sense? I mean now `epoch` is `uint64` type, and `prevEpoch` is just `epoch`- 1.

c633:
@arlolra Wow, it's so good. Thank you very much!!!

merkletree.go
c633: I think I agree with @arlolra, since if we change the `computePrivateIndex` then some tests would be broken. May @Liamsi  make a separate pull for that?

policy.go
c633: Is that what you mean? f7c37be1dd11ba690b198102056953c109d6de6d

merkletree.go
Liamsi: I also think it makes sense to make this a separate PR. Will do so in the following days, OK?

c633: Great. Thanks!

str.go
masomel: If the ``epoch`` is a monotonically increasing integer, then ``uint64`` seems like a perfectly valid type. My issue was more that ``counter`` isn't a very descriptive name for what it actually represents, the ``epoch``. The CONIKS design doesn't require that ``epoch`` be of a specific type, it only represents a point in the history of the PAD at which we take a snapshot.

policy.go
masomel: Yes, this looks good!

proof_test.go
arlolra: I agree with @masomel here that starting from the deepest level is more intuitive.  Here's a little implementation of what that looks like, 780d27bcacdc9dc8d560dd3654c98fe68089547c. @c633 how do you feel about it?

c633: Yeah. It's look a lot better now. Thanks for the commit to `str` branch. I will refine some more little things then I think this PR would be ready to be merged.

c633: I added a `TimeStamp` type for `epochDeadline` in d0747e4901c64d68eec7c9c7cdb8a92b3a411223 to distinguish from the integer `epoch`s?

pad.go
```diff
@@ -0,0 +1,105 @@
...
+	pad.loadedEpochs = make([]uint64, 0, len)
+	pad.updateInternal(policies, 1)
+	return pad, nil
+}
+
+// if policies is nil, the previous policies will be used
+func (pad *PAD) generateNextSTR(policies Policies, m *MerkleTree, epoch uint64) {
+	var prevStrHash []byte
+	if pad.currentSTR == nil {
+		prevStrHash = make([]byte, crypto.HashSizeByte)
```
arlolra: Shouldn't this be a `seed`?

pad.go
```diff
@@ -0,0 +1,105 @@
...
+	} else {
+		prevStrHash = crypto.Digest(pad.currentSTR.serialize())
```
arlolra: I think this should just be, `prevStrHash = crypto.Digest(pad.currentSTR.sig)`.

str.go
```diff
@@ -0,0 +1,85 @@
...
+	strBytes = append(strBytes, util.ULongToBytes(str.prevEpoch)...) // t_prev - previous epoch number
+	strBytes = append(strBytes, str.tree.hash...)                    // root
+	strBytes = append(strBytes, str.prevStrHash...)                  // previous STR hash
+	strBytes = append(strBytes, str.policies.Serialize()...)         // P
+	return strBytes
+}
+
+// serialize encodes the str to a byte array with the following format:
+// [tree hash, epoch, previous epoch, previous hash, signature]
+func (str *SignedTreeRoot) serialize() []byte {
```
arlolra: Why is this defined this way?  Are policies purposely omitted?  I don't think this function is necessary, and if it is, it should be defined in terms of `getSTRBytesForSig+str.sig`.

str.go
```diff
@@ -0,0 +1,85 @@
...
+		prevEpoch:   epoch - 1,
+		prevStrHash: prevHash,
+		policies:    policies,
+	}
+	bytesPreSig := str.getSTRBytesForSig()
+	str.sig = crypto.Sign(key, bytesPreSig)
+	return str
+}
+
+func (str *SignedTreeRoot) getSTRBytesForSig() []byte {
```
arlolra: Maybe just call this one `Serialize`

pad.go
```diff
@@ -0,0 +1,105 @@
...
+	var err error
+	pad := new(PAD)
+	pad.key = key
+	pad.tree, err = NewMerkleTree()
+	if err != nil {
+		return nil, err
+	}
+	pad.snapshots = make(map[uint64]*SignedTreeRoot, len)
+	pad.loadedEpochs = make([]uint64, 0, len)
+	pad.updateInternal(policies, 1)
```
arlolra: Why doesn't this start at `0`?

str.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func NewSTR(key crypto.SigningKey, policies Policies, m *MerkleTree, epoch uint64, prevHash []byte) *SignedTreeRoot {
+	if epoch < 1 {
+		panic(ErrorBadEpoch)
+	}
+	str := &SignedTreeRoot{
+		tree:        m,
+		epoch:       epoch,
+		prevEpoch:   epoch - 1,
```
arlolra: At epoch `0`, `prevEpoch` is `-1`.   It looks like the paper is saying that `prevEpoch` should be omitted from the STR at epoch `0` so we should probably check for that.

pad.go
c633: Yes. I just re-read the paper. It should start at `0`. Thanks!

pad.go
c633: May @masomel has a explanation for this? I also have same thought.

pad.go
c633: Let @masomel weigh in this question.

str.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	bytesPreSig := str.getSTRBytesForSig()
+	str.sig = crypto.Sign(key, bytesPreSig)
+	return str
+}
+
+func (str *SignedTreeRoot) getSTRBytesForSig() []byte {
+	var strBytes []byte
+	strBytes = append(strBytes, util.ULongToBytes(str.epoch)...)     // t - epoch number
+	strBytes = append(strBytes, util.ULongToBytes(str.prevEpoch)...) // t_prev - previous epoch number
```
arlolra:
```
if (str.prevEpoch > -1) {
    strBytes = append(strBytes, util.ULongToBytes(str.prevEpoch)...)
}
```

c633: since `prevEpoch` is `uint64` type. so I think it would be
```
if str.epoch > 0 {
        strBytes = append(strBytes, util.ULongToBytes(str.prevEpoch)...)
}

File: pad.go (hunk 56 above)
Comment (masomel) [inline pad.go]: We had imagined ``prevStrHash`` would be ``hash(currentSTR.serialize())`` because it's a hash chain of STRs, not a hash chain of their signatures (per [wikipedia's definition](https://en.wikipedia.org/wiki/Hash_chain)). I suppose this may just be an implementation detail, though?

File: str.go (hunk 57 above)
Comment (masomel) [inline str.go]: No, the policies should be included in this function. This function is needed if we define ``prevStrHash = hash(currentSTR.serialize())`. I agree, though, defining this function as ``getSTRBytesForSig+str.sig`` seems cleaner.

File: str.go (hunk 58 above)
Comment (masomel) [inline str.go]: Yeah, I was thinking this as well. This one should also be called serialize. Maybe have this function take a flag indicating if the signature should be included in the output buffer? Then we only need a single ``serialize()`` function.

File: str.go (hunk 60 above)
Comment (masomel) [inline str.go]: I'm actually not sure if this is what the paper is implying (could have just omitted it to make the figure look a bit cleaner). But because of https://github.com/coniks-sys/libmerkleprefixtree-go/pull/8#discussion_r70645139, I think it's fine to omit this value.

File: pad.go (hunk 55 above)
Comment (masomel) [inline pad.go]: Yes, this can be a ``seed``.

File: str.go (hunk 58 above)
Comment (c633) [inline str.go]: How's about `Serialize()` and `SerializeWithSignature()`?

Comment (masomel) [inline str.go]: This works, too.

File: pad.go (hunk 56 above)
Comment (c633) [inline pad.go]: So it's fine to keep the current implementation, I think.

**Comment by c633** [comment]
Updated some small changes in cc55e8e74d699b65ac36c9254a1638a1de2fc88e. What do you think? @masomel

**Comment by masomel** [comment]
> Updated some small changes in cc55e8e. What do you think? @masomel

I made a couple comments. Other than that, looks good!

File: proof_test.go [hunk 62]
```diff
@@ -63,18 +63,18 @@ func verifyProof(t *testing.T, ap *AuthenticationPath, treeHash []byte, key stri
 	} else {
 		// proof of absence
 		// step 2. Check that where i and j differ is at bit l
```
masomel: I think step 2 and step 3 should be switched. I don't think we should worry about checking the first ``l`` bits of the proof's index if ``vrf_verify`` returns false.

proof_test.go
```diff
@@ -10,22 +10,22 @@ import (
 
 func computeLeafHash(ap *AuthenticationPath) (leafHash []byte) {
 	leaf := ap.Leaf()
-	if !leaf.IsEmpty() {
+	if !ap.Leaf().IsEmpty() {
```
masomel: Why do you use ``ap.Leaf()`` here and below if you have ``leaf := ap.Leaf()`` above?

proof.go
```diff
@@ -17,18 +15,61 @@ func (ap *AuthenticationPath) PrunedTree() [][]byte {
...
 	return ap.leaf
 }
+
+type ProofNode interface {
+	Level() int
+	Index() []byte
+	Value() []byte
+	IsEmpty() bool
+	Commitment() []byte
+}
```
masomel: Could ``ProofNode`` also expose a ``Hash(treeNonce []byte)`` function? This would clean up the leaf hash computation a bit in the proof verification.

proof_test.go
c633: Agrr, my bad. Thanks!

proof.go
c633: Is it what the client needs to do manually? I think functions in `proof_test.go` are what the client needs to implement when receiving an authentication path, no?

proof_test.go
c633: Yes, you're right!

proof.go
masomel: Yeah, the client would do this manually. I was thinking that this could be facilitated if we had such a ``Hash()`` function. But that would only be helpful for a client written in Go. Nevermind then :)

node.go
```diff
@@ -111,13 +111,15 @@ func (n *interiorNode) Clone(parent *interiorNode) MerkleNode {
 			parent: parent,
 			level:  n.level,
 		},
+		leftHash:  append([]byte{}, n.leftHash...),
+		rightHash: append([]byte{}, n.leftHash...),
```
masomel: This should be ``rightHash: append([]byte{}, n.rightHash...)``, no?

masomel:
I'll make a separate issue for addressing https://github.com/coniks-sys/libmerkleprefixtree-go/pull/8#discussion_r70639162. @c633 Is this ready to be merged?

c633:
If it is a random seed, I already did it in https://github.com/coniks-sys/libmerkleprefixtree-go/blob/790fb35b2f61d76a0930a73ba543f9ffd6f43502/pad.go#L48-L51. Is that right?

masomel:
Yeah, that's what I would consider a random seed. I'll keep the issue open anyway to see if @arlolra was referring to something else.

c633:
If there is nothing else, I think it's ready to be merged.

Issue #9, opened 2016-06-22
Per user salt


arlolra:
> > ... you want a different nonce per-user so that you can selectively reveal, the tree nonce is for the whole tree.

> It looks to me that ```salt``` is also a tree-wide value. This needs to be changed.

@masomel can you elaborate on how you expect this to work?

c633:
I think the `userLeafNode` struct should have `salt` field to store the `salt` value per user.
As suggested by [1][2] we would use a salt that is the same size as the output of the hash function (i.e. 128-bit). So the `salt` field should be an array of 16 bytes. 

[1] https://crackstation.net/hashing-security.htm#salt
[2] https://stackoverflow.com/questions/184112/what-is-the-optimal-length-for-user-password-salt

arlolra:
The output of SHAKE128 that we're using is 256 bits though. So, if we're using the rule of thumb, let's get that right.

But maybe a 128 bit salt is fine. Let's see what @jcb82 says.

masomel:
Sorry for my delay. Yes, what I had in mind is essentially what @c633 described. ```salt``` should be a field of the ```userLeafNode```, and I believe a new salt should be computed every time the ```value``` is updated.

masomel:
I'm not certain what the size of the salt should be.

arlolra:
Thomas Pornin says http://security.stackexchange.com/a/11224

masomel:
So what this is saying is that we're probably fine with 128-bit salts generated at random, and we don't need to regenerate them each time the ```value``` is updated?

masomel:
Nevermind, I misread part of that StackExchange response. As https://github.com/coniks-sys/libmerkleprefixtree-go/pull/8/files#r68285691 mentions, we *should* be replacing the salt every time the ```value``` changes. Sorry for going back and forth on this.

arlolra:
Part of me thinks we should just reuse `HashSizeByte`.

c633:
> Part of me thinks we should just reuse HashSizeByte.
Agree. Updated in 30123ed70d6ca2c67ff07e005a939fe92357f225

arlolra:
Can you rebase the salt changes on master an open up a new pull for just this issue?

PR #11, opened 2016-06-24
per user salt 

c633:
PR for #9


arlolra:
Thanks!  Merged in f7179f02fbb4bfa6e8ba87b371fbcafb5a94ac2c.

I added added this pattern in the tests though,
```
if err := m.Set(key, val); err != nil {
    t.Fatal(err)
}
```

Issue #5, opened 2016-06-12
Add VRF implementation


masomel:
Currently, the private index is computed only using a cryptographic hash. This will also require adding a VRF key field in the STR policy.

Liamsi:
I'll create a PR later today (if no-one else already started working on this)

Liamsi:
I've added a branch with with the coname's VRF implementation  (almost unmodified): 
https://github.com/coniks-sys/libmerkleprefixtree-go/tree/vrf_impl_coname_agl
and a branch where I changed their code such that it should be exactly the implemetation from CONIKS paper:
https://github.com/coniks-sys/libmerkleprefixtree-go/tree/orig_coniks_vrf

There is no PR (yet) as I first wanted to examine/discuss if their changes bring any security benefits. But if you prefer Bonneau's orig. proposal in the CONIKS paper I can creat a PR from the 2nd branch.

Liamsi:
The main difference of the coname/dename VRF implementation to the one in the CONIKS paper is:
the VRF is defined as
VRF_k(m) = h(m, H(m)^k) instead of VRF_k(m) = H_1(m)^k
(where h is sha3 and H(m):=f(h(m))^8 (with f being the elligator map)). 
Additionally to s and t in the proof they return ii:=H(m)^k.

Everything else is equivalent. I remember @andres-erbsen and @daniel-ziegler mentioning they did some modifications to the original VRF construction to prevent some collision attacks. I'll ask them for some more details/links on that.

masomel:
If
> VRF_k(m) = h(m, H(m)^k) ... where h is sha3

does coname/dename subsequently still hash the result of VRF_k(m) to compute the private index?

masomel:
Looping @jcb82 into this discussion.

Joe did go back and address collision attacks in the Merkle tree design (which is where the tree nonce etc came from), but I can't recall if he modified the VRF design as well. I also know he was in contact with the coname authors last summer, though I don't know to which extent they worked together on the modifications to the VRF construction.

Liamsi:
> does coname/dename subsequently still hash the result of VRF_k(m) to compute the private index?

No it doesn't. So to be more precise while in CONIKS the index is computed by 
i = H(VRF_k (u)) = H(H_1(m)^k) 
in coname it is computed by 
i = h(m, H(m)^k)

masomel:
> No it doesn't the (already hashed VRF output) again.
[...]

PR #12, opened 2016-06-28
coname's VRF implementation (self-contained)

Liamsi:
This is an (almost) unmodified version of the coname's authors VRF implementation. As soon as this is merged I could integrate it into the existing code base.


Liamsi:
Related to issue #5

Liamsi:
We might want to put it or add it to the internal coname package instead?

If you wonder why we have yet another `ed25519`package: agl's implementation of the elligator map needs some methods of the packages which are internal in the original golang repo: https://github.com/golang/crypto/tree/master/ed25519/internal/edwards25519  compare to https://github.com/coniks-sys/libmerkleprefixtree-go/blob/vrf_impl_coname_agl/crypto/ed25519

c633:
About the `ed25519.go` source file, why don't we use current source code in the official `crypto/ed25519` pkg?

Liamsi:
> About the ed25519.go source file, why don't we use current source code in the official crypto/ed25519 pkg? 

Yes, maybe it's worth it to investigate the changes and use the official golang version. If I rememeber correctly I already started that and it didn't work out fine (can't remember the reason though)

Liamsi:
No, seem to work fine. Thanks @c633

masomel:
@Liamsi is this ready to be merged?

Liamsi:
> @Liamsi is this ready to be merged?

From my side yes.

masomel:
Cool. I'll handle the merge.

masomel:
@Liamsi One more question: is the vrf subpackage in crypto/vrf copyrighted to the coname authors?

Liamsi:
As far as I understood it, yes it is copyrighted by the coname authors as it is part of coname repo.
But some of the files are actually copies of golang's internal edwards25519 package (see https://github.com/coniks-sys/libmerkleprefixtree-go/pull/12#issuecomment-229041412). 
But I'm also not too familiar with copyright though. While talking to Andres and Daniel they definitely/endorsed the idea of using their implementation in our project. But I think we still would need to mention that in the license.

masomel:
Ok. I already included a notice about the ed25519 subpackage in our license, so we should be fine on that front. I'll add a note about the vrf subpackage in a similar fashion. Thanks!

Liamsi:
Thank you too for taking care of that!

Issue #18, opened 2016-07-15
Refactor testing code


masomel:
There seems to be a lot of repeated code used for testing (e.g. constructing a tree with 3 entries). It might be worth moving these repeated code segments into appropriate functions into a separate module?

cc @c633 @arlolra

Liamsi:
As I did some moving around of test code, I could take care of this. Does it make sense to start this soon? I would create a (possibly internal) `testutils` package where I would the methods that are used in different packages.

c633:
Sounds good to me!

c633:
- [ ] Tests in `auditlog_test.go` should be rewritten and integrated with the directory tests, after our APIs are complete.

masomel:
- [ ] Refactoring more common testing functionality, to create a simpler testing API, in which developers can directly pass test cases (see https://github.com/coniks-sys/coniks-go/pull/191#discussion_r147791806)

Issue #16, opened 2016-07-13
Stress test adding data to the tree


arlolra:
5ff85ad6ec6fd5214a9edc59232ca4e53a5d6199 is an indication that we should have a test that generates a large number of key/val pairs to set and retrieve from the tree.

c633:
That's right. I also had this idea last night. We should also add tests which involve all features of the library as well.

Liamsi:
As @arlolra pointed out by mail this is also about profiling:

>particularly automatically generated a lot of data to
stick in the tree and then profiling to see where the
bottlenecks may be / memory consumption.

Liamsi:
Here is a list of missing tests (per package):

merkletree
- pad.TB

crypto:
- Digest
- MakeRand

utils:
- ToBits

In general only the success cases are well tested. Branches where we `panic`/return an `error` or return a negative result on a check like `crypto.Verify` aren't covered by tests most of the time.

Just wanted to leave this here for discussion.

masomel:
> In general only the success cases are well tested. Branches where we panic/return an error or return a negative result on a check like crypto.Verify aren't covered by tests most of the time.

This seems like an issue to me. I think we should definitely be covering failure cases as well. This will be especially important if/when we create a common test suite for ``coniks-go`` and ``coniks-java`` (I think we discussed this in #25).

Another thing I wanted to open up for discussion was whether we should be using a code coverage  measurement tool like Coveralls or Codecov.io? These are the main ones I've seen, I'm sure there are more out there.

arlolra:
> Another thing I wanted to open up for discussion was whether we should be using a code coverage measurement tool like Coveralls or Codecov.io?

Couldn't hurt, and [goveralls](https://github.com/mattn/goveralls) seems pretty straightforward.

masomel:
> Couldn't hurt, and goveralls seems pretty straightforward.

Cool! I've authorized Coveralls to access the coniks-sys repos. I'll go set up goveralls.

Liamsi:
Good idea. Also thought about that. After the changes the code coverage is already quite high. I'm not necessarily a friend of aiming for 100% code coverage, though. But goveralls seems like a good idea anyways!

masomel:
So I've set up goveralls. I wasn't able to get it to work without testing each package and subpackage individually, so now we just have to remember to add each new package to the YAML file.

Liamsi:
[...]

PR #19, opened 2016-07-15
Vrf integration

### LINKED ISSUE #14: Use the VRF for computing the private index

Liamsi:
Use coname's vrf implementation to compute the private index
- Should resolve #14 
- commented out parts of a test which is supposed to work with a specific hash (but not with vrf); might be changed with #16 and #18
- replaced `t.Error` in combination with `return` with `t.Fatal`

### DIFF
--- a/merkletree/merkletree.go
+++ b/merkletree/merkletree.go
@@ -37,11 +37,10 @@ func NewMerkleTree() (*MerkleTree, error) {
 	return m, nil
 }
 
-func (m *MerkleTree) Get(key string) *AuthenticationPath {
-	lookupIndex := computePrivateIndex(key)
+func (m *MerkleTree) Get(lookupIndex []byte) *AuthenticationPath {
 	lookupIndexBits := util.ToBits(lookupIndex)
 	depth := 0
-	var nodePointer interface{}
+	var nodePointer MerkleNode
 	nodePointer = m.root
 
 	authPath := &AuthenticationPath{
@@ -92,8 +91,7 @@ func (m *MerkleTree) Get(key string) *AuthenticationPath {
 	panic(ErrorInvalidTree)
 }
 
-func (m *MerkleTree) Set(key string, value []byte) error {
-	index := computePrivateIndex(key)
+func (m *MerkleTree) Set(index []byte, key string, value []byte) error {
 
 	// generate random per user salt
 	salt := make([]byte, crypto.HashSizeByte)
@@ -113,16 +111,10 @@ func (m *MerkleTree) Set(key string, value []byte) error {
 	return nil
 }
 
-// Private Index calculation function
-// would be replaced with Ismail's VRF implementation
-func computePrivateIndex(key string) []byte {
-	return crypto.Digest([]byte(key))
-}
-
 func (m *MerkleTree) insertNode(index []byte, toAdd *userLeafNode) {
 	indexBits := util.ToBits(index)
 	depth := 0
-	var nodePointer interface{}
+	var nodePointer MerkleNode
 	nodePointer = m.root
 
 insertLoop:
@@ -189,6 +181,30 @@ insertLoop:
 		default:
 			panic(ErrorInvalidTree)
 		}
+	}
+}
+
+// visits all leaf-nodes and calls callBack on each of them
+// doesn't modify the underlying tree m
+func (m *MerkleTree) visitLeafNodes(callBack func(*userLeafNode)) {
+	visitULNsInternal(m.root, callBack)
+}
+
+func visitULNsInternal(nodePtr MerkleNode, callBack func(*userLeafNode)) {
+	switch nodePtr.(type) {
+	case *userLeafNode:
+		callBack(nodePtr.(*userLeafNode))
+	case *interiorNode:
+		if leftChild := nodePtr.(*interiorNode).leftChild; leftChild != nil {
+			visitULNsInternal(leftChild, callBack)
+		}
+		if rightChild := nodePtr.(*interiorNode).rightChild; rightChild != nil {
+			visitULNsInternal(rightChild, callBack)
+		}
+	case *emptyNode:
+		// do nothing
+	default:
+		panic(ErrorInvalidTree)
 	}
 }
 
--- a/merkletree/pad.go
+++ b/merkletree/pad.go
@@ -2,9 +2,11 @@ package merkletree
 
 import (
 	"crypto/rand"
+	"crypto/subtle"
 	"errors"
 
 	"github.com/coniks-sys/coniks-go/crypto"
+	"github.com/coniks-sys/coniks-go/crypto/vrf"
 )
 
 var (
@@ -15,10 +17,11 @@ var (
 // PAD is an acronym for persistent authenticated dictionary
 type PAD struct {
 	key          crypto.SigningKey
-	tree         *MerkleTree
+	tree         *MerkleTree // will be used to create the next STR
 	snapshots    map[uint64]*SignedTreeRoot
 	loadedEpochs []uint64 // slice of epochs in snapshots
-	currentSTR   *SignedTreeRoot
+	latestSTR    *SignedTreeRoot
+	policies     Policies // the current policies in place
 }
 
 // NewPAD creates new PAD consisting of an array of hash chain
@@ -34,36 +37,46 @@ func NewPAD(policies Policies, key crypto.SigningKey, len uint64) (*PAD, error) 
 	if err != nil {
 		return nil, err
 	}
+	pad.policies = policies
 	pad.snapshots = make(map[uint64]*SignedTreeRoot, len)
 	pad.loadedEpochs = make([]uint64, 0, len)
-	pad.updateInternal(policies, 0)
+	pad.updateInternal(nil, 0)
 	return pad, nil
 }
 
 // if policies is nil, the previous policies will be used
-func (pad *PAD) generateNextSTR(policies Policies, m *MerkleTree, epoch uint64) {
+func (pad *PAD) signTreeRoot(m *MerkleTree, epoch uint64) {
 	var prevStrHash []byte
-	if pad.currentSTR == nil {
+	if pad.latestSTR == nil {
 		prevStrHash = make([]byte, crypto.HashSizeByte)
 		if _, err := rand.Read(prevStrHash); err != nil {
 			// panic here since if there is an error, it will break the PAD.
 			panic(err)
 		}
 	} else {
-		prevStrHash = crypto.Digest(pad.currentSTR.sig)
-		if policies == nil {
-			policies = pad.currentSTR.policies
-		}
+		prevStrHash = crypto.Digest(pad.latestSTR.sig)
 	}
-	pad.currentSTR = NewSTR(pad.key, policies, m, epoch, prevStrHash)
+	pad.latestSTR = NewSTR(pad.key, pad.policies, m, epoch, prevStrHash)
 }
 
 func (pad *PAD) updateInternal(policies Policies, epoch uint64) {
 	pad.tree.recomputeHash()
 	m := pad.tree.Clone()
-	pad.generateNextSTR(policies, m, epoch)
-	pad.snapshots[epoch] = pad.currentSTR
+	// create STR with the policies that were actually used in the prev.
+	// Set() operation
+	pad.signTreeRoot(m, epoch)
+	pad.snapshots[epoch] = pad.latestSTR
 	pad.loadedEpochs = append(pad.loadedEpochs, epoch)
+
+	if policies != nil { // update the policies if necessary
+		vrfKeyChanged := 1 != (subtle.ConstantTimeCompare(
+			pad.policies.vrfPrivate()[:],
+			policies.vrfPrivate()[:]))
+		pad.policies = policies
+		if vrfKeyChanged {
+			pad.reshuffle()
+		}
+	}
 }
 
 func (pad *PAD) Update(policies Policies) {
@@ -76,16 +89,16 @@ func (pad *PAD) Update(policies Policies) {
 		pad.loadedEpochs = append(pad.loadedEpochs[:0], pad.loadedEpochs[n:]...)
 	}
 
-	pad.updateInternal(policies, pad.currentSTR.epoch+1)
+	pad.updateInternal(policies, pad.latestSTR.epoch+1)
 }
 
 func (pad *PAD) Set(key string, value []byte) error {
-	return pad.tree.Set(key, value)
+	index, _ := pad.computePrivateIndex(key, pad.policies.vrfPrivate())
+	return pad.tree.Set(index, key, value)
 }
 
-func (pad *PAD) Lookup(key string) *AuthenticationPath {
-	str := pad.currentSTR
-	return str.tree.Get(key)
+func (pad *PAD) Lookup(key string) (*AuthenticationPath, error) {
+	return pad.LookupInEpoch(key, pad.latestSTR.epoch)
 }
 
 func (pad *PAD) LookupInEpoch(key string, epoch uint64) (*AuthenticationPath, error) {
@@ -93,31 +106,57 @@ func (pad *PAD) LookupInEpoch(key string, epoch uint64) (*AuthenticationPath, er
 	if str == nil {
 		return nil, ErrorSTRNotFound
 	}
-	ap := str.tree.Get(key)
+	lookupIndex, proof := pad.computePrivateIndex(key, str.policies.vrfPrivate())
+	ap := str.tree.Get(lookupIndex)
+	ap.vrfProof = proof
 	return ap, nil
 }
 
 func (pad *PAD) GetSTR(epoch uint64) *SignedTreeRoot {
-	if epoch >= pad.currentSTR.epoch {
-		return pad.currentSTR
+	if epoch >= pad.latestSTR.epoch {
+		return pad.latestSTR
 	}
 	return pad.snapshots[epoch]
 }
 
 func (pad *PAD) TB(key string, value []byte) (*TemporaryBinding, error) {
-	//FIXME: compute private index twice
-	//it would be refactored after merging VRF integration branch
-	index := computePrivateIndex(key)
-	tb := pad.currentSTR.sig
+	str := pad.latestSTR
+	index, _ := pad.computePrivateIndex(key, pad.policies.vrfPrivate())
+	tb := str.sig
 	tb = append(tb, index...)
 	tb = append(tb, value...)
 	sig := crypto.Sign(pad.key, tb)
 
-	err := pad.Set(key, value)
+	err := pad.tree.Set(index, key, value)
 
 	return &TemporaryBinding{
 		index: index,
 		value: value,
 		sig:   sig,
 	}, err
+}
+
+// reshuffle recomputes indices of keys and store them with their values in new
+// tree with new new position;
+// swaps pad.tree if everything worked out.
+// if there is any error on the way (lack of entropy for randomness) reshuffle
+// will panic
+func (pad *PAD) reshuffle() {
+	newTree, err := NewMerkleTree()
+	if err != nil {
+		panic(err)
+	}
+	pad.tree.visitLeafNodes(func(n *userLeafNode) {
+		newIndex, _ := pad.computePrivateIndex(n.key, pad.policies.vrfPrivate())
+		if err := newTree.Set(newIndex, n.key, n.value); err != nil {
+			panic(err)
+		}
+	})
+	pad.tree = newTree
+}
+
+func (pad *PAD) computePrivateIndex(key string,
+	vrfPrivKey *[vrf.SecretKeySize]byte) (index, proof []byte) {
+	index, proof = vrf.Prove([]byte(key), vrfPrivKey)
+	return
 }
--- a/crypto/ed25519/extra25519/extra25519_test.go
+++ b/crypto/ed25519/extra25519/extra25519_test.go
@@ -22,7 +22,6 @@ func TestCurve25519Conversion(t *testing.T) {
 	var privBytes [64]byte
 	copy(privBytes[:], private)
 
-
 	var curve25519Public, curve25519Public2, curve25519Private [32]byte
 	PrivateKeyToCurve25519(&curve25519Private, &privBytes)
 	curve25519.ScalarBaseMult(&curve25519Public, &curve25519Private)
--- a/crypto/vrf/vrf.go
+++ b/crypto/vrf/vrf.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/coniks-sys/coniks-go/crypto/ed25519/edwards25519"
 	"github.com/coniks-sys/coniks-go/crypto/ed25519/extra25519"
+	"golang.org/x/crypto/ed25519"
 )
 
 const (
@@ -56,6 +57,10 @@ func GenerateKey(rnd io.Reader) (pk []byte, sk *[SecretKeySize]byte, err error) 
 
 	copy(sk[32:], pkBytes[:])
 	return pkBytes[:], sk, err
+}
+
+func Public(sk *[SecretKeySize]byte) []byte {
+	return ed25519.PrivateKey(sk[:]).Public().(ed25519.PublicKey)
 }
 
 func expandSecret(sk *[SecretKeySize]byte) (x, skhr *[32]byte) {
--- a/crypto/vrf/vrf_test.go
+++ b/crypto/vrf/vrf_test.go
@@ -29,6 +29,18 @@ func TestHonestComplete(t *testing.T) {
 	}
 }
 
+func TestConvertSecretKeyToPublicKey(t *testing.T) {
+	pk, sk, err := GenerateKey(nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	pkBytes := Public(sk)
+	if !bytes.Equal(pk, pkBytes) {
+		t.Fatal("Couldn't obtain public key.")
+	}
+}
+
 func TestFlipBitForgery(t *testing.T) {
 	pk, sk, err := GenerateKey(nil)
 	if err != nil {
--- a/merkletree/merkletree_test.go
+++ b/merkletree/merkletree_test.go
@@ -2,12 +2,18 @@ package merkletree
 
 import (
 	"bytes"
-	"reflect"
 	"testing"
 
+	"github.com/coniks-sys/coniks-go/crypto/vrf"
 	"github.com/coniks-sys/coniks-go/utils"
 	"golang.org/x/crypto/sha3"
 )
+
+var _, vrfPrivKey1, _ = vrf.GenerateKey(bytes.NewReader(
+	[]byte("deterministic tests need 256 bit")))
+
+var _, vrfPrivKey2, _ = vrf.GenerateKey(bytes.NewReader(
+	[]byte("deterministic tests need 32 byte")))
 
 func TestOneEntry(t *testing.T) {
 	m, err := NewMerkleTree()
@@ -20,13 +26,11 @@ func TestOneEntry(t *testing.T) {
 
 	key := "key"
 	val := []byte("value")
-
-	if err := m.Set(key, val); err != nil {
+	index := vrf.Compute([]byte(key), vrfPrivKey1)
+	if err := m.Set(index, key, val); err != nil {
 		t.Fatal(err)
 	}
 	m.recomputeHash()
-
-	index := computePrivateIndex(key)
 
 	// Check empty node hash
 	h := sha3.NewShake128()
@@ -41,7 +45,7 @@ func TestOneEntry(t *testing.T) {
 			"get", m.root.rightHash)
 	}
 
-	r := m.Get(key)
+	r := m.Get(index)
 	if r.Leaf().Value() == nil {
 		t.Error("Cannot find value of key:", key)
 		return
@@ -72,7 +76,7 @@ func TestOneEntry(t *testing.T) {
 			"get", m.root.leftHash)
 	}
 
-	r = m.Get("abc")
+	r = m.Get([]byte("abc"))
 	if r.Leaf().Value() != nil {
 		t.Error("Invalid look-up operation:", key)
 		return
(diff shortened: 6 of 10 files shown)


merkletree.go
```diff
@@ -81,7 +83,8 @@ func (m *MerkleTree) Get(key string) (MerkleNode, []ProofNode) {
 }
 
 func (m *MerkleTree) Set(key string, value []byte) error {
-	index := computePrivateIndex(key)
+	vrfPriv := m.policies.vrfPrivate()
```
c633: Shouldn't this be `vrfPrivKey` for consistency?

merkletree.go
```diff
@@ -101,10 +104,8 @@ func (m *MerkleTree) Set(key string, value []byte) error {
 	return nil
 }
 
-// Private Index calculation function
-// would be replaced with Ismail's VRF implementation
-func computePrivateIndex(key string) []byte {
-	return crypto.Digest([]byte(key))
+func computePrivateIndex(key string, vrfPrivKey *[vrf.SecretKeySize]byte) []byte {
+	return vrf.Compute([]byte(key), vrfPrivKey)
```
c633: How's about:

```
func (m *MerkleTree) computePrivateIndex(key string) []byte {
    vrfPrivKey := m.policies.vrfPrivate()
    return vrf.Compute([]byte(key), vrfPrivKey)
}
```

Liamsi:
There were quite some changes on master hence the merge conflicts. 
In the current, changed design, we would need to pass the key from the `PAD` (which now contains the policy via the `currentSTR`) to the places were `computePrivateIndex` is called. 

I wonder if it wasn't actually possible to remove the need of the actual `key` (which at the moment seems to be the actual username/"key-value") in the `userLeafNode` and directly call `merkleTree.Set(key, value)` with the actual precomputed private-index as a `key` instead of computing it inside of `merkleTree.Set(...)` via `computePrivateIndex`. 
This would make it easier to compute the index (at leaster since the recent changes). Also (and maybe more important) I thought the actual usernames were not necessary and even for privacy reasons shouldn't directly be stored in the userLeafNodes? @masomel @c633 @arlolra 

Hope I didn't mix up things here (and sorry for the many questions).

masomel:
> Also (and maybe more important) I thought the actual usernames were not necessary and even for privacy reasons shouldn't directly be stored in the userLeafNodes?

We left the actual usernames in the userLeafNodes since this allows the server to change ``vrfPrivateKey`` and reconstruct the whole tree without any user/client intervention. But we had a longer discussion about the privacy of usernames and keys in https://github.com/coniks-sys/libmerkleprefixtree-go/pull/1#discussion_r67771848. I hope this clears up some of your questions!

masomel:
I will add that you're definitely right about the actual usernames not being necessary to support lookups and auth path verifications. But I think supporting ``vrfPrivateKey`` changes on the server side helps protect users' privacy in a different way, as we discuss in §6.3 "Randomizing the order of directory entries" in the paper.

Liamsi:
Thanks @masomel! OK, makes sense. 

So we need to either 

1) store the (private-)key directly in the `MerkleTree` struct now 
or
2) to pass the vrf-key from the `PAD`s policy down to the call of `MerkleTree`'s `Set`/`Get`
or
3) Change the Set method's signature to 
```
// add index argument to be able to compute the index from outside/from the PAD
func (m *MerkleTree) Set(index []byte, key string, value []byte) error
``` 
and the Get signature to
```
// replace key argument with index
func (m *MerkleTree) Get(index []byte) *AuthenticationPath
```

I think 3) is the best option.

Liamsi:
With option 3) the vrf-computation would happen [here](https://github.com/coniks-sys/libmerkleprefixtree-go/blob/master/pad.go#L82-L89) instead of [here](https://github.com/coniks-sys/libmerkleprefixtree-go/blob/master/merkletree.go#L41) and [here](https://github.com/coniks-sys/libmerkleprefixtree-go/blob/master/merkletree.go#L96) right now.

c633:
> There were quite some changes on master hence the merge conflicts. 
In the current, changed design, we would need to pass the key from the PAD (which now contains the policy via the currentSTR) to the places were computePrivateIndex is called.

It is exactly what I was thinking! I think it makes sense to go with option 3), and computePrivateIndex should be a method of PAD.

masomel:
@Liamsi ce91df5 looks good to me!

pad.go
```diff
@@ -80,20 +81,23 @@ func (pad *PAD) Update(policies Policies) {
...
+	return str.tree.Get(lookupIndex)
 }
 
 func (pad *PAD) LookupInEpoch(key string, epoch uint64) (*AuthenticationPath, error) {
 	str := pad.GetSTR(epoch)
 	if str == nil {
 		return nil, ErrorSTRNotFound
 	}
-	ap := str.tree.Get(key)
+	lookupIndex := pad.computePrivateIndex(key)
```
arlolra: This doesn't look like it's using the right private key.  If possible, we should add a test for this case.

policy.go
```diff
@@ -34,6 +38,11 @@ func (p *DefaultPolicies) Serialize() []byte {
 	return bs
```
arlolra: Should the public key be part of the serialized policy?

arlolra:
The approach here looks good.  Just a few comments to address.

c633:
@Liamsi @arlolra @masomel  What do you think about 24085a3844fd97a4417bae3a80c391f84cc1a5b8?

pad.go
Liamsi: I'm not sure what you mean. Maybe the `key` variable name is a little misleading here? The (private-)key which is used to compute the priv. is this one: https://github.com/coniks-sys/libmerkleprefixtree-go/pull/19/files#diff-73fecc7207210fbe4ae4c1365f307927R112

arlolra: Look at the implementation of `computePrivateIndex`,

```
vrfPivKey := pad.currentSTR.policies.vrfPrivate()
```

meaning that it gets the private key from the current epoch.

But this is a lookup in a specific epoch (the `str := pad.GetSTR(epoch)`)

So, it seems like it needs to be,
```
str.policies.vrfPrivate()
```

And I'm assuming that the private key change in the policy means we need to recompute all the indices.

Liamsi: OK. I see. You're right. Thanks for the clarification! Will fix that (and add a test).

Liamsi: > And I'm assuming that the private key change in the policy means we need to recompute all the indices.

Will dig into that tomorrow.

masomel: > And I'm assuming that the private key change in the policy means we need to recompute all the indices.

Yes. This would be a good feature to expose to the developer as this also has the nice side-effect of re-randomizing the order of leaves in the tree (as discussed in https://github.com/coniks-sys/libmerkleprefixtree-go/pull/19#issuecomment-233067359).

Liamsi:
TODOs (for myself)

- [ ] use correct private-key: https://github.com/coniks-sys/libmerkleprefixtree-go/pull/19#discussion_r71786406
- [ ] resolve serialization: https://github.com/coniks-sys/libmerkleprefixtree-go/pull/19#discussion_r71641381

policy.go
masomel: It should so that clients can obtain it to verify private indices.

Liamsi:
> @Liamsi @arlolra @masomel What do you think about 24085a3?

I think that's useful/helpful. Thanks a lot for chiming in @c633! I added some comments. Let me know what you think (if you want to keep as is or consider some of the minor changes). I'll merge it into this branch, OK?

c633:
I agree with all your comments!

pad.go
c633: > But this is a lookup in a specific epoch (the str := pad.GetSTR(epoch))
So, it seems like it needs to be, str.policies.vrfPrivate()

I think it makes sense to make `computePrivateIndex` a method of `STR`, thoughts?

Liamsi: Makes sense.

Liamsi:
> I agree with all your comments!

OK. I'll merge your branch and do the changes.

merkletree/pad.go
```diff
@@ -114,8 +114,9 @@ func (pad *PAD) GetSTR(epoch uint64) *SignedTreeRoot {
 func (pad *PAD) TB(key string, value []byte) (*TemporaryBinding, error) {
 	//FIXME: compute private index twice
 	//it would be refactored after merging VRF integration branch
-	index, _ := pad.computePrivateIndex(key)
-	tb := pad.currentSTR.sig
+	cStr := pad.currentSTR
```
arlolra: Help me think this through, because this looks problematic.

Currently, when you call `pad.Update(policies)` it generates the next STR as if the passed in `policies` were the ones being used.  But, if the private key changes, all these temporary bindings being handed out no longer have the right index.

I think `PAD` needs a `policies` field reflecting the current policies in place. `currentSTR` may be badly named in that it is really just the most recently signed tree root, whereas `pad.tree` is what's current and will be reflected in the next STR.  Then, when you call `Update(policies)`, it should generate the STR with the policies that were actually used (`pad.policies`), and then swap the `pad.policies` with the ones being passed in to be used going forward.  It can then reshuffle the `pad.tree` if the private key changed.

Hopefully that makes sense.

Liamsi: > Currently, when you call pad.Update(policies) it generates the next STR as if the passed in policies were the ones being used. But, if the private key changes, all these temporary bindings being handed out no longer have the right index.

Makes sense. Will change that.

> I think PAD needs a policies field reflecting the current policies in place. currentSTR may 

OK, I have to think about that (after the talks/program finished). But on first sight it makes sense to add a policies field to `PAD`.  

> currentSTR may be badly named in that it is really just the most recently signed tree root, whereas pad.tree is what's current and will be reflected in the next STR. 

What do you think about `lastSTR` or `latestSTR`?

> Then, when you call Update(policies), it should generate the STR with the policies that were actually used (pad.policies), and then swap the pad.policies with the ones being passed in to be used going forward. It can then reshuffle the pad.tree if the private key changed.

> Hopefully that makes sense.

Sounds reasonable.

c633: Totally makes sense, it's a very good point. Thanks!

arlolra: > What do you think about lastSTR or latestSTR?

`latestSTR` sounds ok.  While we're at it, maybe do `s/generateNextSTR/SignTreeRoot/` as well.

Liamsi: OK, thanks.

Liamsi: >>I think PAD needs a policies field reflecting the current policies in place. currentSTR may

>OK, I have to think about that (hopefully after the talks/program finished). But on first sight it makes sense to add a policies field to PAD.

Didn't have the time to think it through (yet). But do we still need the policies field in `STR` than (for serializing, or sth else)?

arlolra: > But do we still need the policies field in STR then (for serializing, or sth else)?

I think we do.  Anything that gets signed needs to remain so that it can be verified later.  But also, in this case, the when you do the lookup from a previous epoch, you still need to use the key from the old policy.

So, I should have mentioned, `computePrivateIndex` should probably stay on the `pad` but we'll need a `computePrivateIndexInEpoch` as well.  Or, something like that.  You get the idea.

merkletree/pad.go
```diff
@@ -80,20 +80,27 @@ func (pad *PAD) Update(policies Policies) {
...
 
 func (pad *PAD) Set(key string, value []byte) error {
-	return pad.tree.Set(key, value)
+	cStr := pad.currentSTR
+	index, _ := cStr.computePrivateIndex(key)
+	return pad.tree.Set(index, key, value)
 }
 
 func (pad *PAD) Lookup(key string) *AuthenticationPath {
 	str := pad.currentSTR
```
arlolra: It might be simpler here to call out to,

```
return LookupInEpoch(key, pad.latestSTR.epoch);
```

merkletree/pad.go
Liamsi: Thanks!

merkletree/pad.go
```diff
@@ -76,39 +77,43 @@ func (pad *PAD) Update(policies Policies) {
...
 	return pad.snapshots[epoch]
 }
 
 func (pad *PAD) TB(key string, value []byte) (*TemporaryBinding, error) {
 	//FIXME: compute private index twice
 	//it would be refactored after merging VRF integration branch
-	index := computePrivateIndex(key)
-	tb := pad.currentSTR.sig
+	cStr := pad.latestSTR
+	index, _ := cStr.computePrivateIndex(key)
```
c633: Can you help me change something here? 
`pad.Set(key, value)` below to `pad.tree.Set(index, key, value)` and remove the comment above. Many thanks!

Liamsi: Sure, but I'll only be able to do it later or maybe even tomorrow morning.

Am 22.07.2016 6:55 nachm. schrieb "Vu Quoc Huy" <notifications@github.com>:

merkletree/pad.go
```diff
@@ -41,29 +45,39 @@ func NewPAD(policies Policies, key crypto.SigningKey, len int64) (*PAD, error) {
 }
 
 // if policies is nil, the previous policies will be used
-func (pad *PAD) generateNextSTR(policies Policies, m *MerkleTree, epoch uint64) {
+func (pad *PAD) SignTreeRoot(policies Policies, m *MerkleTree, epoch uint64) {
```
c633: Should we export this method?

Liamsi: Good catch. No, I don't think it's neccessary.

merkletree/pad.go
```diff
@@ -76,48 +93,65 @@ func (pad *PAD) Update(policies Policies) {
...
 	}
 
-	pad.updateInternal(policies, pad.currentSTR.epoch+1)
+	pad.updateInternal(policies, pad.latestSTR.epoch+1)
 }
 
 func (pad *PAD) Set(key string, value []byte) error {
-	return pad.tree.Set(key, value)
+	str := pad.latestSTR
+	if str == nil {
```
c633: In which case the `pad.latestSTR` is `nil`?

merkletree/pad.go
```diff
@@ -76,48 +93,65 @@ func (pad *PAD) Update(policies Policies) {
...
 	return pad.snapshots[epoch]
 }
 
 func (pad *PAD) TB(key string, value []byte) (*TemporaryBinding, error) {
-	//FIXME: compute private index twice
-	//it would be refactored after merging VRF integration branch
-	index := computePrivateIndex(key)
-	tb := pad.currentSTR.sig
+	str := pad.latestSTR
+	if str == nil {
```
c633: same question as above.

merkletree/pad.go
```diff
@@ -76,48 +93,65 @@ func (pad *PAD) Update(policies Policies) {
...
+	return pad.LookupInEpoch(key, pad.latestSTR.epoch)
 }
 
 func (pad *PAD) LookupInEpoch(key string, epoch uint64) (*AuthenticationPath, error) {
 	str := pad.GetSTR(epoch)
 	if str == nil {
 		return nil, ErrorSTRNotFound
 	}
-	ap := str.tree.Get(key)
+	lookupIndex, proof:= pad.computePrivateIndexInEpoch(key, str.epoch)
```
c633: The `GetSTR` is called twice here. So I think it's maybe better if we move the `computePrivateIndex` into `STR` instead. What do you think?

merkletree/pad.go
Liamsi: Good question. I saw there was a check for that after calling `Update()`here https://github.com/coniks-sys/coniks-go/pull/19/files/e82f666f1930a475550a8c3eb00194932c003a19#diff-475ca4a45242c730e6134d970fe64530L46 and wanted to be consistent here. If no combination of calling public methods can lead to `pad.latestSTR == nil` I'll remove it.

merkletree/pad.go
Liamsi: Thought about that, too. It makes sense: currently `computePrivateIndexInEpoch` is always called with an `str` in hand.

Liamsi: An alternative could be to directly pass the str (and keep `computePrivateIndex` in `PAD`).

merkletree/pad.go
Liamsi: > also, in this case, when you do the lookup from a previous epoch, you still need to use the key from the old policy.

I think this is done in e82f666f1930a475550a8c3eb00194932c003a19 

> Then, when you call Update(policies), it should generate the STR with the policies that were actually used (pad.policies), and then swap the pad.policies with the ones being passed in to be used going forward. It can then reshuffle the pad.tree if the private key changed.

Besides the reshuffling part this is also done in the latest commit. Please review. I have some questions regarding the design of the reshuffling (below).

merkletree/pad.go
c633: I see. This check is performed when the `PAD` is initialized, and it should be true once only when `NewPAD` is called. I presume that no later calls can lead to `pad.latestSTR == nil`.

Liamsi: Thanks!

merkletree/pad.go
c633: > currently computePrivateIndexInEpoch is always called with an str in hand.

I understand the reason why @arlolra  and you want to keep `computePrivateIndex` in `PAD` now. Actually, we have to eventually look through our persistent storage to see if we have an appropriate STR. I'm working on this, in the `db` branch. 

So I think it makes sense to keep `computePrivateIndex` in `PAD`, and `panic` if passed `str` is `nil`.

Liamsi:
Currently, the reshuffling of the `pad.tree` is still missing. So after a (vrf) key-change `LookUpInEpoch` will succeed (as it uses the index of the `str` of that epoch) but a plain `LookUp` will fail if the key was inserted before (or immediately after) a vrf-key-change (as it will use the wrong vrf-key/index)

To make this work @arlolra suggested a re-computation of the indices (https://github.com/coniks-sys/coniks-go/pull/19#discussion_r71883289). As far as I understand it the only way to re-shuffle the pad.tree (from outside the package `merkletree`) is to actually know all the `key`s in order to re-compute their indices, right? This would mean to have a way to get all keys from one MerkleTree.
If we wanted to re-compute the tree from within the `merkletree` package (where one could get all keys for instance by traversing the tree), we would need to pass the vrf-private-key to the `MerkleTree` (so that it can recompute its indices without exposing the keys for which the re-computation of indices will happen)

Liamsi:
I thought a little about my above question: I think exposing the 'key's from the `tree` isn't the way to go, as it's to easy (as an user of this library) to get it wrong (and for instance expose the `key`s somehow or misuse it in another way).
The other option (give the tree the vrf-key/the vrf-function) also comes with downsides: The clear separation between the tree and the pad gets a little less clear (both compute the private indices for keys now). 
The easiest way to implement the latter would be to store the keys in a private array and iterate over them to get their values and to insert both in a fresh tree with the new index.

c633:
> I thought a little about my above question: I think exposing the 'key's from the tree isn't the way to go, as it's to easy (as an user of this library) to get it wrong (and for instance expose the keys somehow or misuse it in another way).

It seems I didn't understand your idea completely. Since the tree and pad are in the same package (`merkletree`), I don't know how the developers can get "exposing keys" if we don't export them?  

> The easiest way to implement the latter would be to store the keys in a private array and iterate over them to get their values and to insert both in a fresh tree with the new index.

I prefer this way.

c633:
>The easiest way to implement the latter would be to store the keys in a private array and iterate over them to get their values (with the old index) and to insert both in a fresh tree with the new index.

I thought a little about this implementation. There are some concerns here:

- If we store the keys in a private array, I presumably we should remove the `key` field in `userLeafNode` struct. 
- We also need to store the 'value's along with the keys. Back to https://github.com/coniks-sys/coniks-go/pull/1#discussion_r67771848, @masomel said that "(We do have a protocol for also keeping the value private from the server, though it's not the default, so I'm not sure if it should be built into the tree design)". Should we use that protocol? 
- Should we separate this into another issue, and merge current changes first?

arlolra:
> To make this work @arlolra suggested a re-computation of the indices (#19 (comment)). As far as I understand it the only way to re-shuffle the pad.tree (from outside the package merkletree) is to actually know all the keys in order to re-compute their indices, right? This would mean to have a way to get all keys from one MerkleTree.

Could you add a method of MerkleTree,

```
WalkTree(m *MerkleTree) (call func(*node)) {
   // visit all the leaf nodes and call func
}
```

Then  you could,

```
ReShuffle(pad *PAD) {
  newTree := NewMerkleTree()
  pad.tree.WalkTree(func(node) {
     key = node.key
     newIndex = pad.computePrivateKey(key)
     newTree.Set(key, node.value)
  })
  pad.tree = newTree
}
```

merkletree/pad.go
```diff
@@ -76,48 +93,65 @@ func (pad *PAD) Update(policies Policies) {
...
+	pad.updateInternal(policies, pad.latestSTR.epoch+1)
 }
 
 func (pad *PAD) Set(key string, value []byte) error {
-	return pad.tree.Set(key, value)
+	str := pad.latestSTR
+	if str == nil {
+		return ErrorSTRNotFound
+	}
+	index, _ := pad.computePrivateIndexInEpoch(key, str.epoch)
```
arlolra: This needs to use `pad.policies`, not the `latestSTR`.

There are two distinct times that we need to compute a private index,

* When setting a key.  This happens before a root is signed, and therefore cannot get the key from an STR.
* When looking up a key.  This always happens from a STR, and therefore uses the key from the policy in the STR.

So, it's fine to make `computePrivateIndex` a method of STR, but then you'll also need another method to handle the Set case.

Or, you can make it a method of PAD but one of the arguments needs to be a key.

Liamsi:
> It seems I didn't understand your idea completely. Since the tree and pad are in the same package (merkletree), I don't know how the developers can get "exposing keys" if we don't export them?

You are absolutely right. I actually confused the packages while thinking about it ... Thanks for your feedback and sorry for the noise!

>> The easiest way to implement the latter would be to store the keys in a private array and iterate over them to get their values and to insert both in a fresh tree with the new index.

> I prefer this way.

OK, traversing the tree (as @arlolra suggested in https://github.com/coniks-sys/coniks-go/pull/19#issuecomment-234790778) maybe more efficient and it certainly makes storing the keys twice obsolete. I think, I'll go this way now.

merkletree/pad.go
Liamsi: Thanks again for your feedback! Changed in https://github.com/coniks-sys/coniks-go/pull/19/commits/ad750faa6163f71e18e46d51c516124291503768

Liamsi:
> We also need to store the 'value's along with the keys. Back to #1 (comment), @masomel said that "(We do have a protocol for also keeping the value private from the server, though it's not the default, so I'm not sure if it should be built into the tree design)". Should we use that protocol?

@c633 @masomel depending on the protocol and its complexity this would be a nice (optional?) addition

> Should we separate this into another issue, and merge current changes first?

I think the above should be a separate issue. Regarding the recomputation of the tree/indices I think it's fine to keep it in this issue (as it is definitely related to the vrf-integration, also I feel all open questions are resolved now). On the other hand we thought of this as changing a few lines only, yo if you want to keep it in small separate issues I wouldn't vote against it.

merkletree/pad.go
```diff
@@ -41,29 +45,42 @@ func NewPAD(policies Policies, key crypto.SigningKey, len int64) (*PAD, error) {
 }
 
 // if policies is nil, the previous policies will be used
-func (pad *PAD) generateNextSTR(policies Policies, m *MerkleTree, epoch uint64) {
+func (pad *PAD) signTreeRoot(policies Policies, m *MerkleTree, epoch uint64) {
```
arlolra: The `policies` argument here is unnecessary, since it's always going to be `pad.policies`, which the `pad` has access to.

merkletree/pad.go
```diff
@@ -76,48 +93,60 @@ func (pad *PAD) Update(policies Policies) {
...
-	return str.tree.Get(key)
+func (pad *PAD) Lookup(key string) (*AuthenticationPath, error) {
+	return pad.LookupInEpoch(key, pad.latestSTR.epoch)
 }
 
 func (pad *PAD) LookupInEpoch(key string, epoch uint64) (*AuthenticationPath, error) {
 	str := pad.GetSTR(epoch)
 	if str == nil {
-		return nil, ErrorSTRNotFound
+		panic(ErrorSTRNotFound)
```
arlolra: If you're going to panic here, the return error is always nil now. But why make this change?

merkletree/pad.go
```diff
@@ -76,48 +93,60 @@ func (pad *PAD) Update(policies Policies) {
...
 func (pad *PAD) TB(key string, value []byte) (*TemporaryBinding, error) {
-	//FIXME: compute private index twice
-	//it would be refactored after merging VRF integration branch
-	index := computePrivateIndex(key)
-	tb := pad.currentSTR.sig
+	str := pad.latestSTR
+	if str == nil {
+		panic(ErrorSTRNotFound)
+	}
+	index, _ := pad.computePrivateIndex(key, str.policies.vrfPrivate())
```
arlolra: No, `TB()` is the same as `Set`.  You should be using the `pad.policies` here.

merkletree/pad.go
arlolra: Hmm, this makes me think that the `seed` should be generated at `PAD` construction time, since it's needed right away for the temporary bindings.

merkletree/pad.go
Liamsi: Done. Thanks!

merkletree/pad.go
Liamsi: True, it doesn't make sense to panic here. Changed back to returning the error.

merkletree/pad.go
```diff
@@ -41,29 +45,41 @@ func NewPAD(policies Policies, key crypto.SigningKey, len int64) (*PAD, error) {
...
+	// create STR with the policies that were actually used in the prev.
+	// Set() operation
+	pad.signTreeRoot(m, epoch)
+	pad.snapshots[epoch] = pad.latestSTR
 	pad.loadedEpochs = append(pad.loadedEpochs, epoch)
+
+	if policies != nil { // update the policies if necessary
+		// TODO use constant time compare from "crypto/subtle" instead (?)
+		vrfKeyChanged := !bytes.Equal(pad.policies.vrfPrivate()[:],
+			policies.vrfPrivate()[:])
```
Liamsi: Any thoughts on this? /cc @c633 @arlolra  @masomel

Liamsi:
I think this ready for another review. Sorry, that this took longer than expected... And thanks everyone for all the valuable feedback so far.

merkletree/pad.go
```diff
@@ -31,6 +34,7 @@ func NewPAD(policies Policies, key crypto.SigningKey, len int64) (*PAD, error) {
 	pad := new(PAD)
 	pad.key = key
 	pad.tree, err = NewMerkleTree()
+	pad.policies = policies
 	if err != nil {
 		return nil, err
 	}
```
arlolra: Can just do `pad.updateInternal(nil, 0)` on L43

c633:
> Regarding the recomputation of the tree/indices I think it's fine to keep it in this issue (as it is definitely related to the vrf-integration, also I feel all open questions are resolved now).

Never mind. I'm okay with this.

merkletree/pad.go
c633: Unless @masomel and @arlolra raise any problems about it, I think it's fine. But if we use constant time compare, it also give a sense of security and it doesn't hurt anybody.

merkletree/merkletree.go
```diff
@@ -192,6 +184,30 @@ insertLoop:
 	}
 }
 
+// visits all leaf-nodes and calls callBack on each of them
+// doesn't modify the underlying tree m
+func (m *MerkleTree) visitLeafNodes(callBack func(*userLeafNode)) {
+	visitULNsInternal(m.root, callBack)
+}
+
+func visitULNsInternal(nodePtr interface{}, callBack func(*userLeafNode)) {
```
c633: I would change `nodePtr interface{}` into `nodePtr MerkleNode`. I also want to change L43 and L117, too. So if it makes sense to you ...

merkletree/pad.go
```diff
@@ -31,6 +34,7 @@ func NewPAD(policies Policies, key crypto.SigningKey, len int64) (*PAD, error) {
 	pad := new(PAD)
 	pad.key = key
 	pad.tree, err = NewMerkleTree()
+	pad.policies = policies
```
c633: Should we swap L37 and L36?

merkletree/pad.go
```diff
@@ -76,48 +92,77 @@ func (pad *PAD) Update(policies Policies) {
...
 }
 
 func (pad *PAD) TB(key string, value []byte) (*TemporaryBinding, error) {
-	//FIXME: compute private index twice
-	//it would be refactored after merging VRF integration branch
-	index := computePrivateIndex(key)
-	tb := pad.currentSTR.sig
+	str := pad.latestSTR
+	if str == nil {
+		panic(ErrorSTRNotFound)
```
c633: We have discussed about `pad.latestSTR == nil` in https://github.com/coniks-sys/coniks-go/pull/19#discussion_r71984425. I think this check would be removed.

merkletree/pad.go
```diff
@@ -76,48 +92,77 @@ func (pad *PAD) Update(policies Policies) {
...
+func (pad *PAD) reshuffle() error {
+	var err error
+	newTree, err := NewMerkleTree()
+	if err != nil {
+		return err
+	}
+	pad.tree.visitLeafNodes(func(n *userLeafNode) {
+		newIndex, _ := pad.computePrivateIndex(n.key, pad.policies.vrfPrivate())
+		if e := newTree.Set(newIndex, n.key, n.value); e != nil {
+			// or should we panic here?
```
c633: I think we should panic here, since we may have many same errors returned by `Set()`. In that case, we probably want to have some ways to recovery it ASAP. Hope it makes sense to you.

Liamsi: The reason I decided against it was that the `reshuffle` method doesn't change anything in the pad/tree (if there was an error). This only happens outside. But I'll change it. It also makes sense to fail/panic earlier here.

Liamsi: The same holds for https://github.com/coniks-sys/coniks-go/pull/19/files/4de96a8c4cadd0d4e971c476e514d1351d849890#diff-475ca4a45242c730e6134d970fe64530R149 I guess.

merkletree/merkletree.go
Liamsi: Makes sense! I'll change the other lines for consistency.

Liamsi:
Incorporated all comments (so far) here https://github.com/coniks-sys/coniks-go/pull/19/commits/77a868dbced4ce08dd25de50efe39948be372f02 and changed to crypto/subtle for the pivate-key comparison here https://github.com/coniks-sys/coniks-go/pull/19/commits/a2cf95783cf9f68a29b7990a0190b87e98e91dd7

merkletree/merkletree.go
c633: Yeah, thanks!

merkletree/pad.go
c633: Yes. Totally agree!

c633:
Thanks! It looks great.

masomel:
> OK, traversing the tree (as @arlolra suggested in #19 (comment)) maybe more efficient and it certainly makes storing the keys twice obsolete. I think, I'll go this way now.

Seems fine with me.

> @c633 @masomel depending on the protocol and its complexity this would be a nice (optional?) addition

Our idea was that users seeking to keep their key(s) or other data private from the server as well as other users would encrypt the ``value`` blob in the ``userLeafNode`` with a symmetric key that only those clients that the user authorizes (i.e. gives the symmetric key to) can decrypt. This way, the ``value`` still looks like some random blob and it's up to the authorized clients to parse this blob correctly. The main downside I see in this protocol is how the clients distribute this symmetric key and how they revoke this key whenever they need to revoke some client's access to the ``value``. But I think if we wanted to add this protocol as an additional privacy feature, it would probably have to be on the client side anyway. 

This is a protocol we sketched out as part of a small follow-on project, and we never tested it out or anything, so I'm open to improvements/suggestions. Either way, I don't think that adding this protocol as part of this PR makes sense right now.

arlolra:
Nice work @Liamsi

PR #21, opened 2016-07-15
Temporary Binding implementation

### LINKED ISSUE #13: Temporary Bindings implementation
@masomel said:
> I'm thinking it might be better to have a separate TB struct, because the server needs to keep track of all temporary bindings issued during a given epoch, and return a TB whenever a client makes a lookup for a name which may not be in the tree yet but which has been registered/changed.
I think the functionality of keeping track can probably be implemented in the server itself, but I think it would facilitate handling TBs if they had their own struct.

This issues would be a part of #3

c633:
Implement the TB struct for registration protocol. 

* Part of #13 

* Currently TB implementation is being used in https://github.com/coniks-sys/libconiks-server-go/blob/registration/coniks/handler.go#L45-L47

### DIFF
--- a/pad.go
+++ b/pad.go
@@ -103,3 +103,21 @@ func (pad *PAD) GetSTR(epoch uint64) *SignedTreeRoot {
 	}
 	return pad.snapshots[epoch]
 }
+
+func (pad *PAD) TB(key string, value []byte) (*TemporaryBinding, error) {
+	//FIXME: compute private index twice
+	//it would be refactored after merging VRF integration branch
+	index := computePrivateIndex(key)
+	tb := pad.currentSTR.sig
+	tb = append(tb, index...)
+	tb = append(tb, value...)
+	sig := crypto.Sign(pad.key, tb)
+
+	err := pad.Set(key, value)
+
+	return &TemporaryBinding{
+		index: index,
+		value: value,
+		sig:   sig,
+	}, err
+}
--- a/tb.go
+++ b/tb.go
@@ -1,0 +1,19 @@
+package merkletree
+
+type TemporaryBinding struct {
+	index []byte
+	value []byte
+	sig   []byte
+}
+
+func (tb *TemporaryBinding) Index() []byte {
+	return tb.index
+}
+
+func (tb *TemporaryBinding) Value() []byte {
+	return tb.value
+}
+
+func (tb *TemporaryBinding) Signature() []byte {
+	return tb.sig
+}


masomel:
LGTM! @arlolra @Liamsi Thoughts?

tb.go
```diff
@@ -0,0 +1,36 @@
...
+type TemporaryBinding struct {
+	index []byte
+	value []byte
+	sig   []byte
+}
+
+func NewTemporaryBinding(pad *PAD, key string, value []byte) *TemporaryBinding {
+	str := pad.currentSTR
+	index := computePrivateIndex(key)
+	tb := str.SerializeWithSignature()
```
arlolra: This is a similar comment to what I had in https://github.com/coniks-sys/libmerkleprefixtree-go/pull/8#discussion_r70639337

In §3.3, `STR` is defined as,

> STR = SignSKd (t||tprev||roott ||H(STRprev)||P)

which to me look like it's just the signature.  I know I'm picking at nits, but now when I go look at the definition of `TB` in §4.1.1,

> TB = SignKd (STRt, i, k)

(which, in an aside to @masomel, probably wants to be, `TB = SignSKd (STRt || i || k)`, consistent with SKd and concatenation)

it's again not clear what is meant by `STR`.

arlolra:
I feel like `pad.Set` should be issuing the temporary bindings and storing them so that `pad.Get` can return them when necessary.  You can also write tests for that, which I don't see in the current implementation.

tb.go
masomel: I see your confusion now. I went back to look at Certificate Transparency's [Signed Tree Head spec](https://tools.ietf.org/html/rfc6962#section-3.5), after which STRs are modeled, and I think technically, the STR should be only the signature. 

What this means for ``prevStrHash`` and also ``TB`` (which yes, the notation should really use concatenation for consistency, though I believe commas are more proper crypto notation), is that these should only use ``STR.signature`` in their respective computations.

However, when TBs and STRs are sent to the client (and eventually to auditors in STRs' case), we will still have to include all the values to allow for signature verification.

masomel:
> I feel like ``pad.Set`` should be issuing the temporary bindings and storing them so that ``pad.Get`` can return them when necessary.

I had the same thought, but decided it would be best to leave it up to the developer to use TBs in the server implementation. The reason is that we've found that temporary bindings may not actually be the best approach to solving the immediate issuance problem, as we've been discussing in #13 and as we've learned from other contacts in industry (all of which are looking to move away from TBs). So we should definitely continue refining our solution for TBs, and I think it's fine to provide the TB interface for those developers who are willing to use TBs, but I don't think the ``PAD`` should issue temporary bindings by default.

tb.go
arlolra: Thanks for the clarification.  I opened #22 for `prevStrHash`.

arlolra:
> I think it's fine to provide the TB interface for those developers who are willing to use TBs, but I don't think the PAD should issue temporary bindings by default.

Ok, I'm on board with that.  I guess I'm most uncomfortable with `NewTemporaryBinding` taking the `PAD` as input.  I'd rather see a `pad.TB()` which returns the new binding, for those that wish to use it.

masomel:
> I guess I'm most uncomfortable with ``NewTemporaryBinding`` taking the ``PAD`` as input. I'd rather see a ``pad.TB()`` which returns the new binding, for those that wish to use it. 

This makes sense to me.

c633:
> Ok, I'm on board with that. I guess I'm most uncomfortable with NewTemporaryBinding taking the PAD as input. I'd rather see a pad.TB() which returns the new binding, for those that wish to use it.

The reason I did it is I thought it would be more consistent with other `NewXxx` functions in other source files. But it's a bad idea. 
I also changed the TB to `str.sig` in 60ee33f3c7e6adfd01374e83e6d394ca826f2aa5. The `verify` function would be implemented in a separate pull as #24

tb.go
```diff
@@ -0,0 +1,35 @@
...
+
+import "github.com/coniks-sys/libmerkleprefixtree-go/crypto"
+
+type TemporaryBinding struct {
+	index []byte
+	value []byte
+	sig   []byte
+}
+
+func (pad *PAD) NewTemporaryBinding(key string, value []byte) *TemporaryBinding {
```
arlolra: Can we move this to pad.go though, to be closer to `PAD`'s other methods?  And maybe rename it to just `.TB()`.

Liamsi:
I approve of everything @arlolra commented on. Especially of making `NewTemporaryBinding` a method of `PAD` and calling it `TB` or `TempBinding`. 

Also I'm not sure if we'll need all the getters (`Index()`, `Value()`, ...). We could remove them for now.

c633:
> Also I'm not sure if we'll need all the getters (Index(), Value(), ...). We could remove them for now.

I think we still need these getters since we need to expose these values to the client to allow them verify the signature. Does it make sense?

pad.go
```diff
@@ -103,3 +103,17 @@ func (pad *PAD) GetSTR(epoch uint64) *SignedTreeRoot {
...
 	return pad.snapshots[epoch]
 }
+
+func (pad *PAD) TB(key string, value []byte) *TemporaryBinding {
+	index := computePrivateIndex(key)
+	tb := pad.currentSTR.sig
+	tb = append(tb, index...)
+	tb = append(tb, value...)
+	sig := crypto.Sign(pad.key, tb)
+
```
arlolra: Should we also call `pad.Set(key, value)` in here?  If we're issuing a TB, it's kind of implying that, no?

c633: Yes, I couldn't find out any other reasons to not to do it, so we should do it.

PR #29, opened 2016-07-22
Implement registration function

### LINKED ISSUE #28: Implement registration
Moving from [libconiks-server-go](https://github.com/coniks-sys/libconiks-server-go/) repo.

For the registration protocol, the server needs to handle registration requests, insert the new name-to-key mappings into the tree, and return a temporary binding to the client.

c633:
* Should resolve #28


main.go
```diff
@@ -0,0 +1,28 @@
+package main
```
arlolra: This file should be in `server/`, no?

c633: My intention is it just a sample usage of the library.

arlolra: Ok, then please rename it to `main.example.go` or some such, and add a comment at the top of the file indicating the intention.

c633: Yes!

arlolra:
Another general comment is that the main repository README should reflect the addition of the `server` package.

main.go
arlolra: I still think it should be in `server/` though, since it's a sample usage of this package.

Also, the `server/README.md` seems to describe it that way.

arlolra: I guess the other question I have is why do you think this should only be an example?  This package (`server/`) is implementing a service that should compile to an executable.  The reusable components have already been packaged elsewhere.

Stated another way, why are we asking the ops team to implement main.go?  If anything, I would double down on this and move `-mkkey` and `-mkconfig` as flags here.

server/main.example.go
```diff
@@ -0,0 +1,26 @@
+package server
```
arlolra: See https://github.com/coniks-sys/coniks-go/pull/29#discussion_r71908317

main.go
masomel: I thought the server package was meant to contain the reusable components of the server, and that the server executable was going to be built elsewhere. But I do like the idea of providing an executable as well. 

Maybe we should separate these components out a bit more? E.g. we could move all reusable protocol implementation code to a separate package (maybe ``protocol``), and keep the server package for the reference implementation?

README.md
```diff
@@ -20,6 +20,7 @@ The packages in this library implement the various components of the CONIKS syst
 
 - ``crypto``: Cryptographic algorithms and operations
 - ``merkletree``: Merkle prefix tree and related data structures
+- ``server``: CONIKS library for the server side
```
masomel: I would reword this description to "Server-side protocol implementation"

main.go
c633: > I guess the other question I have is why do you think this should only be an example? This package (server/) is implementing a service that should compile to an executable. The reusable components have already been packaged elsewhere.

I added the description into the README. Maybe I'm misunderstanding here. I thought that the `server` package is still a CONIKS library which implements the registration, key lookup functions... Then the developers can add some extra layers on the top of the library, e.g. storage backend which stores data that not in the tree, or something else ... Or, they can choose to whether they want to use the registration bots library. 

> Stated another way, why are we asking the ops team to implement main.go? If anything, I would double down on this and move -mkkey and -mkconfig as flags here.

Yes. If it is an executable, I will do it.

c633:
> Another general comment is that the main repository README should reflect the addition of the server package.

Added. Thanks!

main.go
c633: > I thought the server package was meant to contain the reusable components of the server, and that the server executable was going to be built elsewhere. But I do like the idea of providing an executable as well.

I have the same thought here. 

> Maybe we should separate these components out a bit more? E.g. we could move all reusable protocol implementation code to a separate package (maybe protocol), and keep the server package for the reference implementation?

I'm also thinking about it. `protocol` probably includes source code in `handler.go`, `message.go`, and the others go to `server` package.

arlolra: > we could move all reusable protocol implementation code to a separate package (maybe protocol), and keep the server package for the reference implementation?

Yup.  From what's here, it seems like `server/message.go` and `server/serialization.go` can move to `protocol/` since I imagine they'd be used by both `client/` and `server/`.  But correct me if I'm wrong.

> Maybe I'm misunderstanding here. I thought that the server package is still a CONIKS library which implements the registration, key lookup functions... Then the developers can add some extra layers on the top of the library, e.g. storage backend which stores data that not in the tree, or something else ...

Maybe; then again, maybe not.  The storage backend just seems like a configuration though, not a reason for implementing your own layer.  Think about how, say MediaWiki, configure MySQL vs PostgreSQL.

> Or, they can choose to whether they want to use the registration bots library.

Hmm, I can see that working two ways:

 * If these are handled in process, the executable would pluggable (or, if we want to include them in the ref implementation, then configurable and off by default);
* But how I originally thought about these was that they'd be separate processes, and ther server would only accept registrations from the privileged places (ie. only registrations from localhost)

But ...

Please push back if the above isn't what you want.  The layering approach is also interesting and I can be convinced of it.

server/serialization.go
```diff
@@ -0,0 +1,22 @@
...
+	"encoding/base64"
+	"encoding/json"
+
+	"github.com/coniks-sys/coniks-go/merkletree"
+)
+
+var b64en = base64.StdEncoding.EncodeToString
+
+func MarshalTemporaryBinding(tb *merkletree.TemporaryBinding) ([]byte, error) {
+	return json.Marshal(&struct {
```
masomel: The TB needs to include the STR proof and the index proof if this is what is being returned to the client.

main.go
c633: > Yup. From what's here, it seems like server/message.go and server/serialization.go can move to protocol/ since I imagine they'd be used by both client/ and server/. But correct me if I'm wrong.

It makes sense. 

> But how I originally thought about these was that they'd be separate processes, and the server would only accept registrations from the privileged places (ie. only registrations from localhost)

Yeah, same thought here. 

For now, I would go with your advices. Thanks for clarifying!

server/message.go
```diff
@@ -0,0 +1,57 @@
...
+type RegistrationResponse struct {
+	TempraryBinding string `json:"tb"`
+}
+
+func NewErrorResponse(err error) *ErrorResponse {
+	return &ErrorResponse{Error: err.Error()}
+}
+
+func (reg *RegistrationRequest) serialize() []byte {
+	// serialize scheme:
```
masomel: Why is ``serialize()`` not in ``serialization.go``?

server/serialization.go
c633: Right. I will fix it after `vrf_integration` branch is merged.

server/message.go
masomel: ``serialize()``also doesn't seem to be used anywhere right now.

c633: I should rename `serialization.go` to something else. `serialization.go` includes functions that encode the messages between client & server, whereas this `serialization()` is used internally by the server (for the db storage, exactly).

c633: > serialize()also doesn't seem to be used anywhere right now.

My mistake. It should be used to serialize the user's data blob to store to db.

server/message.go
```diff
@@ -0,0 +1,57 @@
+package server
```
masomel: If we split the reusable server code and the executable implementation into separate packages, I would include this module in the "reusable" package.

masomel: Looks like this is being discussed as part of the https://github.com/coniks-sys/coniks-go/pull/29#discussion_r71884112 thread.

server/message.go
masomel: > I should rename ``serialization.go`` to something else. ``serialization.go`` includes functions that encode the messages between client & server

Yeah, I was a bit confused about the naming. Maybe something like ``messaging.go`` could work? I figured ``serialize()`` was used for something like the DB storage. We might want to move everything storage related to its own module?

server/main.example.go
c633: Should I put it into a sub-directory, namely `bin`? /cc @masomel

server/message.go
c633: > Maybe something like messaging.go could work?

But we already have `message.go`. I think we can keep that name and move `serialize()` to something like `messagekv.go`, like what we are doing in `db` branch. What do you think?

masomel: > But we already have ``message.go``. I think we can keep that name and move ``serialize()`` to something like ``messagekv.go``

I meant rename ``serialization.go`` to ``messaging.go`` since it contains functions for messaging between clients and the server. ``messagekv.go`` doesn't make so much sense to me unless the server needs an interface to store things like ``RegistrationRequest``s in the DB. I'm not sure this is necessary. Do you think the server should store all the requests it receives from the client in the DB?

> All ``***kv.go`` then go to a subpackage, e.g. ``kv`` package within ``server/kv`` directory. What do you think? 

This still sounds fine to me if the server needs additional interface(s) to the DB that aren't covered by the ``kv`` package we're going to introduce in the PR for #4. But as I mentioned above, I'm not sure it will actually need these.

main.go
masomel: > The layering approach is also interesting and I can be convinced of it.

I think there's a little more layering we can do, though I agree that any server executable we build should be considered to be a reference implementation. Other developers might want to have a different server architecture, and we should allow for that. As @c633 and I had originally thought, I think the ``protocols`` we should also include a high-level API for the CONIKS protocols and transparency operations: registration, lookups, key changes, updating the directory and generating (and publishing) new STRs, all of these with the option to store them to a DB. The handlers in the reference implementation could then simply use this API. How does this sound?

> But how I originally thought about these was that they'd be separate processes, and the server would only accept registrations from the privileged places (ie. only registrations from localhost)

Also agree on this one. This also just seems like a configuration in the server, no?

server/main.example.go
masomel: Isn't ``bin`` conventionally the place where compiled binaries are placed? Maybe naming the package something like ``keyserver`` or ``coniksServer`` would be more descriptive.

server/handler.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// decode key string
+	key, err := base64.StdEncoding.DecodeString(reg.Key)
+	if err != nil {
+		return NewErrorResponse(ErrorMalformedClientMessage),
+			ErrorMalformedClientMessage
+	}
+
+	server.state.RLock()
+	// check whether the name already exists
```
masomel: I think L21-45 could be moved to a ``Register(name, key, db)`` function in the ``protocols`` package.

server/directory.go
```diff
@@ -0,0 +1,20 @@
+package server
+
+import "github.com/coniks-sys/coniks-go/merkletree"
+
```
masomel: I might add a typedef ``type directory merkletree.PAD`` here. Then we can make ``Update()`` a function of ``directory``. The server policies and signing key could be passed as parameters?

server/main.example.go
c633: If so, I would move all the source files except main.example.go into coniksServer subpackage.

server/server.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+type tlsInfo struct {
+	TLSCertPath string `toml:"cert"`
+	TLSKeyPath  string `toml:"key"`
+}
+
+type serverState struct {
+	sync.RWMutex
+	pad *merkletree.PAD
```
masomel: If we go with the changes suggested in https://github.com/coniks-sys/coniks-go/pull/29/commits/e731ef9812bc9484dbaaef62a4d92d24de8250f4#r71962278, this could be ``directory *ConiksDirectory``

server/message.go
c633: > Actually, now that I think about it some more, does the server need an interface to store things like RegistrationRequests in the DB? I'm not sure why the server should store these, so I don't know if messagekv.go would be needed. Do you think the server should store all the requests it receives from the client in the DB?

I think we do need a DB, not for registration request, but for users' data blob including users' policies. The server needs to store users' policies and also cannot store it in the tree, right?

main.go
c633: > But how I originally thought about these was that they'd be separate processes, and the server would only accept registrations from the privileged places (ie. only registrations from localhost)

I would like to count @Liamsi in, since we all agree on this. 
> This also just seems like a configuration in the server, no?

I don't think so, since the server also has to accept lookup requests, too. I just have only 2 solutions for this:

1) We can check if the registration requests are originated from the localhost or not. (something like `conn.RemoteAddr() == conn.LocalAddr()`). But I don't know whether there is any type of attacks that can fake the remote address? 

2) Sign the request with registration bot's private key, and the key server verifies the request using the bot's public key. This way, the bot and the key server can run on 2 separated machines. 

What do you think?

Liamsi: >> But how I originally thought about these was that they'd be separate processes, and the server would only accept registrations from the privileged places (ie. only registrations from localhost)

>I would like to count @Liamsi in, since we all agree on this. 

I also thought about it in exactly this way. But I also agree on the security concerns @c633 raised below.

> We can check if the registration requests are originated from the localhost or not. (something like conn.RemoteAddr() == conn.LocalAddr()). But I don't know whether there is any type of attacks that can fake the remote address?

I fear this alone doesn't give you any reasonable security. We have to find ways to additionally secure that (firewall/network setup, TLS). But all this makes only sense if we want both processes also to be able to run on different machines (on the same machine they could use other means to communicate with each other).

c633: > I fear this alone doesn't give you any reasonable security. We have to find ways to additionally secure that (firewall/network setup, TLS). But all this makes only sense if we want both processes also to be able to run on different machines (on the same machine they could use other means to communicate with each other).

That's why I suggest the 2) option.

server/main.example.go
c633: @arlolra @masomel : What do you think about fd73f1ccca5c29fcc6a93e55cabe097dd5fc5778  and 78db200a2a10b46ece5c5a973d384e3bbe330921?

server/message.go
c633: I think it is done in 78db200a2a10b46ece5c5a973d384e3bbe330921

server/handler.go
c633: Fixed in 8be171fe5d2019717636019f32e5df888b3a0ea5.

server/server.go
c633: I changed `serverState` into `ConiksDirectory` in 8be171fe5d2019717636019f32e5df888b3a0ea5. Does it make sense?

server/directory.go
c633: I answered in https://github.com/coniks-sys/coniks-go/pull/29#discussion_r71993140

server/server.go
masomel: Yes, I think the changes make sense. 

A couple comments, though. I would merge ``protocol/ops.go`` and ``protocol/directory.go`` since the ops (at least the ones I can think of) are all related to the directory. In the same fashion, I think it makes as much sense to put ``Update()`` in ``ops.go`` as it does to keep it in ``directory.go``. 

My other high-level comment was a question about the ``tbs`` array in ``ConiksDirectory``. Do we want to force the developer to use TBs at this point? Maybe I misunderstood, but I thought we were going to use TBs as part of the reference implementation in ``keyserver``, but not build them into the main protocol implementation.

README.md
```diff
@@ -22,6 +22,7 @@ The packages in this library implement the various components of the CONIKS syst
 - ``merkletree``: Merkle prefix tree and related data structures
 - ``keyserver``: Server-side protocol implementation
```
masomel: Now that we've split up the code, I think the description here should be something like "Key server executable" or "Key server reference implementation".

README.md
```diff
@@ -22,6 +22,7 @@ The packages in this library implement the various components of the CONIKS syst
...
 - ``utils``: Utility functions 
+- ``protocol``: Reusable CONIKS server library and the server executable implementation
```
masomel: ``protocol`` also contains some client-side protocol code, no? And the server executable is in ``keyserver``? I would generalize the description here to something like "CONIKS protocols implementation/library"

keyserver/server.go
```diff
@@ -94,8 +86,7 @@ func New(conf *ServerConfig) *ConiksServer {
 	server.reloadChan = make(chan os.Signal, 1)
 	signal.Notify(server.reloadChan, syscall.SIGUSR2)
 	server.epochTimer = time.NewTimer(time.Duration(policies.EpochDeadline()) * time.Second)
-	server.state.pad = server.InitDirectory(conf.LoadedHistoryLength)
-	server.state.tbs = make(map[string]*merkletree.TemporaryBinding, Capacity)
+	server.directory = protocol.Init(policies, sk, conf.LoadedHistoryLength, conf.RegistrationCapacity)
```
masomel: Calling ``protocol.Init()`` seems a bit awkward to me when what you're doing is initializing the server's directory. Maybe ``InitDirectory()`` would be more clear?

protocol/ops.go
```diff
@@ -0,0 +1,37 @@
...
+	"log"
+
+	"github.com/coniks-sys/coniks-go/merkletree"
+)
+
+func (d *ConiksDirectory) Register(uname string, key []byte) (*merkletree.TemporaryBinding, int) {
+	d.RLock()
+	// check whether the name already exists
+	// in the directory before we register
+	ap := d.pad.Lookup(uname)
```
masomel: I think we need to be return ``ap`` as the proof of absence along with the TB and the error.

c633: True.

protocol/directory.go
```diff
@@ -2,67 +2,43 @@ package protocol
...
 
 type ConiksDirectory struct {
-	sync.RWMutex
 	pad *merkletree.PAD
-	tbs map[string]*merkletree.TemporaryBinding
 }
 
 func InitDirectory(policies merkletree.Policies, secretKey crypto.SigningKey,
-	directorSize, capacity uint64) *ConiksDirectory {
+	directorSize uint64) *ConiksDirectory {
```
masomel: Seems like you're cutting off just one letter from ``directorSize``. If you want to abbreviate, maybe just call it something like ``dirSize``?

keyserver/server.go
```diff
@@ -31,7 +31,9 @@ type tlsInfo struct {
 }
 
 type ConiksServer struct {
+	sync.Mutex
 	directory *protocol.ConiksDirectory
+	tbs       map[string]*merkletree.TemporaryBinding
```
masomel: Moving the mutex to the server as well makes a lot of sense!

keyserver/handler.go
```diff
@@ -19,10 +19,22 @@ func (server *ConiksServer) handleRegistrationMessage(reg *p.RegistrationRequest
 			p.Error(p.ErrorMalformedClientMessage)
 	}
 
+	server.Lock()
+	// check the temporary bindings array first
+	// currently the server allows only one registration/key change per epoch
+	if server.tbs[reg.Username] != nil {
```
masomel: If we already have a TB for the name, do we want to return some kind of proof to the client along with the error? cc @arlolra @Liamsi

keyserver/handler.go
```diff
@@ -28,18 +28,28 @@ func (server *ConiksServer) handleRegistrationMessage(reg *p.RegistrationRequest
...
+	ap, tb, errCode := server.directory.Register(reg.Username, key)
 	if errCode != p.Success {
 		server.Unlock()
 		return p.NewErrorResponse(errCode), p.Error(errCode)
 	}
 	server.tbs[reg.Username] = tb
 	server.Unlock()
 
+	// TODO: do we want to store the user policies into DB?
+
```
masomel: By user policies you mean flags such as ``allowUnsignedKeyChange``? I think they should be stored in the DB as well since other clients need to use them.

keyserver/handler.go
```diff
@@ -28,18 +28,28 @@ func (server *ConiksServer) handleRegistrationMessage(reg *p.RegistrationRequest
...
 		return p.NewErrorResponse(p.ErrorInternalServer), err
 	}
+	apEncoded, err := p.MarshalAuthenticationPath(ap)
+	if err != nil {
+		return p.NewErrorResponse(p.ErrorInternalServer), err
+	}
 
-	return &p.RegistrationResponse{TempraryBinding: string(tbEncoded)}, nil
+	return &p.RegistrationResponse{
+		RegEpoch: server.directory.LatestEpoch(),
```
masomel: Shouldn't this be the next epoch since the registration won't appear in the tree until then?

protocol/message.go
```diff
@@ -25,7 +25,9 @@ type ErrorResponse struct {
 }
 
 type RegistrationResponse struct {
-	TempraryBinding string `json:"tb"`
+	RegEpoch uint64 `json:"reg_epoch"`
+	AP       string `json:"ap"`
+	TB       string `json:"tb"`
```
masomel: I think the response should also contain the STR for ``RegEpoch`` since the client needs it to verify the TB signature as well as to validate the AP's root hash. If the STR is included do we still need ``RegEpoch``, too?

protocol/directory.go
```diff
@@ -26,19 +26,27 @@ func (d *ConiksDirectory) Update(policies merkletree.Policies) {
...
 	// in the directory before we register
-	ap := d.pad.Lookup(uname)
-	if ap.Leaf().Value() != nil {
-		return nil, ErrorNameExisted
+	ap, err := d.pad.Lookup(uname)
+	if err != nil {
+		return nil, nil, ErrorInternalServer
+	}
+	if !ap.Leaf().IsEmpty() {
+		return ap, nil, ErrorNameExisted
```
masomel: Couldn't a non-empty AP leaf also mean there's a different index that shares the prefix with ``uname``? This would be a different case than ``ErrorNameExisted``, no?

keyserver/policy.go
```diff
@@ -1,14 +1,18 @@
 package keyserver
 
 import (
+	"io/ioutil"
 	"log"
 
 	"github.com/BurntSushi/toml"
 	"github.com/coniks-sys/coniks-go/merkletree"
+	"github.com/yahoo/coname/vrf"
```
masomel: Why aren't we using our version of ``vrf`` here?

keyserver/policy.go
```diff
@@ -1,14 +1,18 @@
...
 )
 
 type ServerPolicies struct {
 	EpochDeadline merkletree.TimeStamp `toml:"epoch_deadline"`
+	VRFKeyPath    string               `toml:"vrf_key_path"`
+	VRFKey        [vrf.SecretKeySize]byte
```
masomel: I think the public key of the VRF key pair should be part of the ``Policies`` interface in ``merkletree/policy.go``

masomel: Nevermind, I didn't see it already is. Sorry for the noise.

protocol/encoding.go
```diff
@@ -15,12 +15,18 @@ var b64en = base64.StdEncoding.EncodeToString
 
 func MarshalTemporaryBinding(tb *merkletree.TemporaryBinding) ([]byte, error) {
 	return json.Marshal(&struct {
-		Index     string `json:"index"`
-		Value     string `json:"value"`
-		Signature string `json:"signature"`
+		Index      string `json:"index"`
+		IndexProof string `json:"index_proof"`
```
masomel: I don't think we need an ``IndexProof`` field in the TB if it's already contained in the auth path.

keyserver/handler.go
Liamsi: Not sure, if this is a good idea: But wouldn't it make more sense to return the already existing TB instead of an error here?

c633: If so, I would return the proof of inclusion along with the error if the registering name also already exists in the tree.

protocol/directory.go
c633: No, my bad. The check must be `ap.Leaf().Value() == nil`

keyserver/policy.go
c633: Yup, sorry for the noise.

protocol/encoding.go
c633: Yup, I agree. Thanks!

protocol/message.go
c633: Yes, makes sense.
> If the STR is included do we still need RegEpoch, too?

I think no.

keyserver/handler.go
c633: Yes. Thanks!

keyserver/handler.go
c633: Yes. It's what I asked you in https://github.com/coniks-sys/coniks-go/pull/29#discussion_r71963914

protocol/directory.go
c633: Oh sorry. :(

Btw, `dirSize` sounds good!

server/message.go
masomel: Right, though the policies should be included in the tree. I was thinking the client could use the first few bytes of the user data to encode these flags? Or use the first byte as a bitmap? This way, they're stored in the tree, and the server still doesn't know exactly what's in the user's data blob.

keyserver/handler.go
masomel: Oh, sorry I missed that! I just responded.

keyserver/handler.go
masomel: I agree that the error should be returned to the client to distinguish from the success case. We could return the existing TB along with the auth path, as @c633 suggested.

protocol/directory.go
masomel: No worries at all!

c633:
Another question need to be addressed: https://github.com/coniks-sys/coniks-go/pull/29#discussion_r71985491

main.go
arlolra: I had assumed we'd accept registrations on a separate port, and have the firewall restrict access to it.  But 2) is also a viable option.  Does anyone have a strong preference?

masomel:
> Another question need to be addressed: #29 (comment)

What I meant by "This also just seems like a configuration in the server" is that the port and/or IP address that the server accepts registration requests on would be part of the server configuration so that the server could then make the check you suggest in 1). So I agree that we should check the bot's IP address/port, but I also think that the bot should sign each registration request. 

But I think these checks don't need to happen in the server. As @Liamsi and @arlolra mentioned, checking the IP address and port could be done by a firewall, and having the registration bot sign each request will be a lot like using TLS between the server and the bot. I'm fine with either approach to 1). I think using TLS is good approach to 2).

keyserver/handler.go
```diff
@@ -56,3 +65,12 @@ func (server *ConiksServer) handleRegistrationMessage(reg *p.RegistrationRequest
 		TB:   tb,
 	}, nil
 }
+
+func (server *ConiksServer) StoreUserPoliciesToKV(up *p.ConiksUserPolicies) error {
```
masomel: As I suggested in https://github.com/coniks-sys/coniks-go/pull/29#discussion_r72646673, if we store the user policies as part of the user leaf node's ``value()``, I don't think we need a separate ``StoreUserPoliciesToKV()``.

protocol/directory.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/coniks-sys/coniks-go/merkletree"
 )
 
+type ConiksUserPolicies struct {
```
masomel: Same as https://github.com/coniks-sys/coniks-go/pull/29/commits/5584edca68630937362c755ce085b2a0d156fd81#r72910690 goes for this.

c633:
> I think using TLS is good approach to 2).

Do you mean mutual TLS? I think we could use a separate mutual TLS connection between the server and the bot for the registration.

keyserver/handler.go
c633: I forgot to reply your comment :(

Do you think that the server still needs to know the user's policies in some cases, e.g., _signed/unsigned key change_ policy? If so, we have to store the user policies into db, right?

masomel: No problem, I've definitely forgotten to reply to comments too!

> Do you think that the server still needs to know the user's policies in some cases, e.g., *signed/unsigned key change* policy? If so, we have to store the user policies into db, right?

I think in the case of the lookup policy, I don't think the server needs to know. Depending on whether the server will verify key changes, then you're right, the server would still need to know the user's policy. In my suggestion, it's not that the user policies aren't stored in the DB, they're just made part of the blob of data mapped to the user's index and not stored separately. Do you think there would be too much overhead in retrieving the entire user leaf node from the DB to get something like a user policy bitmap or flag every time the server wants to check the key change policy?

masomel:
> Do you mean mutual TLS? I think we could use a separate mutual TLS connection between the server and the bot for the registration.

I'm guessing you mean mutual authentication in TLS? I'm not sure what you mean by "separate TLS connections between the server and the bot", though. But yes, the server and bot would definitely verify each other's certs.

c633:
> I'm guessing you mean mutual authentication in TLS? ... But yes, the server and bot would definitely verify each other's certs.

Yes, it is.

keyserver/handler.go
c633: > In my suggestion, it's not that the user policies aren't stored in the DB, they're just made part of the blob of data mapped to the user's index and not stored separately.

I'm kind of confused here. Why do we need to store the user policies to both db and the data blob mapped to the user's index? 
Another question, do we want to preserve the user's privacy (in this case, it is the user policies) from the lookup operation? Could we just return all the data blob to the user and the user policies would be exposed; or do we have to do some kind of filtering/trimming the data blob before returning to the client? If we follow the latter way, then it also cause some overhead, much like querying the db, right?

masomel: > I'm kind of confused here. Why do we need to store the user policies to both db and the data blob mapped to the user's index? 

Sorry. I probably didn't explain clearly what I meant. What I'm trying to say is that if the data blob is already in the DB, and we store the user's policies as part of the blob, we don't need to also store them in the DB again. The current implementation in ``handler.go`` seems to be storing the user policies in a separate entry in the DB, and the main goal of my suggestion was to save space in the DB and store all user-related data in the same place, i.e. the blob. I hope this makes more sense.

> Another question, do we want to preserve the user's privacy (in this case, it is the user policies) from the lookup operation? Could we just return all the data blob to the user and the user policies would be exposed; or do we have to do some kind of filtering/trimming the data blob before returning to the client?

Good question. I think the policies must be revealed to the client, otherwise the client won't be able to parse the rest of the data blob. In a proof of absence, the policies shouldn't be included anyway since the entire blob is dropped from the proof. Bu if the server removes the user policies before responding to a proof of inclusion, how will the client know if the data it's receiving is the user's key or an encrypted blob? And how will the client know if it should look for a signature on a key change message? 

In the case of the lookup policy, revealing the policy to the client shouldn't violate the user's privacy if the client isn't authorized to decrypt the encrypted blob. In the case of the key change policy, I do think there might be a risk to exposing the policy to a malicious client (which might attempt to change a user's key if it detects that the user allows unsigned key changes). But I think the solution to this issue might be some kind of account authentication again.

c633: It makes sense!

>  What I'm trying to say is that if the data blob is already in the DB, and we store the user's policies as part of the blob, we don't need to also store them in the DB again.

Yup, I got it. However, in my latest implementation, I decided to not store all the data blob in the DB anymore since I see there is no reasons to do it. What do you think?

> I think the policies must be revealed to the client, otherwise the client won't be able to parse the rest of the data blob. In a proof of absence, the policies shouldn't be included anyway since the entire blob is dropped from the proof. But if the server removes the user policies before responding to a proof of inclusion, how will the client know if the data it's receiving is the user's key or an encrypted blob? And how will the client know if it should look for a signature on a key change message?

In the case of the lookup policy, revealing the policy to the client shouldn't violate the user's privacy if the client isn't authorized to decrypt the encrypted blob. In the case of the key change policy, I do think there might be a risk to exposing the policy to a malicious client (which might attempt to change a user's key if it detects that the user allows unsigned key changes). But I think the solution to this issue might be some kind of account authentication again.

Okay, we should store the user policies as part of the blob mapped to the user's index as well. Makes sense! 
I would like to open separate issue for the key change function, though.
Just a side-note, we should de-serialize the data blob to user's key, user policies before responding (it maybe looks like `{"key": "...", "allow_public_lookup": true, "allow_unsigned_key_change": true}` instead of `{"blob": "allow_public_lookup||allow_unsigned_key_change||key"}`

masomel: > However, in my latest implementation, I decided to not store all the data blob in the DB anymore since I see there is no reasons to do it. What do you think?

Oh, I missed that. Sorry! Let me go look.

> I would like to open separate issue for the key change function, though.

Absolutely, makes sense!

> Just a side-note, we should de-serialize the data blob to user's key, user policies before responding

If we do this, though, don't we lose a bit of privacy from the server? I.e. the server now has more information about what is stored in the blob. I'd like to loop @arlolra @Liamsi into this discussion, too.

c633: > If we do this, though, don't we lose a bit of privacy from the server? 

Oh I see. Also if the data blob is encrypted, there is no way the server can de-serialize the data blob. Sorry for the noise!

masomel: >  if the data blob is encrypted, there is no way the server can de-serialize the data blob.

The user policies wouldn't be included in the encrypted portion, only the public key, so I think the server could still theoretically de-serialize the blob. But it would still be a privacy concern.

c633: Oh, thanks for clarifying!

c633:
I decided to make the server to listen to 2 separate ports, one for local connection with registration bots and the other for public connection. Since this way is so common and familiar with all of you here, I think it would be safe to go this way. 

However, both listeners use the same handlers, so the registration port can also accept the lookup requests as well. I don't know if there is any problem with this? 

I think this pull is already for another review. 
/cc @arlolra @masomel @Liamsi

Liamsi:
> However, both listeners use the same handlers, so the registration port can also accept the lookup requests as well. I don't know if there is any problem with this?

That sounds concerning. But I'll have a look into it now. Maybe, it's not a problem.

keyserver/server.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	config := &tls.Config{Certificates: []tls.Certificate{cer}}
+	// server public port
+	ln, err := tls.Listen("tcp", conf.Address, config)
+	if err != nil {
+		panic(err)
+	}
+
+	// server regitration port
```
Liamsi: Typo: s/regitration/registration

keyserver/server.go
```diff
@@ -0,0 +1,183 @@
...
+	regLn, err := tls.Listen("tcp", conf.RegistrationAddress, config)
+	if err != nil {
+		panic(err)
+	}
+
+	server.waitStop.Add(1)
+	go server.listenForRequests(ln)
+
+	server.waitStop.Add(1)
+	go server.listenForRequests(regLn)
```
Liamsi: Can you help me here a little: If I understand this correctly, it means that we start the same server on two different ports. Currently, the server handles all message with `ConiksServer.handleClientMessage`. On this branch it this can only handle messages of type `RegistrationRequest`. So basically a client can register connecting to one port or the other. Is that correct?

c633: Oh, you're right. I have to have separate listener/handlers for each port.

c633:
> ~~However, both listeners use the same handlers, so the registration port can also accept the lookup requests as well. I don't know if there is any problem with this?~~

Thanks @Liamsi  for pointing out the problem. For now, each listener takes a function pointer to the handler corresponding with the port.

keyserver/server.go
c633: Thanks @Liamsi . I fixed it in 7f46856f48111af595a8aa42e26da02363939803. Each listener now takes a function pointer to the handler corresponding with the port.

utils/configen/mkconfig.go
```diff
@@ -0,0 +1,75 @@
...
+const ConfigFile = "config.toml"
+const PoliciesFile = "policies.toml"
+
+func main() {
+
+	var conf = keyserver.ServerConfig{
+		SigningKeyPath:       "ed25519.secret",
+		DatabasePath:         "coniks.db",
+		RegistrationAddress:  "0.0.0.0:3001",
+		Address:              "0.0.0.0:3000",
```
Liamsi: Another question: Are both of these ports meant to be reachable from outside?

Liamsi:
That was fast :-) I'll have another more detailed look into this later. 

One (maybe stupid) question regarding the (coniks-)server from/to registration-bot communication: did we ever consider using channels instead of letting them communicate using a local network connection?

utils/configen/mkconfig.go
c633: Thanks! `RegistrationAddress` should be `127.0.0.1`, right?

c633:
> One (maybe stupid) question regarding the (coniks-)server from/to registration-bot communication: did we ever consider using channels instead of letting them communicate using a local network connection?


I did. But I think we maybe run server/bot on difference machines in the future ( @arlolra  said that they should be run in the same machine but ..., so I would like to use the local network connection instead.

utils/configen/mkconfig.go
Liamsi: If you only want to listen on the loopback interface (localhost), yes.

c633: Thanks!

keyserver/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+Generation of self-signed(x509) public key (PEM-encodings `.pem`) based on the private (`.key`)
+```
+openssl req -new -x509 -sha256 -key server.key -out server.pem -days 3650
+```
+
+Run the server
+```
+go run keyserver/coniksserver/main.go
+```
```
Liamsi: Is this supposed to be a complete set of instructions? It seems like the signing key needs to be generated for this to work and the above command `go run utils/keygen/mkkey.go -vrf -signing` only uses the first flag. Hence, trying to run the server will fail.

utils/keygen/mkkey.go
```diff
@@ -0,0 +1,99 @@
...
+	var vrf = flag.Bool("vrf", false, "Generate VRF key pair")
+	var sign = flag.Bool("signing", false, "Generate Signing key pair")
+	flag.Parse()
+
+	switch true {
+	case *vrf:
+		mkVrfKey()
+	case *sign:
+		mkSigningKey()
+	}
```
Liamsi: if ran with both flags `-vrf -signing` this will only generate the vrf key

keyserver/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Usage
+Generate new config file with proper format
+```
+go run utils/configen/mkconfig.go
+```
+
+Generate new key pair for signing
+``` 
+go run utils/keygen/mkkey.go -vrf -signing
```
Liamsi: Make this two commands (one with signing and one with vrf flag) or change mkkey.go accordingly

utils/keygen/mkkey.go
c633: Oh I will fix it.

keyserver/README.md
c633: Thanks! I fixed it in f82d43c83ac69eabb5024122556c779addca2933. It should generate both keys now.

keyserver/README.md
c633: > Is this supposed to be a complete set of instructions? 

Yes, it is. 

>  It seems like the signing key needs to be generated for this to work and the above command go run utils/keygen/mkkey.go -vrf -signing only uses the first flag.

I fixed it (see the answer below). Thanks!

keyserver/README.md
Liamsi: Cool! Thanks. Haven't had time for a complete review but is it possible to already play around with the client to let it interact with the server?

keyserver/listener.go
```diff
@@ -100,6 +100,37 @@ func (server *ConiksServer) handleClientMessage(msg []byte) ([]byte, error) {
...
+
+marshalling:
+	res, e := json.Marshal(response)
+	if e != nil {
+		panic(p.Error(p.ErrorInternalServer))
+	}
+	return res, err
+}
+
+func (server *ConiksServer) handleBotMessage(msg []byte) ([]byte, error) {
```
masomel: It seems like ``handleClientMessage()`` shares a fair amount of code with ``handleBotMessage()``. Maybe it's possible to refactor some of this repeated code into small functions? For example, L121-L130 is reused, and the default case of the switch is identical in both functions.

keyserver/README.md
c633: Yes. This is a quick & dirty go client I'm using for testing: https://play.golang.org/p/aQUaz2MstE

keyserver/listener.go
c633: Exactly, I was aware this issue, too. But I didn't find out any effective way to refactor this scope yet. I'm thinking about it.

masomel: No worries if you don't have a solution for this yet. This isn't super urgent, but it would be nice to refactor this as this will be useful if/when we need to write additional handlers.

Liamsi:
I think this PR could have a few more (unit) tests and needs some manual testing (using https://github.com/coniks-sys/coniks-go/pull/29#discussion_r73724926 for instance). Maybe, the first could be overkill (because of the networking connections) and the 2nd could be done later.
But the code looks good to me now.

c633:
> I think this PR could have a few more (unit) tests and needs some manual testing (using #29 (comment) for instance). Maybe, the first could be overkill (because of the networking connections) and the 2nd could be done later.

Yup, I agree that we need some tests. I'm working on this.

c633:
I added some tests for server in d78018287dc4f437fcba768fab3278f3f940e298.
Do you have any suggestion on https://github.com/coniks-sys/coniks-go/pull/29#discussion_r73723745?

masomel:
> Another question, my `directory.Register` involves `TemporaryBinding` (https://github.com/coniks-sys/coniks-go/blob/eef5d811d4b3c087753d5da9f42d118957337e35/protocol/directory.go#L40). But we have discussed that `TemporaryBinding` should be an option, and it should not be included in the `protocol`. If I replace this `pad.TB` with `pad.Set` to avoid using the `TB`, then I have to export another method for the developer to retrieve the `TB`. Lookup the current `TB` implementation (https://github.com/coniks-sys/coniks-go/blob/master/merkletree/pad.go#L127-L132), we probably have to call `pad.Set` twice.
On the other hand, should we exclude the TB from protocol, since it is the only way to allow immediate use of the key currently?

This is a good question. I think we should not include `TB` directly in `directory.Register()`, but like you noticed, we shouldn't be calling `pad.Set` twice. My suggestion would be to remove L130 from `pad.go` to decouple `pad.TB` and `pad.Set` and so that `pad.Set` isn't called twice. `directory.go` should possibly export a wrapper around `pad.TB`. What do you think?

keyserver/handler.go
```diff
@@ -47,7 +48,15 @@ func (server *ConiksServer) handleRegistrationMessage(reg *p.RegistrationRequest
 	server.tbs[reg.Username] = tb
 	server.Unlock()
 
-	// TODO: store the user policies into DB after kv branch is merged
+	// store the user policies into DB
```
masomel: If we're making the user policies part of the data blob stored into the DB, as we've been discussing in #46, I don't think we should be storing the user policies into the DB separately here, too.

keyserver/server.go
```diff
@@ -38,7 +42,8 @@ type ConiksServer struct {
 	stop     chan struct{}
 	waitStop sync.WaitGroup
 
-	secretKey crypto.SigningKey
+	db        kv.DB
+	secretKey sign.PrivateKey
```
masomel: Are we going to keep the server's secret key in here for now, and move it to the STR policies as we were discussing in https://github.com/coniks-sys/coniks-go/pull/39#discussion_r73080612, once this pull is merged/when we finalize #47?

Liamsi:
> I added some tests for server in eef5d81

Looks good!

keyserver/testutil/testutil.go
```diff
@@ -0,0 +1,113 @@
...
+	"time"
+)
+
+const (
+	TestDir          = "coniksServerTest"
+	PublicConnection = "127.0.0.1:3000"
+	LocalConnection  = "127.0.0.1:3001"
+)
+
+func CreateTLSCert(t *testing.T) (dir string, teardown func()) {
```
Liamsi: Nitpicking: If you give the return value a name (like here), I think it would be good to actually use it below (like you did with `dir`)

Liamsi:
> directory.go should possibly export a wrapper around pad.TB. What do you think?

@masomel Wouldn't returning a wrapper around TB still make the temporary binding non-optional? (Maybe I didn't get the full idea here...)

keyserver/README.md
```diff
@@ -0,0 +1,30 @@
+# CONIKS Server implementation in Golang
+
+## Usage
+Generate new config file with proper format
+```
+go run utils/configen/mkconfig.go
+```
+
+Generate new key pair for signing
```
arlolra: doc is out-of-date.  signing and vrf now

protocol/directory.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		panic(err)
+	}
+	d := new(ConiksDirectory)
+	d.pad = pad
+	return d
+}
+
+func (d *ConiksDirectory) Update(policies merkletree.Policies) {
+	d.pad.Update(policies)
```
arlolra: Probably want to use an anonymous filed for `pad`.

```
diff --git a/protocol/directory.go b/protocol/directory.go
index 9869168..f2f4ede 100644
--- a/protocol/directory.go
+++ b/protocol/directory.go
@@ -14,7 +14,7 @@ type ConiksUserPolicies struct {
 }
 
 type ConiksDirectory struct {
-	pad *merkletree.PAD
+	*merkletree.PAD
 }
 
 func InitDirectory(policies merkletree.Policies, secretKey sign.PrivateKey,
@@ -23,24 +23,14 @@ func InitDirectory(policies merkletree.Policies, secretKey sign.PrivateKey,
 	if err != nil {
 		panic(err)
 	}
-	d := new(ConiksDirectory)
-	d.pad = pad
-	return d
-}
-
-func (d *ConiksDirectory) Update(policies merkletree.Policies) {
-	d.pad.Update(policies)
-}
-
-func (d *ConiksDirectory) LatestSTR() *merkletree.SignedTreeRoot {
-	return d.pad.LatestSTR()
+	return &ConiksDirectory{pad}
 }
 
 func (d *ConiksDirectory) Register(uname string, key []byte) (
 	*merkletree.AuthenticationPath, *merkletree.TemporaryBinding, int) {
 	// check whether the name already exists
 	// in the directory before we register
-	ap, err := d.pad.Lookup(uname)
+	ap, err := d.Lookup(uname)
 	if err != nil {
 		return nil, nil, ErrorInternalServer
 	}
@@ -49,7 +39,7 @@ func (d *ConiksDirectory) Register(uname string, key []byte) (
 	}
 
 	// insert new data to the directory on-the-fly
-	tb, err := d.pad.TB(uname, key)
+	tb, err := d.TB(uname, key)
 	if err != nil {
 		log.Printf(err.Error())
 		return nil, nil, ErrorInternalServer
```

utils/keygen/mkkey.go
```diff
@@ -0,0 +1,99 @@
+// Copyright 2014 The Dename Authors.
```
arlolra: Means you need to update the LICENSE file.

utils/configen/mkconfig.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
+
+	"github.com/BurntSushi/toml"
+	"github.com/coniks-sys/coniks-go/keyserver"
```
arlolra: This util belongs in `keyserver/utils/`.  Generally, utils probably doesn't want to depend on other subpackages.

arlolra:
> But we have discussed that TemporaryBinding should be an option, and it should not be included in the protocol.

Sorry to be joining the party late, but why shouldn't it be part of the protocol? Can you give me a pointer to that discussion.

It seems like a protocol extension and maybe I'd model it like,

```
type ConiksDirectory struct {
  *merkletree.PAD
  useTBs bool
}
```

> My suggestion would be to remove L130 from pad.go to decouple pad.TB and pad.Set and so that pad.Set isn't called twice.

I'm wary of decoupling the two.

keyserver/server.go
c633: I think it could be removed for now. I just re-checking the source code, and `server.secretKey` is being used nowhere.

keyserver/testutil/testutil.go
c633: Thanks!

c633:
> Sorry to be joining the party late, but why shouldn't it be part of the protocol? Can you give me a pointer to that discussion.

It's here: https://github.com/coniks-sys/coniks-go/pull/21#issuecomment-234103819

> My suggestion would be to remove L130 from pad.go to decouple pad.TB and pad.Set and so that pad.Set isn't called twice.

This way, we need to compute the private index twice, or `pad.TB` would take `index` as its argument.

protocol/directory.go
c633: Cool! Thanks!

arlolra:
> It's here: #21 (comment)

I read that only as saying that it should be optional, and not mandating that it shouldn't be part of the `ConiksDirectory`.

c633:
Here is another comment of @masomel : https://github.com/coniks-sys/coniks-go/pull/29#discussion_r72115576

utils/configen/mkconfig.go
c633: I'm not sure if this util should belong in `keyserver/util` since, I also use it to generate bot configurations. https://github.com/coniks-sys/coniks-go/blob/aed779393105af0d4658c389f097574e40373d6e/utils/configen/mkconfig.go

Should I make it 2 separate utils?

utils/keygen/mkkey.go
c633: Thanks! I updated it in bab52f1f013d590fba34478545d95d7e9daa97ca

keyserver/README.md
c633: Thanks! I updated it in bab52f1f013d590fba34478545d95d7e9daa97ca

keyserver/handler.go
c633: > If we're making the user policies part of the data blob stored into the DB

Do you mean after #37  is merged? 

> I don't think we should be storing the user policies into the DB separately here, too.

Agree!

arlolra:
> Here is another comment of @masomel : #29 (comment)

Yeah, so the problem you're trying to solve is,

> then I have to export another method for the developer to retrieve the TB

because `TB`s are inherently part of the `PAD`, which you've abstracted away with `ConiksDirectory`.  There's no getting around that.

That's why I'm saying they're a protocol extension, that we're better off exposing as an option at construction time, namely `useTBs`.

arlolra:
The corollary to this is that,

```
tbs       map[string]*merkletree.TemporaryBinding
```

and subsequent logic, would move to the `ConiksDirectory`.

Going one step further, maybe the `keyserver` should only be importing the `protocol` and not `merkletree`.  I don't know, just thinking out loud.

arlolra:
We should probably wait on @masomel though.

utils/configen/mkconfig.go
arlolra: > Should I make it 2 separate utils?

I think you should move this stuff to,

```
coniksbot/main.go
coniksserver/main.go
```

which could each take a flag, `-genconfig` or something.

c633: Yeah, it makes sense.

protocol/directory.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (d *ConiksDirectory) Register(uname string, key []byte) (
+	*merkletree.AuthenticationPath, *merkletree.TemporaryBinding, ErrorCode) {
+	// check whether the name already exists
+	// in the directory before we register
+	ap, err := d.Lookup(uname)
+	if err != nil {
+		return nil, nil, ErrorInternalServer
+	}
+	if ap.Leaf.Value() != nil {
```
arlolra: This seems wrong.  `Lookup` returns an `AuthenticationPath`, the leaf of which might not be empty.  You probably want to reuse some logic from proof of absence here.

c633: I'm pretty sure about it. We're doing it in our tests, e.g., https://github.com/coniks-sys/coniks-go/blob/master/merkletree/pad_test.go#L88-L92

arlolra: Please think about it some more.

c633:
I created new branch https://github.com/coniks-sys/coniks-go/tree/server-reg-tb with some changes related to `TB` we discussed above.

protocol/directory.go
c633: What do you think about eb7245df72cebc2de19f43e042d4d68314262b96? Agrr, I probably missed something here.

arlolra: Probably unnecessary.   I don't think you need to worry about the `Value()` here.  The check you need to do is,

```
if bytes.Equal(pad.computePrivateIndex(uname, pad.policies.vrfPrivate()), ap.Leaf.Index()) {
  return nil, nil, ErrorInternalServer
}
```

Does that help?

c633: Thanks! It looks much better now.

arlolra: Ok, but, just to be sure, do you understand why the index is the important attribute to check here?

keyserver/coniksserver/main.go
```diff
@@ -1,17 +1,28 @@
...
+const ConfigFile = "config.toml"
+const PoliciesFile = "policies.toml"
+
 func main() {
+	genConfigPtr := flag.Bool("genconfig", false, "Generate server config")
 	configPathPtr := flag.String("config", "config.toml", "path to config file")
 	flag.Parse()
 
+	if *genConfigPtr {
+		mkConfig()
```
arlolra: return

protocol/directory.go
c633: I'm not sure but I think because index is unique?

protocol/error.go
```diff
@@ -0,0 +1,30 @@
+// Defines constants representing the types
+// of errors that the server may return to a client.
+
+package protocol
+
+import "errors"
+
+type ErrorCode int
+
+const (
```
arlolra:
```
const (
	Success                     ErrorCode = iota + 10
	ErrorInternalServer         ErrorCode
	ErrorNameExisted            ErrorCode
	ErrorMalformedClientMessage ErrorCode
)
```

protocol/directory.go
c633: And protocol functions are based on the index.

arlolra: Unique **to the `uname`**, yes.  Your comment says,

> check whether the name already exists

It's not enough to check that the `Value()` is `nil`, because the `ap.Leaf` can be one of three things,

 - empty
 - another leaf with a shared prefix, but `i != j`
 - a leaf with `i == j`  // the error case

This is the same as the proof of absence / inclusion, but we don't need to verify the auth path, because we're the directory here, not a client, so we trust ourself.

masomel:
> That's why I'm saying they're a protocol extension, that we're better off exposing as an option at construction time, namely `useTBs`.

I agree with this, I like the idea of exposing a construction-time option to use TBs.

> I read that only as saying that it should be optional, and not mandating that it shouldn't be part of the `ConiksDirectory`.

In the current implementation of `ConiksDirectory` the TB is part of the core protocol, as there is only one `Register` function that uses TBs. This is why I said that TBs shouldn't be part of `ConiksDirectory`. I suppose my original thought to completely decouple TBs entirely from PAD and the protocol wasn't well thought out, though. Maybe the solution then is to expose a `Register` function that uses TBs in addition to one that doesn't. At the `merkletree` level, we already expose a `pad.Set` function that uses TBs (i.e. `pad.TB`).

> maybe the `keyserver` should only be importing the `protocol` and not `merkletree`.

This is an interesting thought, and my idea for having a `ConiksDirectory` was to abstract away the underlying Merkle tree data structure, so I think we should definitely consider doing this. I'm sorry, @c633, for going back and forth on this. I know I originally (like a month ago) said that the `tbs` array should be part of the server, not the directory. But if we want to be consistent, we should probably propagate the fact that there's a TB option in the PAD up to the `ConiksDirectory` so that the `keyserver` doesn't need to know anything about the `merkletree` package, and all interactions with it happen through the `ConiksDirectory`.

arlolra:
> Maybe the solution then is to expose a Register function that uses TBs in addition to one that doesn't.

It might be fine to just ignore (`_`) the returned TB in the `!useTBs` case.  They'll be `nil` anyways.  Instead of having multiple versions of `Register()`.

masomel:
> It might be fine to just ignore (_) the returned TB in the !useTBs case. They'll be nil anyways. Instead of having multiple versions of Register().

Oh right, because Go let's you do that ^.^ Sorry, been writing/reading a lot of Java and C lately...

masomel:
> @masomel Wouldn't returning a wrapper around `TB` still make the temporary binding non-optional? (Maybe I didn't get the full idea here...)

@Liamsi What I meant was that ``pad.TB`` shouldn't be called in `directory.Register` but in a separate `ConiksDirectory` function instead so that the key server would call `Register` and `pad.TB-wrapper` separately. But based on the https://github.com/coniks-sys/coniks-go/pull/29#issuecomment-238711278 thread and https://github.com/coniks-sys/coniks-go/pull/29#issuecomment-238755293, we won't be following the approach I suggested originally.

protocol/error.go
c633: Thanks!

protocol/directory.go
c633: I totally agree that we have to use index to check for proof of inclusion. However, I just want to clarify a little bit here: the only case `leaf.Value() != nil` is where the auth path returns a leaf with `i ==j`.
And yeah, thank you for this catch!

c633:
> In the current implementation of ConiksDirectory the TB is part of the core protocol, as there is only one Register function that uses TBs.

I think it would also be used in `KeyLookup` as well.

protocol/directory.go
arlolra: > the only case leaf.Value() != nil is where the auth path returns a leaf with i == j.

No, in this case too, "another leaf with a shared prefix, but i != j"

If the leaf is not empty (`!leaf.Empty()`), then `Value()` is the `node.value`

See https://github.com/coniks-sys/coniks-go/blob/master/merkletree/proof.go#L123

That's precisely what was problematic here.

c633: > No, in this case too, "another leaf with a shared prefix, but i != j"

I remember that to preserve the privacy, the value of this leaf should not be returned. It's what we're actually doing: https://github.com/coniks-sys/coniks-go/blob/master/merkletree/merkletree.go#L77-L84

arlolra: Ok, so all this time you've been right and I'm trying to explain something to you that *I* don't understand.  That's kind of sad.

However, judging by how long it took me to catch on, I would say two things,

 * Checking the index is always the correct thing to do here, whereas value == nil feels like an implementation detail, in that someone could choose to suppress the value at another location, like serialization time. So, I think we should leave the LookupIndex == Index() comparison you implemented.

* Now that I see what's going, I absolutely agree that we should merge eb7245df72cebc2de19f43e042d4d68314262b96.

Finally, I can't believe I said,

> Please think about it some more.

I'm going to go hide in shame.

masomel:
> I think it would also be used in `KeyLookup` as well. 

Yes, and also key changes.

> The only thing belonging to `merkletree` is `Policies` now. I'm not sure if we can abstract it away since I'm using it for updating the policies. 

Ah, that's right. Could we add some kind of typedef in `ConiksDirectories`? (I apologize if there's something obviously wrong about this question because I don't fully understand how Go works yet.) I also noticed that there's a `policy.go` in `keyserver` containing some information that's also contained in `merkletree.policy`. I think we need to find a way to dedup those.

protocol/directory.go
c633: > Checking the index is always the correct thing to do here

Absolutely agree! Before this discussion, I wasn't aware that we could use `lookupindex == index` to check for proof of inclusion. This's definitely helpful/useful to me.

c633:
> Could we add some kind of typedef in ConiksDirectories? (I apologize if there's something obviously wrong about this question because I don't fully understand how Go works yet.) I also noticed that there's a policy.go in keyserver containing some information that's also contained in merkletree.policy. I think we need to find a way to dedup those.

I just give a try for this in b5907c7e2faa2cff4938056e20489ac40a401798. What do you think? If `server-reg-tb` is okay, I will merge it into this branch.

masomel:
> I just give a try for this in b5907c7. What do you think? If server-reg-tb is okay, I will merge it into this branch.

I think it's pretty good. I left some comments.

masomel:
Btw, please don't forget to add all the new packages with tests to the coveralls reports in .travis.yml.

masomel:
> Do you have any suggestion on #29 (comment)? 

I gave this a shot in https://github.com/coniks-sys/coniks-go/commit/0d06d54bda2514570ed909a69397126da3072a4d. What do you think?

c633:
How do you feel about 7e17fee4eb9e1721530f253585db7f2c3a47b01d?

masomel:
> How do you feel about 7e17fee? 

LGTM

keyserver/handler.go
masomel: > Do you mean after #37 is merged? 

Yes.

protocol/message.go
```diff
@@ -0,0 +1,46 @@
...
+
+type RegistrationResponse struct {
+	Type int
+	STR  *merkletree.SignedTreeRoot
+	AP   *merkletree.AuthenticationPath
+}
+
+// RegistrationResponseWithTB is used to replace RegistrationResponse
+// with addition TB field
+type RegistrationResponseWithTB struct {
```
arlolra: Is this struct necessary? Couldn't we just use `RegistrationResponse` but make `TB` `omitempty`?

protocol/encoding.go
```diff
@@ -0,0 +1,103 @@
...
+			Level:      ap.Leaf.Level(),
+			Index:      b64en(ap.Leaf.Index()),
+			Value:      b64en(ap.Leaf.Value()),
+			Commitment: b64en(ap.Leaf.Commitment()),
+			IsEmpty:    ap.Leaf.IsEmpty(),
+		},
+	})
+}
+
+func MarshalRegResponseWithTB(Type int, strEnc, apEnc, tbEnc []byte) ([]byte, error) {
```
arlolra: This is probably naive, but why is this function necessary? You've annotated the `RegistrationRequest` so that you can call `json.Marshal` on it.

masomel: This function is for the response, which isn't annotated I believe, not the request. I decided to add this function mostly since it made the code in listener.go a bit cleaner.

arlolra: Couldn't you annotate the response then?  This looks like you're defining a new type in order to marshall it, which seems like an additional unnecessary step.

keyserver/listener.go
```diff
@@ -0,0 +1,158 @@
...
+		case <-server.stop:
+			conn.Close()
+		case <-closed:
+		}
+	}()
+
+	conn.SetDeadline(time.Now().Add(5 * time.Second))
+
+	// handle request
+	buf := make([]byte, 4<<10)
```
Liamsi: Maybe that's a dump question but why not simply write 4096 instead of `4<<10` for the buffer size?

protocol/encoding.go
masomel: `RegistrationResponseWithTB` was defined before this function existed, and I took this code and moved it to `encoding.go` because it made more sense to me to handle all marshaling operations in the same place. 

I do agree though, that annotating the response would simplify things, and for consistency we should annotate all message types in `message.go`.

protocol/directory.go
```diff
@@ -0,0 +1,92 @@
...
+	useTBs   bool
+	tbs      map[string]*merkletree.TemporaryBinding
+	policies merkletree.Policies
+}
+
+func InitDirectory(epDeadline merkletree.TimeStamp, vrfKey *vrf.PrivateKey,
+	signKey sign.PrivateKey, dirSize uint64,
+	useTBs bool, capacity uint64) *ConiksDirectory {
+	d := new(ConiksDirectory)
+	d.policies = merkletree.NewPolicies(epDeadline, vrfKey)
```
arlolra: ``d.SetPolicies(epDeadline, vrfKey)`

protocol/directory.go
```diff
@@ -0,0 +1,92 @@
...
+		tb, err := d.pad.TB(uname, key)
+		if err != nil {
+			log.Printf(err.Error())
+			return nil, nil, ErrorInternalServer
+		}
+		d.tbs[uname] = tb
+		return ap, tb, Success
+	}
+
+	if err = d.pad.Set(uname, key); err != nil {
```
arlolra: This needs to be in the else block of the above (`useTBs`).  You've already done the `Set` in `pad.TB`.

Also, remove the print statement.

merkletree/policy.go
```diff
@@ -69,3 +48,7 @@ func (p *ConiksPolicies) Serialize() []byte {
 func (p *ConiksPolicies) vrfPrivate() *vrf.PrivateKey {
 	return p.vrfPrivateKey
 }
+
+func (p *ConiksPolicies) EpDeadline() TimeStamp {
```
arlolra: This getter seems redundant.

keyserver/server.go
```diff
@@ -0,0 +1,177 @@
...
+	directory *protocol.ConiksDirectory
+
+	stop     chan struct{}
+	waitStop sync.WaitGroup
+
+	db kv.DB
+
+	policies         *ServerPolicies
+	policiesFilePath string // should be final
+	policiesMutex    sync.Mutex
```
arlolra: `policiesMutex` doesn't seem to be used anymore.

keyserver/coniksserver/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"os"
+	"os/signal"
+
+	"github.com/BurntSushi/toml"
+	"github.com/coniks-sys/coniks-go/keyserver"
+	"github.com/coniks-sys/coniks-go/utils"
+)
+
+const ConfigFile = "config.toml"
+const PoliciesFile = "policies.toml"
```
arlolra: Why a separate policy file?

keyserver/coniksserver/main.go
```diff
@@ -0,0 +1,73 @@
...
+
+func main() {
+	genConfigPtr := flag.Bool("genconfig", false, "Generate server config")
```
arlolra: If there was only one config file, we could pass in a path for where to generate it,

```
flag.Bool("genconfig", "config.toml", "Generate server config")
```

and get rid of the globals (`ConfigFile`, `PoliciesFile`).

keyserver/server.go
```diff
@@ -0,0 +1,177 @@
...
+func LoadServerConfig(path string) *ServerConfig {
+	var conf ServerConfig
+	if _, err := toml.DecodeFile(path, &conf); err != nil {
+		log.Fatalf("Failed to load config: %v", err)
+		return nil
+	}
+	return &conf
+}
+
+func New(conf *ServerConfig) *ConiksServer {
```
arlolra: NewConiksServer ?

keyserver/server_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+	_, vrfKey, err := vrf.GenerateKey(nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var loadedHistoryLength uint64 = 100
+	var registrationCapacity uint64 = 100
+
+	server := new(ConiksServer)
```
arlolra: This stuff looks similar to what's in `New(ConiksServer)`.  Can we call the constructor here and make whatever needs to be different configurable while constructing?

keyserver/server.go
```diff
@@ -0,0 +1,177 @@
...
+	server.epochTimer = time.NewTimer(time.Duration(policies.EpochDeadline) * time.Second)
+	server.directory = protocol.InitDirectory(
+		policies.EpochDeadline, vrfKey,
+		sk, conf.LoadedHistoryLength,
+		true, conf.RegistrationCapacity)
+
+	return server
+}
+
+func (server *ConiksServer) RunWithConfig(conf *ServerConfig) {
```
arlolra: As opposed to just `Run` (which doesn't exist)?  You've passed in `conf` at construction time.  Should you be holding a pointer to it?

The question here is, is the `conf` expected to differ between construction and use?

keyserver/coniksserver/main.go
```diff
@@ -0,0 +1,73 @@
...
+		mkConfig()
+		return
+	}
+
+	// set up a CONIKS server from config file
+	conf := keyserver.LoadServerConfig(*configPathPtr)
+	serv := keyserver.New(conf)
+
+	// run the server until receiving an interrupt signal
+	serv.RunWithConfig(conf)
```
arlolra: Passing in the `conf` again seems unnecessary.

keyserver/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Usage
+Generate new config file with proper format
+```
+go run keyserver/coniksserver/main.go -genconfig
+```
+
+Generate new key pair for VRF and signing
+``` 
+go run utils/keygen/mkkey.go -vrf -signing
```
arlolra: Think I'd prefer to use this as,

```
go run keyserver/coniksserver/main.go -vrf -signing
```

utils/keygen/mkkey.go
```diff
@@ -0,0 +1,99 @@
...
+	if _, err := os.Stat(SECRET_KEY); err == nil {
+		fmt.Fprintf(os.Stderr, "%s already exists\n", SECRET_KEY)
+		return
+	}
+	if _, err := os.Stat(PUBLIC_KEY); err == nil {
+		fmt.Fprintf(os.Stderr, "%s already exists\n", PUBLIC_KEY)
+		return
+	}
+
+	pk, sk, err := ed25519.GenerateKey(rand.Reader)
```
arlolra: `sign.GenerateKey()`

Also, maybe our sign should take a reader like `vrf`.

utils/keygen/mkkey.go
```diff
@@ -0,0 +1,99 @@
...
+	"io/ioutil"
+	"log"
+	"os"
+
+	"github.com/coniks-sys/coniks-go/crypto/vrf"
+	"golang.org/x/crypto/ed25519"
+)
+
+const (
+	SECRET_KEY string = "ed25519.secret"
```
arlolra: Calling these "ed25519" goes against our 'hide the primitive' principal.  `sign.secret`, `sign.pub` would be better.

utils/keygen/mkkey.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"crypto/rand"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+
+	"github.com/coniks-sys/coniks-go/crypto/vrf"
+	"golang.org/x/crypto/ed25519"
```
arlolra: `github.com/coniks-sys/coniks-go/crypto/sign`

utils/keygen/mkkey.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+const (
+	SECRET_KEY string = "ed25519.secret"
+	PUBLIC_KEY string = "ed25519.pub"
+	VRF_SECRET string = "vrf.secret"
+	VRF_PUBLIC string = "vrf.pub"
+)
+
+func main() {
```
arlolra: As I said above, I don't think we want a separate bin for this.  It'd be nicer if the `coniksserver` had an option to produce the keys as well.

keyserver/coniksserver/main.go
```diff
@@ -0,0 +1,73 @@
...
+	"flag"
+	"os"
+	"os/signal"
+
+	"github.com/BurntSushi/toml"
+	"github.com/coniks-sys/coniks-go/keyserver"
+	"github.com/coniks-sys/coniks-go/utils"
+)
+
+const ConfigFile = "config.toml"
```
arlolra: We probably want to put,

```
config.toml
sign.pub
sign.secret
vrf.pub
vrf.secret
```

in the .gitignore file.

utils/keygen/mkkey.go
```diff
@@ -0,0 +1,99 @@
...
+
+func mkSigningKey() {
+	if _, err := os.Stat(SECRET_KEY); err == nil {
+		fmt.Fprintf(os.Stderr, "%s already exists\n", SECRET_KEY)
+		return
+	}
+	if _, err := os.Stat(PUBLIC_KEY); err == nil {
+		fmt.Fprintf(os.Stderr, "%s already exists\n", PUBLIC_KEY)
+		return
+	}
```
arlolra: Naive questions, why do we separate files for the pub/private parts?

utils/keygen/mkkey.go
```diff
@@ -0,0 +1,99 @@
...
+	"os"
+
+	"github.com/coniks-sys/coniks-go/crypto/vrf"
+	"golang.org/x/crypto/ed25519"
+)
+
+const (
+	SECRET_KEY string = "ed25519.secret"
+	PUBLIC_KEY string = "ed25519.pub"
+	VRF_SECRET string = "vrf.secret"
```
arlolra: Can we say private instead of secret?

merkletree/policy.go
```diff
@@ -1,8 +1,6 @@
 package merkletree
```
arlolra: Can you pull the changes to this file out into a separate pull?

protocol/directory.go
c633: > This needs to be in the else block of the above (useTBs). You've already done the Set in pad.TB.

Since I have returned in the `if` block, so `else` seems to be redundant?

arlolra:
This pull is very big and unwieldy.

I think you break it up into a few new pulls,

 * The changes to merkletree/policy.go and utils/util.go, which seem uncontroversial and can be merged right away.
 * The protocol/ package
 * The keyserver/ package, but when you make this pull, the base you choose is the pull for the protocol/ package so that we only see the diff between those two.

Then we can close this conversation with 180+ comments that's threatening to crash my browser.

keyserver/coniksserver/main.go
c633: I and @masomel  have had a discussion over this (https://github.com/coniks-sys/coniks-go/commit/b5907c7e2faa2cff4938056e20489ac40a401798#commitcomment-18595647). She pointed out for me that policy file was redundant.  I probably should remove this `policies.toml`. Thank you!

c633:
Thanks for all your feedback so far! I'm working on this!
> I think you should break it up into a few new pulls,

Yes, I totally agree!
> Then we can close this conversation with 180+ comments that's threatening to crash my browser.

Me too. That's my bad. I'm sorry!

protocol/directory.go
arlolra: Oops, I missed the return; you're right.  However, I think maybe the if/else is clearer, in that shows those two operations are disjoint.

```
if (useTBs) {
  pad.TB()
} else {
  pad.Set()
}
```

masomel:
> The keyserver/ package, but when you make this pull, the base you choose is the pull for the protocol/ package so that we only see the diff between those two.

One more question, are we going to wait to open this pull after the protocol pull is merged? It seems only pulls for the protocol and the policy/util stuff were created.

Issue #30, opened 2016-07-22
Identity management


c633:
Tor Messenger doesn't handout any identities, and it supports various protocols. In such a case, do we need to provide a way to manage identities of each protocol for the developer? 

In the current implementation of CONIKS for Tor Messenger, the CONIKS server stores all the identities of all protocols in the same tree, the key of each user leaf node is formatted as `username + '@' + protocol`. We probably want a better way.

Or the developers could have multi CONIKS server instances, each of them corresponding to a specific protocol. 

* Maybe the title of the issue is incorrect, so I'm glad if someone helps me replace it with a proper title.

masomel:
> Maybe the title of the issue is incorrect, so I'm glad if someone helps me replace it with a proper title.

I didn't think the title was incorrect, but I tried to make it a bit more specific. I hope it makes sense!

> Tor Messenger doesn't handout any identities, and it supports various protocols. In such a case, do we need to provide a way to manage identities of each protocol for the developer? 

This is a good question. I think that this is more of an issue of managing the identities of various services, and not various protocols. CONIKS should be protocol-agnostic, so I think managing all the identities for all the different services and protocols in a single directory should be fine. If the developers wanted to run multiple server instances for each protocol or service they support, they could, and this should work without any changes to our current design.

> In the current implementation of CONIKS for Tor Messenger, the CONIKS server stores all the identities of all protocols in the same tree, the key of each user leaf node is formatted as ``username + '@' + protocol``. We probably want a better way.

This does seem a bit hacky, but I think it's fine for now. Once we start adding support for other services such as XMPP servers, we might want to change this naming scheme. But as I said above, I don't have a problem with managing all identities in a single tree.

The only way I can see this becoming an issue is if one of those services (e.g. Twitter) decided to start running their own CONIKS server. Would Tor Messenger have to transfer all Twitter names it currently knows to Twitter? How would this transition look like? How can we ensure that the transition is safe, i.e. nothing about the protection of the names in Tor Messenger's directory should change when they are transferred over. This might be a separate problem than the one you're addressing here, but I think it's related, so if we think this is a real concern, I think it's worth looking into better ways for managing all these various identities.

c633:
> I didn't think the title was incorrect, but I tried to make it a bit more specific. I hope it makes sense!

Great. Thanks!

[...]

Issue #17, opened 2016-07-13
Add support for multiple data entries per key


masomel:
Each user leaf node should be able to map multiple data entries to the key. This is particularly useful for supporting multiple app instances (e.g. on multiple devices).

PR #33, opened 2016-07-24
Registration bots

### LINKED ISSUE #73: Use Unix sockets for communication between the bot and the server
TLS seems overkill when the two processes are on the same machine. If we use sockets, we also don't need to have an open port, which needs firewall protection.

c633:
- Should resolve #73 as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ hijacking secure communications without getting caught.
 ## Golang Library
 The packages in this library implement the various components of the CONIKS system and may be imported individually.
 
+- ``bots``: Registration bots for third-party account verification
 - ``client``: Client-side protocol cgo hooks
 - ``crypto``: Cryptographic algorithms and operations
 - ``merkletree``: Merkle prefix tree and related data structures
--- a/bots/bot.go
+++ b/bots/bot.go
@@ -6,6 +6,10 @@ import (
 	"net"
 )
 
+const (
+	messagePrefix = "?CONIKS?"
+)
+
 type BotHandleFunc func(string, []byte)
 
 type Bot interface {
--- a/bots/bot.go
+++ b/bots/bot.go
@@ -1,0 +1,39 @@
+package bots
+
+import (
+	"bytes"
+	"io"
+	"net"
+)
+
+type BotHandleFunc func(string, []byte)
+
+type Bot interface {
+	HandleRegistration(string, []byte) string
+	Run()
+	Stop()
+}
+
+func SendRequestToCONIKS(addr string, msg []byte) ([]byte, error) {
+	scheme := "unix"
+	unixaddr := &net.UnixAddr{Name: addr, Net: scheme}
+
+	conn, err := net.DialUnix(scheme, nil, unixaddr)
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Close()
+
+	_, err = conn.Write(msg)
+	if err != nil {
+		return nil, err
+	}
+
+	conn.CloseWrite()
+	var buf bytes.Buffer
+	if _, err := io.CopyN(&buf, conn, 8192); err != nil && err != io.EOF {
+		return nil, err
+	}
+
+	return buf.Bytes(), nil
+}
--- a/bots/coniksbot/cmd/run.go
+++ b/bots/coniksbot/cmd/run.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"os"
 	"os/signal"
-	"path"
 
 	"github.com/coniks-sys/coniks-go/bots"
 	"github.com/spf13/cobra"
@@ -19,18 +18,14 @@ This will look for config files with default names (botconfig.toml)
 in the current directory if not specified differently.
 	`,
 	Run: func(cmd *cobra.Command, args []string) {
-		dir := cmd.Flag("dir").Value.String()
 		config := cmd.Flag("config").Value.String()
-
-		confPath := path.Join(dir, config)
-		run(confPath)
+		run(config)
 	},
 }
 
 func init() {
 	RootCmd.AddCommand(runCmd)
-	runCmd.Flags().StringP("dir", "d", ".", "Location of bot working directory")
-	runCmd.Flags().StringP("config", "c", "botconfig.toml", "Configuration filename")
+	runCmd.Flags().StringP("config", "c", "botconfig.toml", "Path to bot configuration file")
 }
 
 func run(confPath string) {
--- a/bots/coniksbot/cmd/run.go
+++ b/bots/coniksbot/cmd/run.go
@@ -1,0 +1,47 @@
+package cmd
+
+import (
+	"os"
+	"os/signal"
+	"path"
+
+	"github.com/coniks-sys/coniks-go/bots"
+	"github.com/spf13/cobra"
+)
+
+// runCmd represents the run command
+var runCmd = &cobra.Command{
+	Use:   "run",
+	Short: "Run a CONIKS bot instance",
+	Long: `Run a CONIKS bot instance
+
+This will look for config files with default names (botconfig.toml) 
+in the current directory if not specified differently.
+	`,
+	Run: func(cmd *cobra.Command, args []string) {
+		dir := cmd.Flag("dir").Value.String()
+		config := cmd.Flag("config").Value.String()
+
+		confPath := path.Join(dir, config)
+		run(confPath)
+	},
+}
+
+func init() {
+	RootCmd.AddCommand(runCmd)
+	runCmd.Flags().StringP("dir", "d", ".", "Location of bot working directory")
+	runCmd.Flags().StringP("config", "c", "botconfig.toml", "Configuration filename")
+}
+
+func run(confPath string) {
+	bot, err := bots.NewTwitterBot(confPath)
+	if err != nil {
+		panic(err)
+	}
+
+	bot.Run()
+	ch := make(chan os.Signal, 1)
+	signal.Notify(ch, os.Interrupt)
+	<-ch
+	bot.Stop()
+}
--- a/bots/twitterbot.go
+++ b/bots/twitterbot.go
@@ -81,13 +81,13 @@ func (bot *TwitterBot) Run() {
 			return
 		}
 		// check if received DM has proper format
-		if strings.HasPrefix(dm.Text, "?CONIKS?") {
-			msg := strings.TrimPrefix(dm.Text, "?CONIKS?")
+		if strings.HasPrefix(dm.Text, messagePrefix) {
+			msg := strings.TrimPrefix(dm.Text, messagePrefix)
 			res := bot.HandleRegistration(dm.SenderScreenName, []byte(msg))
 			// Hackity, hack, hack!
 			// Twitter APIs probably doesn't want people call them so fast
 			time.Sleep(5 * time.Second)
-			err := bot.SendDM(dm.SenderScreenName, "?CONIKS?"+res)
+			err := bot.SendDM(dm.SenderScreenName, messagePrefix+res)
 			if err != nil {
 				log.Printf("[registration bot] " + err.Error())
 			}
--- a/bots/twitterbot.go
+++ b/bots/twitterbot.go
@@ -3,6 +3,7 @@ package bots
 import (
 	"fmt"
 	"log"
+	"os"
 	"strings"
 	"time"
 
@@ -41,6 +42,11 @@ func NewTwitterBot(path string) (Bot, error) {
 		return nil, fmt.Errorf("Failed to load config: %v", err)
 	}
 
+	// Notify if the CONIKS key server is down
+	if _, err := os.Stat(conf.CONIKSAddress); os.IsNotExist(err) {
+		return nil, fmt.Errorf("CONIKS Key Server is down")
+	}
+
 	auth := conf.TwitterOAuth
 	config := oauth1.NewConfig(auth.ConsumerKey, auth.ConsumerSecret)
 	token := oauth1.NewToken(auth.AccessToken, auth.AccessSecret)
@@ -49,6 +55,16 @@ func NewTwitterBot(path string) (Bot, error) {
 
 	// Twitter Client
 	client := twitter.NewClient(httpClient)
+	// Verify the tokens
+	handle, _, err := client.Accounts.VerifyCredentials(
+		&twitter.AccountVerifyParams{
+			IncludeEntities: twitter.Bool(false),
+			IncludeEmail:    twitter.Bool(true),
+		})
+	if err != nil ||
+		handle.ScreenName != conf.Handle {
+		return nil, fmt.Errorf("Could not authenticate you.")
+	}
 
 	bot := new(TwitterBot)
 	bot.client = client
@@ -104,7 +120,7 @@ func (bot *TwitterBot) HandleRegistration(username string, msg []byte) string {
 	} else {
 		request, ok := req.Request.(*p.RegistrationRequest)
 		if req.Type != p.RegistrationType || !ok ||
-			// issue: https://github.com/coniks-sys/coniks-go/issues/30
+			// FIXME: Agree on a convention in issues #17 / #30
 			!strings.EqualFold(strings.ToLower(username)+"@twitter", request.Username) {
 			invalid = true
 		}
--- a/bots/twitterbot.go
+++ b/bots/twitterbot.go
@@ -1,0 +1,140 @@
+package bots
+
+import (
+	"fmt"
+	"log"
+	"strings"
+	"time"
+
+	"github.com/BurntSushi/toml"
+	"github.com/coniks-sys/coniks-go/keyserver"
+	p "github.com/coniks-sys/coniks-go/protocol"
+	"github.com/dghubble/go-twitter/twitter"
+	"github.com/dghubble/oauth1"
+)
+
+type TwitterBot struct {
+	client        *twitter.Client
+	stream        *twitter.Stream
+	coniksAddress string
+	handle        string
+}
+
+var _ Bot = (*TwitterBot)(nil)
+
+type TwitterConfig struct {
+	CONIKSAddress string `toml:"coniks_address"`
+	TwitterOAuth  `toml:"twitter_oauth"`
+	Handle        string `toml:"twitter_bot_handle"`
+}
+
+type TwitterOAuth struct {
+	ConsumerKey    string
+	ConsumerSecret string
+	AccessToken    string
+	AccessSecret   string
+}
+
+func NewTwitterBot(path string) (Bot, error) {
+	var conf TwitterConfig
+	if _, err := toml.DecodeFile(path, &conf); err != nil {
+		return nil, fmt.Errorf("Failed to load config: %v", err)
+	}
+
+	auth := conf.TwitterOAuth
+	config := oauth1.NewConfig(auth.ConsumerKey, auth.ConsumerSecret)
+	token := oauth1.NewToken(auth.AccessToken, auth.AccessSecret)
+	// OAuth1 http.Client will automatically authorize Requests
+	httpClient := config.Client(oauth1.NoContext, token)
+
+	// Twitter Client
+	client := twitter.NewClient(httpClient)
+
+	bot := new(TwitterBot)
+	bot.client = client
+	bot.coniksAddress = conf.CONIKSAddress
+	bot.handle = conf.Handle
+
+	return bot, nil
+}
+
+func (bot *TwitterBot) Run() {
+	demux := twitter.NewSwitchDemux()
+	demux.DM = func(dm *twitter.DirectMessage) {
+		if strings.EqualFold(dm.SenderScreenName, bot.handle) {
+			return
+		}
+		// check if received DM has proper format
+		if strings.HasPrefix(dm.Text, "?CONIKS?") {
+			msg := strings.TrimPrefix(dm.Text, "?CONIKS?")
+			res := bot.HandleRegistration(dm.SenderScreenName, []byte(msg))
+			// Hackity, hack, hack!
+			// Twitter APIs probably doesn't want people call them so fast
+			time.Sleep(5 * time.Second)
+			err := bot.SendDM(dm.SenderScreenName, "?CONIKS?"+res)
+			if err != nil {
+				log.Printf("[registration bot] " + err.Error())
+			}
+		}
+	}
+
+	userParams := &twitter.StreamUserParams{
+		StallWarnings: twitter.Bool(true),
+	}
+	stream, err := bot.client.Streams.User(userParams)
+	if err != nil {
+		log.Fatal(err)
+	}
+	bot.stream = stream
+
+	// Receive messages until stopped or stream quits
+	go demux.HandleChan(stream.Messages)
+}
+
+func (bot *TwitterBot) Stop() {
+	bot.stream.Stop()
+}
+
+func (bot *TwitterBot) HandleRegistration(username string, msg []byte) string {
+	// validate request message
+	invalid := false
+	req, err := keyserver.UnmarshalRequest(msg)
+	if err != nil {
+		invalid = true
+	} else {
+		request, ok := req.Request.(*p.RegistrationRequest)
+		if req.Type != p.RegistrationType || !ok ||
+			// issue: https://github.com/coniks-sys/coniks-go/issues/30
+			!strings.EqualFold(strings.ToLower(username)+"@twitter", request.Username) {
+			invalid = true
+		}
+	}
+	if invalid {
+		log.Println("[registration bot] Malformed client request")
+		res, err := keyserver.MarshalResponse(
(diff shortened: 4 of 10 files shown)


bots/bot.go
```diff
@@ -0,0 +1,36 @@
...
+
+type Bot interface {
+	HandleRegistration(string, []byte) string
+	Run()
+	Stop()
+}
+
+func SendRequestToCONIKS(address string, msg []byte) ([]byte, error) {
+	conf := &tls.Config{
+		InsecureSkipVerify: true,
```
Liamsi: I would prefer to either use no TLS connection between the bot and the coniks server (as long as they run on the same machine) or to not use the `InsecureSkipVerify` mode (see https://golang.org/src/crypto/tls/common.go?s=#L302 for details)

bots/bot.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	conn, err := tls.Dial("tcp", address, conf)
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Close()
+
+	_, err = conn.Write(msg)
```
Liamsi: I didn't read the underlying `Write` function, but shouldn't we check that we actually wrote the whole message (and not only a part of it)?

bots/bot.go
c633: Thanks! `InsecureSkipVerify` is used for testing only.

The reason I would like to use TLS connection here is to prevent (say) malicious sysadmins from intercepting the connection between the bots and the server, and thus to preserve the privacy of the protocol.

bots/bot.go
c633: Probably should, I think.

c633: I just take a look at `dename` and `coname` implementation. It looks like they didn't check this, too. For example:
https://github.com/yahoo/coname/blob/master/httpfront/httpfront.go#L143-L145
https://github.com/andres-erbsen/chatterbox/blob/master/transport/transport.go#L141-L143

Liamsi: The first pointer only writes 2 chars and the 2nd method returns the number of written bytes to the callee. Oh, and actually, they ignore it their, too. But this doesn't really matter: we need to make sure we `Write` (and probably `Read`) what we want to send (and recv.). I skimmed through the tls code: https://golang.org/src/crypto/tls/conn.go#L718 Looks like it's OK to ignore the number of bytes written... Sorry for the noise.

protocol/encoding.go
```diff
@@ -0,0 +1,83 @@
...
+		Policies        json.RawMessage
+		Signature       []byte
+	}{
+		TreeHash:        str.Root(),
+		Epoch:           str.Epoch,
+		PreviousEpoch:   str.PreviousEpoch,
+		PreviousSTRHash: str.PreviousSTRHash,
+		Policies:        policies,
+		Signature:       str.Signature,
+	})
```
Liamsi: Should we put these currently anonymous structs into a separate file encoding.go?

protocol/encoding.go
```diff
@@ -0,0 +1,83 @@
...
+	type Leaf struct {
+		Level      int
+		Index      string
+		Value      string
+		IsEmpty    bool
+		Commitment string
+	}
+
+	return json.Marshal(&struct {
+		TreeNonce   string
```
Liamsi: Same as https://github.com/coniks-sys/coniks-go/pull/33/files#r73868093

keyserver/listener.go
```diff
@@ -0,0 +1,182 @@
...
+				}
+				res, e := json.Marshal(&struct {
+					STR json.RawMessage `json:"str"`
+					AP  json.RawMessage `json:"ap"`
+					TB  json.RawMessage `json:"tb"`
+				}{
+					STR: strEncoded,
+					AP:  apEncoded,
+					TB:  tbEncoded,
+				})
```
Liamsi: same as https://github.com/coniks-sys/coniks-go/pull/33/files#r73868093 ?

protocol/encoding.go
c633: Could you please elaborate on this? I'm not sure what you mean exactly. 

The source code of this pull is mostly in the `bots` directory. The others belong to #29 pull. Sorry for this noise :(

Liamsi: Whoops, I see they are already in a separate encoding.go file and you won't need them anywhere else. So it's fine! Sorry for the noise, too.

bots/bot.go
masomel: > The reason I would like to use TLS connection here is to prevent (say) malicious sysadmins from intercepting the connection between the bots and the server, and thus to preserve the privacy of the protocol.

But if the sysadmin has access to both TLS certs, the connection can still be intercepted, no?

README.md
```diff
@@ -19,6 +19,7 @@ hijacking secure communications without getting caught.
 ## Golang Library
 The packages in this library implement the various components of the CONIKS system and may be imported individually.
 
+- ``bots``: Registration bots for identity verification
```
masomel: I would change "identity" with "third-party account".

bots/bot.go
c633: Yes, you're right. But that chance for they to have access to both TLS certs is so small, I think. Hopefully, this would be an extra security layer for our implementation.

README.md
c633: Thanks. I updated it.

bots/bot.go
masomel: But if the sysadmin set up both the server and the registration bot, I think that the chance of them having access to both TLS certs isn't low, is it?

c633: Yes, this's also true. Hmm, so basically we shouldn't use TLS here?

c633:
I think this pull is already for reviewing. Maybe the concern is https://github.com/coniks-sys/coniks-go/pull/33#discussion_r73850206.
/cc @arlolra

bots/bot.go
c633: @arlolra suggested that we could use unix socket here instead (see #73 )

keyserver/listener.go
```diff
@@ -10,11 +10,16 @@ import (
 	. "github.com/coniks-sys/coniks-go/protocol"
 )
 
-func (server *ConiksServer) listenForRequests(ln *net.TCPListener, handler func(msg []byte) ([]byte, error)) {
+func (server *ConiksServer) listenForRequests(ln net.Listener, handler func(msg []byte) ([]byte, error)) {
```
c633: Should we have separate `listenForRequests` for `net.TCPListener` (which is the public connection) and `net.UnixListener` (which is the unix socket connection), to eliminate type switches below? This way, maybe we will have duplicate code but a better performance? 
/cc @arlolra @masomel @Liamsi

masomel: Hmm, the type switches do seem a bit clunky, but I'm not sure if separating the listeners would really give us a significant performance improvement. Maybe that way there would be some parallelism would could take advantage of?

bots/twitterbot.go
```diff
@@ -0,0 +1,150 @@
...
+			err := bot.SendDM(dm.SenderScreenName, "?CONIKS?"+res)
+			if err != nil {
+				log.Printf("[registration bot] " + err.Error())
+			}
+		}
+	}
+
+	userParams := &twitter.StreamUserParams{
+		StallWarnings: twitter.Bool(true),
+		With:          "followings",
```
masomel: If I've understood correctly, `With: "followings"` makes it possible to authenticate those who the user follows (from https://github.com/dghubble/go-twitter#user). Are we sure we need/want this?

c633: Yep, I don't remember the reason why I need this in the first place. I just removed it and the bot still works. Thanks!

keyserver/listener.go
```diff
@@ -10,11 +10,16 @@ import (
...
 	defer ln.Close()
 	go func() {
 		<-server.stop
-		ln.SetDeadline(time.Now())
+		switch ln.(type) {
```
arlolra: Can you do something like?

```
if l, ok := ln.(interface { SetDeadline() }); ok {
  l.SetDeadline()
}
```

c633: Absolutely. It's so nice. Thanks!

bots/twitterbot.go
```diff
@@ -0,0 +1,148 @@
...
+		log.Println("[registration bot] Malformed client request: " + err.Error())
+		res, err := keyserver.MarshalResponse(
+			p.NewErrorResponse(p.ErrorMalformedClientMessage))
+		if err != nil {
+			panic(err)
+		}
+		return string(res)
+	}
+
+	// create new CONIKS registration request
```
arlolra: Should you just use `keyserver.UnmarshalRequest` and then validate the request type and `strings.EqualFold(dm.SenderScreenName, req.Username)`?

bots/twitterbot.go
```diff
@@ -0,0 +1,148 @@
...
+	rc := p.Request{
+		Type: p.RegistrationType,
+		Request: p.RegistrationRequest{
+			// issue: https://github.com/coniks-sys/coniks-go/issues/30
+			Username: strings.ToLower(username) + "@twitter",
```
arlolra: Hmm, this is tricky because the lookups need it too.  It should already be set to https://github.com/arlolra/ctypes-otr/blob/master/chrome/content/coniks/coniks.js#L11 no?

keyserver/listener.go
arlolra: I think it's fine to leave this combined for now, until we have some evidence it's a performance bottleneck.

bots/twitterbot.go
c633: I'm not sure if we want to have `username` field in the request. Why do you prefer this way?

keyserver/testutil/testutil.go
```diff
@@ -126,3 +126,27 @@ func NewClient(t *testing.T, address string, msg []byte) ([]byte, error) {
 
 	return buf.Bytes(), nil
 }
+
+func NewBot(t *testing.T, msg []byte) ([]byte, error) {
```
arlolra: I see why you did this here, but you can split out the socket changes.

Maybe `NewTCPClient` and `NewUnixClient`.

keyserver/listener.go
c633: Yep, thank you all for yours advices.

bots/twitterbot.go
c633: > Although, does that lowercase? We probably need to agree on a convention.

I agree! I did this because I saw there is a `getCorrectHandle` in the `twitter.js` module (or something like this, that converts `conikstormess` to `ConiksTorMess`). I'm not sure if this function causes any inconsistency when we get the handle from, say, the conversation for lookup. In addition, if I add new Twitter contact with handle `Conikstormess`, I will get `conikstormess` in https://github.com/arlolra/ctypes-otr/blob/master/chrome/content/coniks/coniks.js#L46.

bots/twitterbot.go
arlolra: I just prefer reusing `UnmarshalRequest`.  The username isn't strictly needed, true, we can construct it, but it does establish the convention that's going to be used later in lookups, so maybe validating it ensures the client understands.

bots/twitterbot.go
arlolra: We need to lookup what `normalizedName` does.

arlolra: But, hmm https://github.com/mozilla/releases-comm-central/blob/master/chat/protocols/twitter/twitter.js#L471-L474

c633: This convention should be handled completely in the client. I totally agree with your comment https://github.com/coniks-sys/coniks-go/pull/33#discussion_r76509600.

bots/twitterbot.go
```diff
@@ -0,0 +1,140 @@
...
+	// validate request message
+	invalid := false
+	req, err := keyserver.UnmarshalRequest(msg)
+	if err != nil {
+		invalid = true
+	} else {
+		request, ok := req.Request.(*p.RegistrationRequest)
+		if req.Type != p.RegistrationType || !ok ||
+			// issue: https://github.com/coniks-sys/coniks-go/issues/30
+			(ok && strings.EqualFold(strings.ToLower(username)+"@twitter", request.Username)) {
```
arlolra: `ok` is is already assured by `|| !ok`

Shouldn't this be `!strings.EqualFold`?

bots/twitterbot_test.go
```diff
@@ -0,0 +1,55 @@
...
+	if response != `{"Error":14}` {
+		t.Error("Expected response", "got", response)
+	}
+}
+
+func TestInvalidRequestType(t *testing.T) {
+	username := "alice"
+	request := `
+{
+    "type": 1,
```
arlolra: RegistrationType is `0`

bots/twitterbot_test.go
```diff
@@ -0,0 +1,55 @@
...
+	if response != `{"Error":14}` {
+		t.Error("Expected response", "got", response)
+	}
+}
+
+func TestInvalidUsername(t *testing.T) {
+	username := "bob"
+	request := `
+{
+    "type": 1,
```
arlolra: RegistrationType is `0`

arlolra: Should you be using the var instead of a hardcoded number?

bots/twitterbot_test.go
c633: I'm testing for a invalid request type here.

bots/twitterbot_test.go
c633: I agree! Fixed in c89c5d573b703f001ff679bf5737ab3a49b50ec2.

bots/twitterbot_test.go
arlolra: Whooops!

bots/twitterbot.go
```diff
@@ -0,0 +1,140 @@
...
+	// validate request message
+	invalid := false
+	req, err := keyserver.UnmarshalRequest(msg)
+	if err != nil {
+		invalid = true
+	} else {
+		request, ok := req.Request.(*p.RegistrationRequest)
+		if req.Type != p.RegistrationType || !ok ||
+			// issue: https://github.com/coniks-sys/coniks-go/issues/30
+			!strings.EqualFold(strings.ToLower(username)+"@twitter", request.Username) {
```
arlolra: We need to patch Tor Messenger to remove https://github.com/mozilla/releases-comm-central/blob/master/chat/protocols/twitter/twitter.js#L471-L474 otherwise handles with underscores will never match.

Valid chars are `[a-z0-9_]{1,15}` https://support.twitter.com/articles/101299#error

Probably need to audit the other normalization functions as well to see what nastiness they're doing.

Also, "handle@twitter" seems ok but, for irc, we're going to end up with "handle@irc.net.work@irc".  Should we pick another char?

arlolra: > We need to patch Tor Messenger to remove

Here's a quick guide in case you're excited about doing that,

- clone https://github.com/mozilla/releases-comm-central
- commit your changes and `git format-patch`
- you can also do that with mercurial since you've been building your own Instantbird `hg export > something.patch`
- clone https://gitweb.torproject.org/tor-messenger-build.git (I like this as origin)
- optionally, `git remote add github https://github.com/TheTorProject/tor-messenger-build`
- add the patch in https://github.com/TheTorProject/tor-messenger-build/tree/master/projects/instantbird
- add the filename to the `input_files` https://github.com/TheTorProject/tor-messenger-build/blob/master/projects/instantbird/config#L54
- commit that change
- open a pull on github
- or, `git format-patch` that one, and create a new ticket https://trac.torproject.org/projects/tor/newticket with the patch attached

arlolra: > Here's a quick guide in case you're excited about doing that,

Actually, nevermind, some of this is bad advice.  The patch needs to apply to THUNDERBIRD_45_2_0_RELEASE in https://hg.mozilla.org/releases/comm-esr45/

c633: > The patch needs to apply to THUNDERBIRD_45_2_0_RELEASE in https://hg.mozilla.org/releases/comm-esr45/

In bug 955420 thread (https://bugzilla.mozilla.org/show_bug.cgi?id=955420), they said that the current twitter's normalize is for backwards compatibility. As I understand, the backwards compatibility is related to the logs. So I think this patch needs to apply to Tor Messenger?

Btw, `normalize: aString => aString.toLowerCase(),` seems to be correct and `normalize: aString => aString.replace(/[^a-z0-9_]/gi, "").toLowerCase(),` is also correct, right?

arlolra: > As I understand, the backwards compatibility is related to the logs. So I think this patch needs to apply to Tor Messenger?

Yup

> Btw, normalize: aString => aString.toLowerCase(), seems to be correct and normalize: aString => aString.replace(/[^a-z0-9_]/gi, "").toLowerCase(), is also correct, right?

Yup.  It's best to just remove the override and let it fallback to the default in jsProtoHelper.

> Do you have any proposal for this convention?

Not really

>  I think "handle@irc.net.work@irc" is fine for now.

Ok

arlolra: I did this in https://gitweb.torproject.org/tor-messenger-build.git/commit/?id=877c03082dadf7e503da5acedb2c88f5e7899eff because we're about to make a release tomorrow.

masomel: > Do you have any proposal for this convention? I think "handle@irc.net.work@irc" is fine for now.

I was thinking we could use a format "handle@service", which would look something like `handle@twitter.com` and `handle@irc.network.org`. Would this work? Is there a reason CONIKS needs to record the protocol?

Liamsi: I think the protocol is useful during registration and  after that could be useful if you want to use the same handle for different protocols (and with different public-keys).

arlolra: > Is there a reason CONIKS needs to record the protocol?

I'm not sure, but see https://trac.torproject.org/projects/tor/ticket/17461

Some networks let you use "handle@service", and SRV records disambiguate the protocol.

We could maybe handle this with #17 though.  Let's continue in #30

masomel: > Some networks let you use "handle@service", and SRV records disambiguate the protocol.
 
I see.

> We could maybe handle this with #17 though. Let's continue in #30

I agree.

> I think the protocol is useful during registration and after that could be useful if you want to use the same handle for different protocols (and with different public-keys).

In this case, "handle@protocol1" and "handle@protocol2" would be treated as different users, thoug, unless we do #17.

c633:
Can we merge this pull?

bots/coniksbot/cmd/run.go
```diff
@@ -0,0 +1,47 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		dir := cmd.Flag("dir").Value.String()
+		config := cmd.Flag("config").Value.String()
+
+		confPath := path.Join(dir, config)
+		run(confPath)
+	},
+}
+
+func init() {
```
arlolra: Can you make this like https://github.com/coniks-sys/coniks-go/blob/master/keyserver/coniksserver/cmd/run.go#L35

with a pid and dir removed.

bots/twitterbot.go
```diff
@@ -0,0 +1,140 @@
...
+func (bot *TwitterBot) HandleRegistration(username string, msg []byte) string {
+	// validate request message
+	invalid := false
+	req, err := keyserver.UnmarshalRequest(msg)
+	if err != nil {
+		invalid = true
+	} else {
+		request, ok := req.Request.(*p.RegistrationRequest)
+		if req.Type != p.RegistrationType || !ok ||
+			// issue: https://github.com/coniks-sys/coniks-go/issues/30
```
arlolra: Can you leave a fixme here

```
// FIXME:  Agree on a convention in issues #17 / #30
```

bots/twitterbot.go
```diff
@@ -0,0 +1,140 @@
...
+				log.Printf("[registration bot] " + err.Error())
+			}
+		}
+	}
+
+	userParams := &twitter.StreamUserParams{
+		StallWarnings: twitter.Bool(true),
+	}
+	stream, err := bot.client.Streams.User(userParams)
+	if err != nil {
```
arlolra: I'm surprised this doesn't complain if my access token is bogus.  Is there's any way to fail early if your credentials are no good?

bots/bot.go
```diff
@@ -0,0 +1,39 @@
...
+	HandleRegistration(string, []byte) string
+	Run()
+	Stop()
+}
+
+func SendRequestToCONIKS(addr string, msg []byte) ([]byte, error) {
+	scheme := "unix"
+	unixaddr := &net.UnixAddr{Name: addr, Net: scheme}
+
+	conn, err := net.DialUnix(scheme, nil, unixaddr)
```
arlolra: Similarly, on startup, it might be helpful to notify early if the coniks server is down.

bots/twitterbot.go
c633: Yeah, I forgot to check this. Fixed in 46c447206aa158cdae955057bce7e3bdd7b2c33d.

bots/coniksbot/cmd/run.go
```diff
@@ -0,0 +1,42 @@
...
+	`,
+	Run: func(cmd *cobra.Command, args []string) {
+		config := cmd.Flag("config").Value.String()
+		run(config)
+	},
+}
+
+func init() {
+	RootCmd.AddCommand(runCmd)
+	runCmd.Flags().StringP("config", "c", "botconfig.toml", "Path to bot configuration file")
```
arlolra: No PID here?

bots/twitterbot.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func NewTwitterBot(path string) (Bot, error) {
+	var conf TwitterConfig
+	if _, err := toml.DecodeFile(path, &conf); err != nil {
+		return nil, fmt.Errorf("Failed to load config: %v", err)
+	}
+
+	// Notify if the CONIKS key server is down
+	if !testCONIKSConnection(conf.CONIKSAddress) {
```
arlolra: I should probably just test, but what happens on the keyserver side here?  Just ignores the close?

bots/coniksbot/cmd/run.go
c633: Do you need a PID for the bot? I think no ... ?

arlolra: Ah, you mean cause there's no configs to reload.  Ya, I guess it's unnecessary.  Please ignore.

bots/twitterbot.go
c633: I'm finding a nicer way for this.

c633: Can we just do it?
```
if _, err := os.Stat(conf.CONIKSAddress); os.IsNotExist(err) {
    return nil, fmt.Errorf("CONIKS Key Server is down")
}
```

arlolra: Ya, stat seems less intrusive.  Let's do that.

Ultimately, we may want to reconsider what I said in #75.  Maybe opening a long-lived connection at the start with the keyserver and doing some "first byte sent indicates how many more bytes to read" protocol is the way to go.

Opening a issue to discuss that may be good.

We could use the same protocol for the public connections but close them after responding, rather than a loop.

Anyways, to discuss ...

bots/twitterbot.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func (bot *TwitterBot) Run() {
+	demux := twitter.NewSwitchDemux()
+	demux.DM = func(dm *twitter.DirectMessage) {
+		if strings.EqualFold(dm.SenderScreenName, bot.handle) {
+			return
+		}
+		// check if received DM has proper format
+		if strings.HasPrefix(dm.Text, "?CONIKS?") {
```
arlolra: Is it worth versioning this prefix?

We need some documentation about this expectation eventually.

bots/twitterbot.go
c633: I opened #77 for dicussions. 
/cc @Liamsi @masomel

arlolra:
> Can we merge this pull?

LGTM

bots/twitterbot.go
Liamsi: Nitpicking: make `"?CONIKS?"` a (not exported) constant (and replace hardcoded usages)?

Liamsi:
>> Can we merge this pull?

>LGTM

Besides some nitpicking (which you may ignore): LGTM, too.

bots/twitterbot.go
c633: Totally agree, should we put it into `protocol` or `bots`? I think it doesn't make sense to put this into `bots` but it's just being used in `bots`. What do you think?

Liamsi: I would leave it in `bots`. If it is only used there it makes sense, I think. Anyways, the package can be easily changed later.

c633: Fixed in 7bf9ca71795bb68360e165a2999731a49adce441.

masomel:
LGTM, too!

c633:
Merged in 90075472f0ae57de3e39cb328f027f69d68e4127.

PR #34, opened 2016-07-24
Add proof for STRs and TBs

c633:
* Resolve #24

* The `TB` also publishes the `IndexProof` after `vrf_integration` branch is merged.

Issue #24, opened 2016-07-20
Add verification function to STRs and TBs


masomel:
See https://github.com/coniks-sys/libmerkleprefixtree-go/pull/21#discussion_r71615961
Thus, it would make sense to expose a ``verify()`` or ``prove()`` function for STRs and temporary bindings.

PR #35, opened 2016-07-25
Public STR and TB fields for verification

arlolra:
Reopening for discussion here since I messed up #34

### DIFF
--- a/merkletree/pad.go
+++ b/merkletree/pad.go
@@ -120,8 +120,8 @@ func (pad *PAD) GetSTR(epoch uint64) *SignedTreeRoot {
 }
 
 func (pad *PAD) TB(key string, value []byte) (*TemporaryBinding, error) {
-	index, vrfProof := pad.computePrivateIndex(key, pad.policies.vrfPrivate())
-	tb := NewTB(pad.key, index, vrfProof, value, pad.latestSTR.sig)
+	index, _ := pad.computePrivateIndex(key, pad.policies.vrfPrivate())
+	tb := NewTB(pad.key, index, value, pad.latestSTR.Signature)
 	err := pad.tree.Set(index, key, value)
 	return tb, err
 }
--- a/merkletree/pad.go
+++ b/merkletree/pad.go
@@ -120,8 +120,8 @@ func (pad *PAD) GetSTR(epoch uint64) *SignedTreeRoot {
 }
 
 func (pad *PAD) TB(key string, value []byte) (*TemporaryBinding, error) {
-	index, indexProof := pad.computePrivateIndex(key, pad.policies.vrfPrivate())
-	tb := NewTB(pad.key, index, indexProof, value, pad.latestSTR.Signature)
+	index, vrfProof := pad.computePrivateIndex(key, pad.policies.vrfPrivate())
+	tb := NewTB(pad.key, index, vrfProof, value, pad.latestSTR.sig)
 	err := pad.tree.Set(index, key, value)
 	return tb, err
 }
--- a/merkletree/pad.go
+++ b/merkletree/pad.go
@@ -121,7 +121,7 @@ func (pad *PAD) GetSTR(epoch uint64) *SignedTreeRoot {
 
 func (pad *PAD) TB(key string, value []byte) (*TemporaryBinding, error) {
 	index, _ := pad.computePrivateIndex(key, pad.policies.vrfPrivate())
-	tb := NewTB(pad.key, index, value, pad.latestSTR.Signature)
+	tb := NewTB(pad.key, pad.latestSTR.Signature, index, value)
 	err := pad.tree.Set(index, key, value)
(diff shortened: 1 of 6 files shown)


merkletree/str.go
```diff
@@ -79,3 +79,15 @@ func (str *SignedTreeRoot) PreviousSTRHash() []byte {
 func (str *SignedTreeRoot) Signature() []byte {
 	return str.sig
 }
+
+func (str *SignedTreeRoot) LibVersion() []byte {
```
masomel: Shouldn't ``LibVersion()``, ``HashID()`` and ``EpochDeadlinePolicy()`` be in ``policy.go``?

merkletree/pad.go
```diff
@@ -122,6 +122,7 @@ func (pad *PAD) GetSTR(epoch uint64) *SignedTreeRoot {
 func (pad *PAD) TB(key string, value []byte) (*TemporaryBinding, error) {
 	str := pad.latestSTR
 	index, _ := pad.computePrivateIndex(key, pad.policies.vrfPrivate())
+
```
masomel: I think we either want to keep the VRF proof here, or return the TB with a proof of absence as I mentioned in https://github.com/coniks-sys/coniks-go/pull/29#discussion_r72140992, no?

merkletree/tb.go
```diff
@@ -1,6 +1,7 @@
 package merkletree
 
 type TemporaryBinding struct {
+	str   []byte
```
masomel: I'm wondering if we need to include a field for the STR signature in the TB if the STR struct also exposes the signature, and the server needs to return the STR to the client for verification anyway?

masomel:
Alternatively to having public fields in the TB and STR, what I had in mind when I opened #24 was to expose a ``Verify()`` function. What do you think of 9480b16?

masomel:
In my branch, I decided to leave the public fields that were added in 812caad since I'm not sure if the client will still need to access those fields using the getters.

merkletree/pad.go
c633: Yeah. That's why I said that this branch should be merged after `vrf_integration` branch. Sorry for the noise :(

merkletree/str.go
c633: These can change in epochs, thus I think it should be here, corresponding with policies of a specific STR. How does that sound?

merkletree/tb.go
c633: Yes. I just realize that the server will return STR proof, VRF proof, proof of absence and TB to the client. So this field should be removed.

c633:
> Alternatively to having public fields in the TB and STR, what I had in mind when I opened #24 was to expose a Verify() function. What do you think of db24bfa?

I got your idea. But the signing key cannot be used here, since I presume that the `Verify()` function should be called in the client-side only. 

> In my branch, I decided to leave the public fields that were added in 812caad since I'm not sure if the client will still need to access those fields using the getters.

Actually, I think the key server would use those getters to return the STR/TB to the client. The client doesn't use those getters directly.

masomel:
> the signing key cannot be used here, since I presume that the Verify() function should be called in the client-side only. 

Yes, the idea would be for the client to call ``Verify()``. I chose to use the ``SigningKey`` since that's what ``crypto.Verify()`` uses as well. If ``SigningKey`` is the server's key pair, I agree that ``Verify()`` shouldn't take it as a param, but then neither should the verification function in ``crypto``.

> Actually, I think the key server would use those getters to return the STR/TB to the client. The client doesn't use those getters directly. 

Ok, I see.

c633:
> Yes, the idea would be for the client to call Verify(). I chose to use the SigningKey since that's what crypto.Verify() uses as well. If SigningKey is the server's key pair, I agree that Verify() shouldn't take it as a param, but then neither should the verification function in crypto.

Sorry for that. I removed it in 3da95ef40a19094ad9f95f556676649c1deb2daf

masomel:
> Sorry for that. I removed it in 3da95ef

Yeah, with all the signature verification wrappers in place, I suppose we no longer need to export ``crypto``'s ``verify()``. I still think we want this function, to allow for the wrappers I created to work. I gave this a shot in d46c806. Please let me know what you think!

masomel:
> I suppose we no longer need to export ``crypto``'s ``verify()``. I don't think we need to remove it entirely if we want something like ``VerifySTR()`` to work.

Argh, this was silly of me. I forgot ``crypto`` is no longer a subpackage of ``merkletree``. Sorry for the noise.

masomel:
Like we've decided to keep the ``VerifyX()`` functions in the lib as we discussed in https://github.com/coniks-sys/coniks-go/commit/d46c806796a4f58b038fc1ee8794b2bccfd8e8aa#commitcomment-18452858, so I'll go ahead and merge in the ``str-tb-verify-idea`` branch and close this PR.

PR #38, opened 2016-07-28
Fix proof test per paper

c633:
Per paper section 3.2:
> That node will either be a leaf node at depth l with an index i ̸= j which matches j in the first l bits

### DIFF
--- a/merkletree/proof_test.go
+++ b/merkletree/proof_test.go
@@ -49,12 +49,13 @@ func authPathHash(ap *AuthenticationPath) []byte {
 }
 
 func verifyProof(t *testing.T, ap *AuthenticationPath, treeHash []byte, key string) {
-	hash := authPathHash(ap)
-	if !bytes.Equal(treeHash, hash) {
-		t.Fatal("Invalid proof")
+	// step 1. vrf_verify(i, alice) == true
+	// we probably want to use vrf.Verify() here instead
+	if !bytes.Equal(vrf.Compute([]byte(key), vrfPrivKey1), ap.LookupIndex) {
+		t.Error("VRF verify returns false")
 	}
 
-	// step 1. Verify the auth path of the returned leaf
+	// step 2. verify if it's a proof of inclusion/proof of absence
 	if bytes.Equal(ap.Leaf.Index(), ap.LookupIndex) {
 		// proof of inclusion
 		// make sure we got a userLeafNode
@@ -63,26 +64,23 @@ func verifyProof(t *testing.T, ap *AuthenticationPath, treeHash []byte, key stri
 		}
 	} else {
 		// proof of absence
-		// step 2. vrf_verify(i, alice) == true
-		// we probably want to use vrf.Verify() here instead
-		if !bytes.Equal(vrf.Compute([]byte(key), vrfPrivKey1), ap.LookupIndex) {
-			t.Error("VRF verify returns false")
-		}
-
-		// step 3. Check that where i and j differ is at bit l
+		// check if i and j match in the first l bits
 		indexBits := util.ToBits(ap.Leaf.Index())
 		lookupIndexBits := util.ToBits(ap.LookupIndex)
 
 		for i := 0; i < ap.Leaf.Level(); i++ {
 			if indexBits[i] != lookupIndexBits[i] {
-				t.Error("Invalid proof of absence. Expect indecies share the same prefix",
+				t.Error("Invalid proof of absence. Expect indices share the same prefix",
 					"lookup index: ", indexBits[:ap.Leaf.Level()],
 					"leaf index: ", lookupIndexBits[:ap.Leaf.Level()])
 			}
 		}
-		if indexBits[ap.Leaf.Level()+1] == lookupIndexBits[ap.Leaf.Level()+1] {
-			t.Error("Invalid proof of absence. Expect indecies differ is at bit", ap.Leaf.Level()+1)
-		}
+	}
+
+	// step 3. Verify the auth path of the returned leaf
+	hash := authPathHash(ap)
+	if !bytes.Equal(treeHash, hash) {
+		t.Fatal("Invalid proof")
 	}
 }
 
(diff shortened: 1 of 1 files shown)


Liamsi:
Looks good to me. I have one question, though.

merkletree/proof_test.go
```diff
@@ -49,12 +49,18 @@ func authPathHash(ap *AuthenticationPath) []byte {
...
+	// step 1. Verify the auth path of the returned leaf
 	hash := authPathHash(ap)
 	if !bytes.Equal(treeHash, hash) {
 		t.Fatal("Invalid proof")
 	}
 
-	// step 1. Verify the auth path of the returned leaf
+	// step 2. vrf_verify(i, alice) == true
+	// we probably want to use vrf.Verify() here instead
+	if !bytes.Equal(vrf.Compute([]byte(key), vrfPrivKey1), ap.LookupIndex()) {
```
Liamsi: Will we move to `vrf.Verify` on the client side? Wouldn't be now a good time to change it also in the tests?

Liamsi: OK, I though this would just mean changing above line to: 
`if !vrf.Verify(vrf.Public(vrfPrivKey1), []byte(key), ap.LookupIndex(),  ap.VrfProof())`
but then the tests fail. So maybe this is for another PR...

c633: > Will we move to vrf.Verify on the client side?

Yes. https://github.com/coniks-sys/coniks-go/blob/e84f7796ee728d6bdf339317e48fffb999821ef3/client/cgo/coniks.go#L67-L68

c633: > OK, I though this would just mean changing above line to: 
if !vrf.Verify(vrf.Public(vrfPrivKey1), []byte(key), ap.LookupIndex(), ap.VrfProof()) {
but then the test fails. So maybe this is for another PR...

Yup. You read my mind 👍 . I think we can do it in #16 ?

merkletree/proof_test.go
```diff
@@ -63,13 +69,7 @@ func verifyProof(t *testing.T, ap *AuthenticationPath, treeHash []byte, key stri
...
 	} else {
 		// proof of absence
-		// step 2. vrf_verify(i, alice) == true
-		// we probably want to use vrf.Verify() here instead
-		if !bytes.Equal(vrf.Compute([]byte(key), vrfPrivKey1), ap.LookupIndex()) {
-			t.Error("VRF verify returns false")
-		}
-
-		// step 3. Check that where i and j differ is at bit l
+		// step 3. Check that where i and j match in the first l bits
```
Liamsi: Shouldn't this be: "Check if i and j differ in the first l bits" (at least that is what is happening below). 
(Also there is a typo "indecies" here: https://github.com/coniks-sys/coniks-go/pull/38/files#diff-5b678c600e78be51602b1116649c7573L78)

merkletree/proof_test.go
Liamsi: > Yup. You read my mind 👍 . I think we can do it in #16 ?

Makes sense!

Liamsi: The reason for the failing test (as explained above, when changing to `vrf.Verify`) is that the `proof` isn't computed (and set in the `AuthenticationPath` when using `MerkleTree.Get(...)`.

c633: > The reason for the failing test (as explained above, when changing to vrf.Verify) is that the proof isn't computed (and set in the AuthenticationPath when using MerkleTree.Get(...).

Exactly.

merkletree/proof_test.go
c633: Awww, thanks!

masomel:
I'm wondering if the order of the auth path verification should be: verify VRF, verify if it's a proof of inclusion/proof of absence and then recompute the root hash of the auth path only if these two leaf checks pass? Maybe this isn't as important since this is just a test, but it makes more sense to me to spend the cycles rehashing the path only if we're sure that the leaf passes the necessary checks.

c633:
> I'm wondering if the order of the auth path verification should be: verify VRF, verify if it's a proof of inclusion/proof of absence and then recompute the root hash of the auth path only if these two leaf checks pass?

Agree! I also fixed it in `client-cgo` branch.

masomel:
LGTM. Thanks @c633!

merkletree/proof_test.go
```diff
@@ -142,15 +140,25 @@ func TestVerifyProof(t *testing.T) {
 	absentIndex := vrf.Compute([]byte("123"), vrfPrivKey1)
```
arlolra: L134-137 above are redundant now that `authPathHash` is moved inside `verifyProof`.

merkletree/proof_test.go
```diff
@@ -142,15 +140,25 @@ func TestVerifyProof(t *testing.T) {
...
 	proof = m.Get(absentIndex) // shares the same prefix with an empty node
 	verifyProof(t, proof, m.hash, "123")
-	authPathHash(proof)
 	if _, ok := proof.Leaf.(*emptyNode); !ok {
```
arlolra: This isn't true; the leaf doesn't have to be empty.

merkletree/proof_test.go
```diff
@@ -142,15 +140,25 @@ func TestVerifyProof(t *testing.T) {
...
-	}*/
+	key1 := "key1"
+	index1 := vrf.Compute([]byte(key1), vrfPrivKey1)
+	val1 := []byte("value1")
+	if err := m.Set(index1, key1, val1); err != nil {
+		t.Fatal(err)
+	}
+	m.recomputeHash()
+	absentIndex := vrf.Compute([]byte("a"), vrfPrivKey1)
+	proof := m.Get(absentIndex) // shares the same prefix with leaf node key1
```
arlolra: I'm not sure I understand this.

First off, all indices share a prefix of at least 1 bit with a node in the tree, right?  Since it starts out with two empty nodes for 1 and 0.

But I'm guessing you mean the index for `key1` has a shared prefix with `absentIndex`.  Is that guaranteed though?  Because `vrfPrivKey1` is randomly generated, that won't always be the case.  A simple way to verify would be to add an assertion that they do.  But I think that'll make you see that you want to randomly generate strings until you get one that does indeed share a prefix, then you can do that test.

merkletree/proof_test.go
```diff
@@ -63,26 +64,23 @@ func verifyProof(t *testing.T, ap *AuthenticationPath, treeHash []byte, key stri
...
 			if indexBits[i] != lookupIndexBits[i] {
-				t.Error("Invalid proof of absence. Expect indecies share the same prefix",
+				t.Error("Invalid proof of absence. Expect indices share the same prefix",
 					"lookup index: ", indexBits[:ap.Leaf.Level()],
 					"leaf index: ", lookupIndexBits[:ap.Leaf.Level()])
 			}
 		}
-		if indexBits[ap.Leaf.Level()+1] == lookupIndexBits[ap.Leaf.Level()+1] {
-			t.Error("Invalid proof of absence. Expect indecies differ is at bit", ap.Leaf.Level()+1)
-		}
```
arlolra: Yup, removing this is good.

merkletree/proof_test.go
c633: Done.

merkletree/proof_test.go
c633: Actually `vrfPrivKey1` is deterministic (https://github.com/coniks-sys/coniks-go/blob/master/merkletree/merkletree_test.go#L12-L13). Hope it explains your concern.

merkletree/proof_test.go
c633: It is since `vrfPrivKey1` is deterministic, right?

merkletree/proof_test.go
arlolra: > I'm not sure I understand this.

Clearly I didn't!  Thanks, I completely missed that.

merkletree/proof_test.go
arlolra: My bad; I was thinking in the general sense.

merkletree/proof_test.go
arlolra: > A simple way to verify would be to add an assertion that they do.

Is it worth asserting though?

c633: Yup, it makes a lot of senses.

c633: Added in 1df40dfa7e469e76626d2821a92b5feebe7cf8e0

PR #39, opened 2016-07-28
Client library for verifications

c633:
I merged branch `no-getset` since we can take the advantages of struct fields (e.g., https://github.com/coniks-sys/coniks-go/blob/9c253690f74c41d9763111bd660cbb8108236b86/client/cgo/coniks.go#L105-L107). 
I will fixed other getters when the #36 is clarified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 *.so
 *.dylib
 
+# Cgo compiled headers
+libconiks.h
+
 # Folders
 _obj
 _test
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ hijacking secure communications without getting caught.
 ## Golang Library
 The packages in this library implement the various components of the CONIKS system and may be imported individually.
 
+- ``client``: Client-side protocol cgo hooks
 - ``crypto``: Cryptographic algorithms and operations
 - ``merkletree``: Merkle prefix tree and related data structures
 - ``utils``: Utility functions 
--- a/client/cgo/cgotest.go
+++ b/client/cgo/cgotest.go
@@ -1,0 +1,245 @@
+package main
+
+/*
+int testVerifyVrf(unsigned char *pk, int pkSize,
+    unsigned char *m, int mSize,
+    unsigned char *index, int indexSize,
+    unsigned char *proof, int proofSize) {
+    return cgoVerifyVrf(pk, pkSize, m, mSize, index, indexSize, proof, proofSize);
+}
+
+int testVerifySignature(
+    unsigned char *pk, int pkSize,
+    unsigned char *m, int mSize,
+    unsigned char *sig, int sigSize) {
+    return cgoVerifySignature(pk, pkSize, m, mSize, sig, sigSize);
+}
+
+int testVerifyHashChain(
+	unsigned char *prevHash, int hashSize,
+    unsigned char *strSig, int sigSize) {
+    return cgoVerifyHashChain(prevHash, hashSize, strSig, sigSize);
+}
+
+int testVerifyAuthPath(
+    unsigned char *treeHash, int treeHashSize,
+    unsigned char *treeNonce, int treeNonceSize,
+    unsigned char *lookupIndex, int lookupIndexSize,
+    unsigned char **prunedTree, int prunedTreeSize, int hashSize,
+    int leafLevel,
+    unsigned char *leafIndex, int leafIndexSize,
+    unsigned char *leafCommitment, int leafCommitmentSize,
+    int isLeafEmpty) {
+    return cgoVerifyAuthPath(treeHash, treeHashSize,
+        treeNonce, treeNonceSize,
+        lookupIndex, lookupIndexSize,
+        prunedTree, prunedTreeSize, hashSize,
+        leafLevel,
+        leafIndex, leafIndexSize,
+        leafCommitment, leafCommitmentSize,
+        isLeafEmpty);
+}
+
+#cgo CFLAGS: -Wno-implicit-function-declaration
+*/
+import "C"
+import (
+	"bytes"
+	"testing"
+	"unsafe"
+
+	"github.com/coniks-sys/coniks-go/crypto/sign"
+	"github.com/coniks-sys/coniks-go/crypto/vrf"
+	"github.com/coniks-sys/coniks-go/merkletree"
+)
+
+func byteSliceToCpchar(buf []byte) *C.uchar {
+	ptr := unsafe.Pointer(&buf[0])
+	return (*C.uchar)(ptr)
+}
+
+func testVerifyVrf(t *testing.T) {
+	pk, sk, err := vrf.GenerateKey(nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	alice := []byte("alice")
+	aliceVRF, aliceProof := sk.Prove(alice)
+	if v := C.testVerifyVrf(byteSliceToCpchar(pk[:]), C.int(len(pk)),
+		byteSliceToCpchar(alice), C.int(len(alice)),
+		byteSliceToCpchar(aliceVRF), C.int(len(aliceVRF)),
+		byteSliceToCpchar(aliceProof), C.int(len(aliceProof))); v != 1 {
+		t.Error("cgoVrfVerify failed")
+	}
+}
+
+func testVerifySignature(t *testing.T) {
+	key, err := sign.GenerateKey()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	message := []byte("test message")
+	sig := key.Sign(message)
+
+	pk, ok := key.Public()
+	if !ok {
+		t.Errorf("bad PK?")
+	}
+
+	if v := C.testVerifySignature(byteSliceToCpchar(pk), C.int(len(pk)),
+		byteSliceToCpchar(message), C.int(len(message)),
+		byteSliceToCpchar(sig), C.int(len(sig))); v != 1 {
+		t.Error("cgoVerifySignature failed")
+	}
+}
+
+func testVerifyHashChain(t *testing.T) {
+	signKey, err := sign.GenerateKey()
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, vrfPrivKey, err := vrf.GenerateKey(nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	pad, err := merkletree.NewPAD(merkletree.NewPolicies(2, vrfPrivKey), signKey, 10)
+	if err != nil {
+		t.Fatal(err)
+	}
+	key1 := "key1"
+	val1 := []byte("value1")
+
+	if err := pad.Set(key1, val1); err != nil {
+		t.Fatal(err)
+	}
+	pad.Update(nil)
+	str0 := pad.GetSTR(0)
+	str1 := pad.GetSTR(1)
+
+	if v := C.testVerifyHashChain(byteSliceToCpchar(str1.PreviousSTRHash), C.int(len(str1.PreviousSTRHash)),
+		byteSliceToCpchar(str0.Signature), C.int(len(str0.Signature))); v != 1 {
+		t.Error("cgoVerifyHashChain failed")
+	}
+}
+
+func testVerifyAuthPath(t *testing.T) {
+	signKey, err := sign.GenerateKey()
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, vrfPrivKey, err := vrf.GenerateKey(bytes.NewReader(
+		[]byte("deterministic tests need 256 bit")))
+	if err != nil {
+		t.Fatal(err)
+	}
+	pad, err := merkletree.NewPAD(merkletree.NewPolicies(2, vrfPrivKey), signKey, 10)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	key1 := "key1"
+	val1 := []byte("value1")
+	key2 := "key2"
+	val2 := []byte("value2")
+	key3 := "key3"
+	val3 := []byte("value3")
+
+	if err := pad.Set(key1, val1); err != nil {
+		t.Fatal(err)
+	}
+	if err := pad.Set(key2, val2); err != nil {
+		t.Fatal(err)
+	}
+	if err := pad.Set(key3, val3); err != nil {
+		t.Fatal(err)
+	}
+	pad.Update(nil)
+
+	// proof of inclusion
+	proof, err := pad.Lookup(key3)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	isLeafEmpty := 0
+	if proof.Leaf.IsEmpty() {
+		isLeafEmpty = 1
+	}
+	if v := C.testVerifyAuthPath(byteSliceToCpchar(pad.GetLatestSTR().Root()), C.int(len(pad.GetLatestSTR().Root())),
+		byteSliceToCpchar(proof.TreeNonce), C.int(len(proof.TreeNonce)),
+		byteSliceToCpchar(proof.LookupIndex), C.int(len(proof.LookupIndex)),
+		(**C.uchar)(unsafe.Pointer(&proof.PrunedTree[0][0])), C.int(len(proof.PrunedTree)), C.int(len(proof.PrunedTree[0])),
+		C.int(proof.Leaf.Level()),
+		byteSliceToCpchar(proof.Leaf.Index()), C.int(len(proof.Leaf.Index())),
+		byteSliceToCpchar(proof.Leaf.Commitment()), C.int(len(proof.Leaf.Commitment())),
+		C.int(isLeafEmpty)); v != 1 {
+		t.Error("Verify proof of inclusion failed")
+	}
+
+	// proof of absence
+	proof, err = pad.Lookup("123")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	isLeafEmpty = 0
+	if proof.Leaf.IsEmpty() {
+		isLeafEmpty = 1
+	}
+	if v := C.testVerifyAuthPath(byteSliceToCpchar(pad.GetLatestSTR().Root()), C.int(len(pad.GetLatestSTR().Root())),
+		byteSliceToCpchar(proof.TreeNonce), C.int(len(proof.TreeNonce)),
+		byteSliceToCpchar(proof.LookupIndex), C.int(len(proof.LookupIndex)),
+		(**C.uchar)(unsafe.Pointer(&proof.PrunedTree[0][0])), C.int(len(proof.PrunedTree)), C.int(len(proof.PrunedTree[0])),
+		C.int(proof.Leaf.Level()),
+		byteSliceToCpchar(proof.Leaf.Index()), C.int(len(proof.Leaf.Index())),
+		nil, C.int(0),
+		C.int(isLeafEmpty)); v != 1 {
+		t.Error("Verify proof of absence failed")
+	}
+}
+
+func testVerifyProofOfAbsenceSamePrefix(t *testing.T) {
+	signKey, err := sign.GenerateKey()
+	if err != nil {
+		t.Fatal(err)
+	}
+	_, vrfPrivKey, err := vrf.GenerateKey(bytes.NewReader(
+		[]byte("deterministic tests need 256 bit")))
+	if err != nil {
+		t.Fatal(err)
+	}
+	pad, err := merkletree.NewPAD(merkletree.NewPolicies(2, vrfPrivKey), signKey, 10)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	key1 := "key1"
+	val1 := []byte("value1")
+
+	if err := pad.Set(key1, val1); err != nil {
+		t.Fatal(err)
+	}
+	pad.Update(nil)
+
+	// proof of inclusion
+	proof, err := pad.Lookup("a")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	isLeafEmpty := 0
+	if proof.Leaf.IsEmpty() {
+		isLeafEmpty = 1
+	}
+	if v := C.testVerifyAuthPath(byteSliceToCpchar(pad.GetLatestSTR().Root()), C.int(len(pad.GetLatestSTR().Root())),
+		byteSliceToCpchar(proof.TreeNonce), C.int(len(proof.TreeNonce)),
+		byteSliceToCpchar(proof.LookupIndex), C.int(len(proof.LookupIndex)),
+		(**C.uchar)(unsafe.Pointer(&proof.PrunedTree[0][0])), C.int(len(proof.PrunedTree)), C.int(len(proof.PrunedTree[0])),
+		C.int(proof.Leaf.Level()),
+		byteSliceToCpchar(proof.Leaf.Index()), C.int(len(proof.Leaf.Index())),
+		byteSliceToCpchar(proof.Leaf.Commitment()), C.int(len(proof.Leaf.Commitment())),
+		C.int(isLeafEmpty)); v != 1 {
+		t.Error("Verify proof of absence failed")
+	}
+}
--- a/client/cgo/verification.go
+++ b/client/cgo/verification.go
@@ -1,0 +1,122 @@
+package main
+
+import "C"
+
+import (
+	"bytes"
+	"unsafe"
+
+	"github.com/coniks-sys/coniks-go/crypto"
+	"github.com/coniks-sys/coniks-go/crypto/sign"
+	"github.com/coniks-sys/coniks-go/crypto/vrf"
+	"github.com/coniks-sys/coniks-go/merkletree"
+)
+
+// main is required to build a shared library, but does nothing
+func main() {}
+
+//export cgoVerifySignature
+func cgoVerifySignature(pk unsafe.Pointer, pkSize C.int,
+	message unsafe.Pointer, size C.int,
+	sig unsafe.Pointer, sigSize C.int) C.int {
+	if int(pkSize) != sign.PublicKeySize ||
+		int(sigSize) != sign.SignatureSize {
+		return 0
+	}
+	pkBytes := C.GoBytes(pk, pkSize)
+	messageBytes := C.GoBytes(message, size)
+	sigBytes := C.GoBytes(sig, sigSize)
+	key := sign.PublicKey(pkBytes)
+	if key.Verify(messageBytes, sigBytes) {
+		return 1
+	}
+	return 0
+}
+
+//export cgoVerifyVrf
+func cgoVerifyVrf(pk unsafe.Pointer, pkSize C.int,
+	m unsafe.Pointer, size C.int,
+	index unsafe.Pointer, indexSize C.int,
+	proof unsafe.Pointer, proofSize C.int) C.int {
+	if int(pkSize) != vrf.PublicKeySize ||
+		int(indexSize) != vrf.Size ||
+		int(proofSize) != vrf.ProofSize {
+		return 0
+	}
+	pkBytes := C.GoBytes(pk, pkSize)
+	mBytes := C.GoBytes(m, size)
+	vrfBytes := C.GoBytes(index, indexSize)
+	proofBytes := C.GoBytes(proof, proofSize)
+	var key vrf.PublicKey
+	copy(key[:], pkBytes)
+	if key.Verify(mBytes, vrfBytes, proofBytes) {
+		return 1
+	}
+	return 0
+}
+
+//export cgoVerifyHashChain
+func cgoVerifyHashChain(prevHash unsafe.Pointer, hashSize C.int,
+	savedStrSig unsafe.Pointer, sigSize C.int) C.int {
+	if int(hashSize) != crypto.HashSizeByte ||
+		int(sigSize) != sign.SignatureSize {
+		return 0
+	}
+	prevHashBytes := C.GoBytes(prevHash, hashSize)
+	sigBytes := C.GoBytes(savedStrSig, sigSize)
+	strHash := crypto.Digest(sigBytes)
+	if bytes.Equal(prevHashBytes, strHash) {
+		return 1
+	}
+	return 0
+}
+
+//export cgoVerifyAuthPath
+//must call cgoVerifyVrf first to verify the returned vrf index.
+func cgoVerifyAuthPath(treeHash unsafe.Pointer, treeHashSize C.int,
+	treeNonce unsafe.Pointer, treeNonceSize C.int,
+	lookupIndex unsafe.Pointer, indexSize C.int,
+	prunedHashes unsafe.Pointer, prunedSize C.int, hashSize C.int,
+	leafLevel C.int,
+	leafIndex unsafe.Pointer, leafIndexSize C.int,
+	leafCommitment unsafe.Pointer, commitmentSize C.int,
+	isLeafEmpty C.int) C.int {
+
+	if int(treeHashSize) != crypto.HashSizeByte ||
+		int(treeNonceSize) != crypto.HashSizeByte ||
+		int(indexSize) != vrf.Size ||
+		int(hashSize) != crypto.HashSizeByte ||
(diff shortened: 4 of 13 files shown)


arlolra:
This is probably just a personal preference thing, but merging in branches complicates rebasing.

Next time you want to depend on another branch, I'd suggest you rebase your change on top of that branch.  If that branch is already pushed, you can even create a pull request against that branch so that only the above changes are shown in the PR here.

c633:
> Next time you want to depend on another branch, I'd suggest you rebase your change on top of that branch. If that branch is already pushed, you can even create a pull request against that branch so that only the above changes are shown in the PR here.

Yeah, thanks for your advices!

.gitignore
```diff
@@ -2,6 +2,7 @@
 *.o
 *.a
 *.so
+*.dylib
```
arlolra: Should we add *.h here too, for libconiks.h?

client/cgo/cgotest.go
```diff
@@ -0,0 +1,149 @@
+package main
+
+/*
+int testVRFVerify() {
+    unsigned char pk[] = {28,149,90,106,83,99,227,227,211,89,29,250,220,141,17,203,254,130,127,134,175,133,156,164,83,58,139,228,79,64,252,75};
+    unsigned char m[] = {97,108,105,99,101};
+    unsigned char index[] = {96,43,9,174,212,169,113,184,98,128,48,210,186,218,171,145,115,82,29,23,177,126,63,169,232,125,248,165,91,134,76,32};
+    unsigned char proof[] = {213,244,36,141,83,73,206,202,94,3,40,174,237,156,101,140,222,18,24,211,136,24,74,63,246,211,142,186,255,123,73,12,33,252,200,161,18,208,124,150,91,42,229,156,96,225,154,226,206,143,161,1,141,188,34,7,124,45,222,189,89,116,36,3,233,126,71,41,140,29,36,34,138,167,217,239,16,210,51,125,28,127,7,34,89,219,181,82,86,154,10,155,159,144,158,179};
```
arlolra: It would be clearer (and easier to maintain) if you moved generating and serializing this data into the go `testVRFVerify` and passed it as arguments to `C.testVRFVerify`.

```
int testVRFVerify(unsigned char *pk, ...) {
```
At present, it's hard to tell what it is and why it's correct.

Same for all the tests below.

crypto/crypto.go
```diff
@@ -34,11 +35,11 @@ func Sign(key SigningKey, message []byte) []byte {
...
-	if !ok {
-		return false
-	}
+func SigningKeyToPublic(key SigningKey) []byte {
+	return ed25519.PrivateKey(key).Public().(ed25519.PublicKey)
+}
+
+func Verify(pk, message, sig []byte) bool {
 	return ed25519.Verify(pk, message, sig)
 }
```
arlolra: Since we already have a type, what if we did something like this instead?

```
--- a/crypto/crypto.go
+++ b/crypto/crypto.go
@@ -13,8 +13,6 @@ const (
        PrivateKeySize = 64
 )
 
-type SigningKey ed25519.PrivateKey
-
 func Digest(ms ...[]byte) []byte {
        h := sha3.NewShake128()
        for _, m := range ms {
@@ -25,21 +23,23 @@ func Digest(ms ...[]byte) []byte {
        return ret
 }
 
-func GenerateKey() (SigningKey, error) {
+type SigningKey ed25519.PrivateKey
+
+func NewSigningKey() (SigningKey, error) {
        _, sk, err := ed25519.GenerateKey(rand.Reader)
        return SigningKey(sk), err
 }
 
-func Sign(key SigningKey, message []byte) []byte {
+func (key SigningKey) Sign(message []byte) []byte {
        return ed25519.Sign(ed25519.PrivateKey(key), message)
 }
 
-func Verify(key SigningKey, message, sig []byte) bool {
-       pk, ok := ed25519.PrivateKey(key).Public().(ed25519.PublicKey)
-       if !ok {
-               return false
-       }
-       return ed25519.Verify(pk, message, sig)
+func (key SigningKey) Public() []byte {
+       return ed25519.PrivateKey(key).Public().(ed25519.PublicKey)
+}
+
+func (key SigningKey) Verify(message, sig []byte) bool {
+       return ed25519.Verify(key, message, sig)
 }
```

client/cgo/coniks.go
```diff
@@ -0,0 +1,182 @@
...
+	ap.LookupIndex = li
+	ap.PrunedTree = pt
+
+	if verifyProof(ap, leafi, leafc, int(leafLevel), int(isLeafEmpty), th) {
+		return 1
+	}
+	return 0
+}
+
+func computeLeafHash(ap *merkletree.AuthenticationPath,
```
arlolra: `computeLeafHash`, `authPathHash`, `verifyProof` are presumably the same from proof_test.go, and should be deduped.  Maybe just expose `merkletree.VerifyProof`?

crypto/crypto.go
masomel: I had a slightly different suggestion in https://github.com/coniks-sys/coniks-go/blob/str-tb-verify-idea/crypto/crypto.go#L17 declaring a type for the public key and changing the signature of ``Verify()`` accordingly. Not sure which approach is better/cleaner.

client/cgo/coniks.go
masomel: Completely agree on this one.

crypto/crypto.go
arlolra: I think your suggestion is compatible with the above change,

We'd just make these two additional changes,

```
-func (key SigningKey) Public() []byte {
+func (key SigningKey) Public() VerifKey {
```

```
-func Verify(pk, message, sig []byte) bool {
-func (pk VerifKey) Verify(message, sig []byte) bool {
```

where `Verify` is now a method of the `VerifKey` type.

masomel: LGTM. I see your changes also suggest that we make ``Sign`` a method of ``SigningKey``. That's interesting. I'll go and make the suggested changes to ``Verify`` in the ``str-tb-verify-idea`` branch so they're in there when @c633 merges is that branch with ``str-tb``.

client/cgo/coniks.go
c633: Yes, I also agree.

.gitignore
c633: Not sure if we have any *.h source code in future, but I think it's okay for now.

client/cgo/cgotest.go
c633: Yes. Will do it.

.gitignore
arlolra: If you want, you can just do the specific `libconiks.h`

client/cgo/cgotest.go
c633: I fixed it in 763c2649aba5826e3320c48b4eff3a78cfb688a9. I also made some changes on `AuthenticationPath` struct: converting `PrunedTree` from slice of slices to slice of arrays; and add method to get latest STR from `PAD` (ec3675936cd8e32df44c0f9b494c60aa2df82c23)

c633:
I think this ready for another review @arlolra @Liamsi @masomel

Liamsi:
Looks good to me!

client/cgo/cgotest.go
```diff
@@ -0,0 +1,235 @@
+package main
+
+/*
+int testVRFVerify(unsigned char *pk, int pkSize,
+    unsigned char *m, int mSize,
+    unsigned char *index, int indexSize,
+    unsigned char *proof, int proofSize) {
+    if (cgoVrfVerify(pk, pkSize, m, mSize,
```
arlolra: Couldn't you just `return cgoVrfVerify(..)`?  Why are these if blocks here?

c633: I just want to make sure the flow: converting from golang types to C types and then from C types to golang types works as expected.

arlolra: You don't trust the golang developers? :)

c633: No, it's not what I mean, haha. I mean my use of something like unsafe.Pointer and others ...
Ok, I agree that they could be simplified as you said.

c633: I just re-read my code and realized that I missed something. What I mean when using these C functions as bridges is that I want to make the same scenario as when the lib being used in the add-on with js-ctypes. That is the input should be type `unsigned char*` and passed to go functions that take arguments type `unsafe.Pointer`. Hopefully it makes sense to you.

c633: Oops, my bad :(. Thanks!

merkletree/proof.go
```diff
@@ -16,6 +23,75 @@ type ProofNode interface {
...
+		depth -= 1
+	}
+	return hash
+}
+
+// Should verify the vrf index first
+func VerifyAuthPath(ap *AuthenticationPath,
+	leafIndex, leafCommitment []byte, leafLevel int, isLeafEmpty bool,
+	treeHash []byte) bool {
+	// step 1. Verify the auth path of the returned leaf
```
masomel: Looks to me like the order of the auth path verification was changed back from what we had discussed in https://github.com/coniks-sys/coniks-go/pull/38#issuecomment-235991441. I.e. step 1 and step 2 should be flipped, no?

merkletree/proof.go
```diff
@@ -16,6 +23,75 @@ type ProofNode interface {
 	Commitment() []byte
 }
 
+func computeLeafHash(ap *AuthenticationPath,
+	leafIndex, leafCommitment []byte, leafLevel int, isLeafEmpty bool) (leafHash []byte) {
+	if isLeafEmpty {
```
masomel: If we're passing in ``ap`` to ``computeLeafHash()`` couldn't we use the index, commitment, level etc from the ``ap.Leaf`` directly? On the other hand, I think this would be a similar issue to what we discussed in https://github.com/coniks-sys/coniks-go/commit/d46c806796a4f58b038fc1ee8794b2bccfd8e8aa#commitcomment-18437585, so I don't know if we should be passing in any auth path pointer at all.

merkletree/proof.go
```diff
@@ -9,7 +9,7 @@ import (
 
 type AuthenticationPath struct {
 	TreeNonce   []byte
-	PrunedTree  [][]byte
+	PrunedTree  [][crypto.HashSizeByte]byte
```
masomel: This is a good idea!

merkletree/proof.go
c633: Thanks! I thought there were some problems with our change of the order. But maybe I was wrong, so I fixed the order as discussed. Correct me if there still has problems.

merkletree/proof.go
c633: > If we're passing in ap to computeLeafHash() couldn't we use the index, commitment, level etc from the ap.Leaf directly inside the function?

I think we cannot use it since we have to instantiate new `AuthenticationPath` from input arrays (https://github.com/coniks-sys/coniks-go/blob/7801e29c2bbee6962ce9f2fb6b9fac583bc66633/client/cgo/coniks.go#L106-L109), and `ap.Leaf` is just an interface instance. 

> On the other hand, I think this would be a similar issue to what we discussed in d46c806#commitcomment-18437585, so I don't know if we should be passing in any auth path pointer at all.

We didn't do any serialization so I think it's okay to pass an auth path pointer, instead passing a lot of arrays.

masomel: > ``ap.Leaf`` is just an interface instance. 

Oh that's right. You're right, we can't use it.

> we have to instantiate new ``AuthenticationPath`` from input arrays 

Isn't this the same issue we were discussing in d46c806 for STRs and TBs? Or is your point that the client doesn't have to re-serialize the AP (like for a signature in the case of the STR)? I know I'm being nitpicky, but I would like to be consistent across the codebase.

c633: > Or is your point that the client doesn't have to re-serialize the AP (like for a signature in the case of the STR)?

Yes, it is what I mean. I think what we're doing here is just packing inputs into a struct and pass it into the verification function, instead of passing a lot of parameters. Maybe it also has some advantages in performance (e.g., smaller stack frame ...) :D .

masomel: > Maybe it also has some advantages in performance (e.g., smaller stack frame ...) :D . 

Haha, this is true. I wonder if it's worth going back and changing the STR and TB verify functions for this reason. Maybe if we notice the memory footprint of these functions being too big, we should.

c633: > I wonder if it's worth going back and changing the STR and TB verify functions for this reason. Maybe if we notice the memory footprint of these functions being too big, we should.

In case of STR and TB verification, I think it is a trade-off here. We have either advantages in memory footprint or cpu computations, but not both. Am I right?

masomel: Right. I think which way we end up going will depend on the types of client devices we expect and which resource we think is more valuable.

c633: > I think which way we end up going will depend on the types of client devices we expect and which resource we think is more valuable.

Totally agree!

client/cgo/coniks.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/coniks-sys/coniks-go/crypto"
+	"github.com/coniks-sys/coniks-go/crypto/vrf"
+	"github.com/coniks-sys/coniks-go/merkletree"
+)
+
+// main is required to build a shared library, but does nothing
+func main() {}
+
+//export cgoEd25519SignatureVerify
+func cgoEd25519SignatureVerify(pk unsafe.Pointer, pkSize C.int,
```
arlolra: Maybe just call this `cgoSignatureVerify` to not expose the scheme?

client/cgo/coniks.go
```diff
@@ -0,0 +1,115 @@
...
+	if int(pkSize) < vrf.PublicKeySize ||
+		int(indexSize) < vrf.Size ||
+		int(proofSize) < vrf.ProofSize {
+		return 0
+	}
+	pkBytes := C.GoBytes(pk, pkSize)
+	mBytes := C.GoBytes(m, size)
+	vrfBytes := C.GoBytes(index, indexSize)
+	proofBytes := C.GoBytes(proof, proofSize)
+	if vrf.Verify(pkBytes, mBytes, vrfBytes, proofBytes) {
```
arlolra: Couple issues here.

In the `PAD`, we have `pad.key` and then `pad.Set(key, value)`.  The paper calls this `pad.Set(name, value)`, and I'd prefer we switch to that.  There're too many keys flying around.

That leads me to my second point.  `pad.key` is sometimes used for signing data, sometimes used for VRF functions.  However, `vrf.GenerateKey` seems to differ from `crypto.GenerateKey`.  Is that a problem?  If not, why all extra business in vrf?

What I think we should do is not expose vrf at all to pad or anywhere else.  We should have one api in crypto that abstracts this all away, and crypto will call vrf where necessary.

So, getting L48 here, pkBytes should be cast to a VerifKey (or whatever public key type we end up on), and then you'd call `pk.VerifyVRF` or something like that.

Hopefully the above make sense.

client/cgo/coniks.go
c633: Agree.

client/cgo/coniks.go
c633: > pad.key is sometimes used for signing data, sometimes used for VRF functions.

`pad.key` is just used for signing data.  `policies.vrfPrivKey` is used for VRF functions. So I think it's not really a problem here. I renamed `pad.key` to `pad.signKey` for clarifying.

> What I think we should do is not expose vrf at all to pad or anywhere else. We should have one API in crypto that abstracts this all away, and crypto will call vrf where necessary.
So, getting to L48 here, pkBytes should be cast to a VerifKey (or whatever public key type we end up on), and then you'd call pk.VerifyVRF or something like that.

Agree! But what if we use 2 different types of verification key for signature and VRF in the future? Then we probably cannot have only one VerifKey type. Should we have some APIs named: `crypto.SignatureGenerateKey`, `crypto.VrfGenerateKey` and key types: `SignatureVerifKey`, `VrfVerifKey`, then I'd call `pk.Verify`. What do you think?

arlolra: > pad.key is just used for signing data. policies.vrfPrivKey is used for VRF functions. So I think it's not really a problem here.

That sounds like a bug though.  The signing key needs to be in the policies.

Is there any reason we can't use the vrf key for signing?

arlolra: > Is there any reason we can't use the vrf key for signing?

That question was a little hasty considering you said,

> But what if we use 2 different types of verification key for signature and VRF in the future?

That's a good point.

arlolra: I think we should consult @masomel and @Liamsi on this one.

c633: I gave it a try in a786a02f0d7e22e6d1f4a3520bb98c10ff56b2c5. What do you think?

c633: > The signing key needs to be in the policies.

I'm not sure about that. According to the paper, it's not.

arlolra: Hmm, well I'm not sure how to rotate the signing key then.  I would have thought it would be included in the policies as like the `nextSigningKey` but I guess that's not the same as having the current signing key in there.  And I guess having the current one in there doesn't make all that much sense since that's what you're signing with ...

...  so there's some evidence that I'm not thinking too clearly right now.

To help people like me out, maybe we should rename `crypto.go` to `sign.go` and then flatten out vrf, like,

```
crypto/
 |- vrf.go
 |- sign.go
 |- util.go
```

Sorry for the poor review here :(
```

Comment (arlolra) [inline client/cgo/coniks.go]: > But what if we use 2 different types of verification key for signature and VRF in the future?

The paper does mention `SKd` and `Kvrf`, so yeah, these two should be distinct for exactly the reason you stated.  All the more reason to emphasize that in structures.

**Comment by c633** [comment]
> To help people like me out, maybe we should rename crypto.go to sign.go and then flatten out vrf, like,

Yup, I'm going to do it, too.
> Sorry for the poor review here :(

You always give me the best advices/comments!

File: client/cgo/coniks.go (hunk 10 above)
Comment (c633) [inline client/cgo/coniks.go]: > To help people like me out, maybe we should rename crypto.go to sign.go and then flatten out vrf, like,

Yup, I'm going to do it, too.
> Sorry for the poor review here :(

You always give me the best advices/comments!

Comment (arlolra) [inline client/cgo/coniks.go]: > I gave it a try in a786a02. What do you think?

Not bad.  You'll probably have a naming clash for `SecretKey / VerifKey` when you flatten things out.

> Should we have some APIs named: crypto.SignatureGenerateKey, crypto.VrfGenerateKey and key types: SignatureVerifKey, VrfVerifKey, then I'd call pk.Verify. What do you think?

I'm not crazy about calling these `Secret / Verif`.  I prefer the conventional`Public / Private` pair.  Also, the verb should come first, as in `GenerateSignKey`.  Maybe `PublicSignKey`, `PrivateSignKey`, `PublicVrfKey`, `PrivateVrfKey`?

Comment (c633) [inline client/cgo/coniks.go]: I just made some more changes for this issue in fa26172adc4ecd775d74f59dd0be968457a1b294.

Comment (arlolra) [inline client/cgo/coniks.go]: > Then we'd have sign.GenerateKey, vrf.GenerateKey and no need for prefix Vrf or Signature

Yup, that works too.  Sounds good.

Comment (c633) [inline client/cgo/coniks.go]: > I'm not crazy about calling these Secret / Verif. I prefer the conventionalPublic / Private pair. Also, the verb should come first, as in GenerateSignKey. Maybe PublicSignKey, PrivateSignKey, PublicVrfKey, PrivateVrfKey?

Agree!

**Comment by arlolra** [comment]
This is going to be annoying, but can you pull out all the crypto api changes into a separate pull, and then rebase this branch on that one.

**Comment by c633** [comment]
> This is going to be annoying, but can you pull out all the crypto api changes into a separate pull, and then rebase this branch on that one. You can include b966173 in that one.

I just open new pull for all changes in the crypto package as well as the proof `VerifyAuthPath` function. Hopefully it doesn't get messed up.

**Comment by c633** [comment]
Is this already for being merged?

**Comment by arlolra** [comment]
> Is this already for being merged?

I wasn't finished reviewing before we went down that rabbit hole last night.

But can you rebase it?

**Comment by masomel** [comment]
I have one final question. Would it be possible to rename ``cgo/coniks.go`` to something like ``cgo/verification.go``? I think this makes it more clear what the module implements.

**Comment by c633** [comment]
> I wasn't finished reviewing before we went down that rabbit hole last night.

Aw, sorry :(

 > Would it be possible to rename cgo/coniks.go to something like cgo/verification.go?

Yup, makes sense.

File: client/cgo/coniks.go (hunk 10 above)
Comment (masomel) [inline client/cgo/coniks.go]: >> The signing key needs to be in the policies.
>
> I'm not sure about that. According to the paper, it's not.

In §2.1, we write: "We assume a separate PKI exists for distributing providers’ public keys, which they use to sign authenticated bindings and to transform users’ names for privacy purposes." And later on in §3.3, we write that the VRF key could be part of the policy. 

However, the statement from §2.1 makes it sound like neither the signing key nor the VRF key *need* to be in the policy. In ``coniks-java``, we've been using the server's TLS key for signing the tree roots, so there is at least some established way of handling the rotating the signing key.

I don't know if this is an approach that we would want to take here as well. I actually like @arlolra's suggestion to include the signing key in the policy as well if it's going to be different than the server's TLS key. I'm wondering if we could use something like certificate transparency for our server keys (i.e. the signing key and the VRF key).

Comment (c633) [inline client/cgo/coniks.go]: Should we open new issue for this, since it's hard for people follow it here ...

File: client/cgo/cgotest.go [hunk 11]
```diff
@@ -0,0 +1,222 @@
...
+		(**C.uchar)(unsafe.Pointer(&proof.PrunedTree[0][0])), C.int(len(proof.PrunedTree)), C.int(len(proof.PrunedTree[0])),
+		C.int(proof.Leaf.Level()),
+		byteSliceToCpchar(proof.Leaf.Index()), C.int(len(proof.Leaf.Index())),
+		nil, C.int(0),
+		C.int(isLeafEmpty)); v != 1 {
+		t.Error("Verify proof of absence failed")
+	}
+}
+
+func testVerifyProofOfAbsenceSamePrefix(t *testing.T) {
```
masomel: Just to make sure I've understood this test correctly. The difference between this proof of absence test and the one in ``VerifyAuthPath`` is that the string "a" in this test definitely shares the same prefix as "key1", which will give you a user leaf node in the proof? While "123" in the previous test does not share the same prefix as "key1", "key2" or "key3", which will give you an empty node? I'm wondering if we should also be explicitly checking the leaf type in these tests. Tthough I know this is checked implicitly when the auth path is hashed, so maybe that's enough?

c633: Did you mean we need to assert the string "a" shares the same prefix as "key1"?

masomel: No, since this is done in ``merkletree.VerifyAuthPath()``, right? I mean, should we assert that ``proof.Leaf.IsEmpty()`` is false for "a" and true for "123" in ``testVerifyAuthPath()`` above.

client/cgo/coniks.go
masomel: Yes, I think you're right. I think #47 is related, though the question about the PKI for the server's keys I think is separate. I just opened #52 for this.

client/cgo/cgotest.go
c633: > No, since this is done in merkletree.VerifyAuthPath(), right?

Yes, it is. So if there is anything go wrong, the hash verification would return fail. So I think we don't need to explicitly assert it in these tests.

masomel: > So if there is anything go wrong, the hash verification would return false.

True. I guess, I thought this was an additional way to test that the hash verification is working. But using the hash verification is probably enough for this assertion.

README.md
```diff
@@ -19,6 +19,7 @@ hijacking secure communications without getting caught.
 ## Golang Library
 The packages in this library implement the various components of the CONIKS system and may be imported individually.
 
+- ``client``: Client-side protocol implementation
```
masomel: Should this explanation say something more like "Client-side protocol cgo (or C) hooks"?

c633: Thanks!

masomel:
LGTM.

client/cgo/verification.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+// main is required to build a shared library, but does nothing
+func main() {}
+
+//export cgoVerifySignature
+func cgoVerifySignature(pk unsafe.Pointer, pkSize C.int,
+	message unsafe.Pointer, size C.int,
+	sig unsafe.Pointer, sigSize C.int) C.int {
+	if int(pkSize) < sign.PublicKeySize {
```
arlolra: Should we define `SignatureSize = 64` in crypto/sign.go?

What's the intention of these quick check anyways?  Who are they benefiting?  We know the expected lengths of these buffers, except for `message`.  Passing in all these length is kind of verbose, but I guess you're trying to be safe and avoid the chance for an out-of-bounds read.

c633: > I guess you're trying to be safe and avoid the chance for an out-of-bounds read.

Yes, since we are dealing with `unsafe.Pointer`. Should I remove these checks?

client/cgo/cgotest.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+int testVerifySignature(
+    unsigned char *pk, int pkSize,
+    unsigned char *m, int mSize,
+    unsigned char *sig, int sigSize) {
+    return cgoVerifySignature(pk, pkSize, m, mSize, sig, sigSize);
+}
+
+int testHashChainInvalidSize() {
```
arlolra: What's going on in this test?  It kind seems like `cgoVerifyHashChain` should be taking `prevHash` (which comes from the data for the new STR) and `root` (which is the signature the client has stored, the previous STR) and then hashes `root` and does the byte comparison.

client/cgo/verification.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+//export cgoVerifyHashChain
+func cgoVerifyHashChain(prevHash unsafe.Pointer, root unsafe.Pointer, size C.int) C.int {
+	if int(size) != crypto.HashSizeByte {
+		return 0
+	}
+	prevHashBytes := C.GoBytes(prevHash, size)
+	rootBytes := C.GoBytes(root, size)
+	if bytes.Equal(prevHashBytes, rootBytes) {
```
arlolra: See above (https://github.com/coniks-sys/coniks-go/pull/39/files#r73228663), but shouldn't `rootBytes` be hashed?

client/cgo/cgotest.go
c633: It should be removed. Stupid me.

arlolra: Well, no, having a test for STR linear history audit is good.  And we need a test covering `cgoVerifyHashChain`.  Just seems incomplete.

c633: Yup, my mistake. I was coding this why reading the old version of the paper and definitely had some misunderstanding. I'm going to fix it now. Thanks a lot!

client/cgo/verification.go
arlolra: > Should I remove these checks?

Nah.  But add one for the `SignatureSize` as well.

client/cgo/cgotest.go
c633: I fixed it in d5f8a7b96348200f68d64805b6a84295e3d685dd.

client/cgo/verification.go
```diff
@@ -19,7 +19,8 @@ func main() {}
 func cgoVerifySignature(pk unsafe.Pointer, pkSize C.int,
 	message unsafe.Pointer, size C.int,
 	sig unsafe.Pointer, sigSize C.int) C.int {
-	if int(pkSize) < sign.PublicKeySize {
+	if int(pkSize) != sign.PublicKeySize ||
+		sigSize != sign.SignatureSize {
```
arlolra: `int(sigSize)`, for consistency

client/cgo/cgotest.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	if v := C.testVerifySignature(byteSliceToCpchar(pk), C.int(len(pk)),
+		byteSliceToCpchar(message), C.int(len(message)),
+		byteSliceToCpchar(sig), C.int(len(sig))); v != 1 {
+		t.Error("cgoVerifySignature failed")
+	}
+}
+
+func testVerifyHashChain(t *testing.T) {
```
arlolra: Nice work!

client/cgo/verification.go
masomel: I thought ``root`` here is the root hash from the STR, no?

c633: Yes, at first it is the root hash from the STR. But it's my mistake. I fixed it in https://github.com/coniks-sys/coniks-go/pull/39/commits/03cf2235a19f260429f7c3c18992bbefe738b1b9#diff-d6b3a9616f7584dffdfc72037ea88e83R20

masomel: Ah ok, thanks for clarifying.

client/cgo/verification.go
```diff
@@ -55,13 +56,15 @@ func cgoVerifyVrf(pk unsafe.Pointer, pkSize C.int,
 }
 
 //export cgoVerifyHashChain
-func cgoVerifyHashChain(prevHash unsafe.Pointer, root unsafe.Pointer, size C.int) C.int {
-	if int(size) != crypto.HashSizeByte {
+func cgoVerifyHashChain(prevHash unsafe.Pointer, hashSize C.int, strSig unsafe.Pointer, sigSize C.int) C.int {
```
masomel: I might rename ``strSig`` to ``prevStrSig`` (and maybe even ``prevHash`` to ``curPrevStrHash``) just to make it absolutely clear what it is that you're comparing. At first glance, it seems like you might be comparing the current ``prevStrHash`` with the current STR's signature.

c633: I'd prefer to keep `prevHash` and rename `strSig` to `savedStrSig`. Fixed in 4fdc1088889c6ae07e4d02b67887fcf635ea8d11

client/cgo/cgotest.go
```diff
@@ -0,0 +1,245 @@
...
+
+	// proof of inclusion
+	proof, err := pad.Lookup(key3)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	isLeafEmpty := 0
+	if proof.Leaf.IsEmpty() {
+		isLeafEmpty = 1
```
arlolra: At first glance, this should just be a fatal failure if empty.

But, maybe it exposes some future work.  These client test could probably be exercising the serialization methods.  And eventually the client is going to want to be able to call methods like `IsEmpty()` to act appropriate.  But we can cross those bridges when we get there.

I'm just thinking out loud.  Ignore me.

arlolra:
LGTM

client/cgo/cgotest.go
c633: > At first glance, this should just be a fatal failure if empty.

Can you elaborate on that?

> But, maybe it exposes some future work. These client test could probably be exercising the serialization methods. And eventually the client is going to want to be able to call methods like IsEmpty() to act appropriate. But we can cross those bridges when we get there.

I didn't get your idea yet. Hope to have some discussions/explanations from you in our next hangout.

arlolra: > Can you elaborate on that?

```
pad.Set(key3, val3);
...
proof, err := pad.Lookup(key3)
```

If `proof.Leaf.IsEmpty()`, something went seriously wrong, no?

> I didn't get your idea yet. Hope to have some discussions/explanations from you in our next hangout.

Sure.  As I said, ignore me :)

c633: > If proof.Leaf.IsEmpty(), something went seriously wrong, no?

I see. I did it with an assumption that the actual client would never know the received `proof` should have an empty leaf or a user leaf. He just does exactly these steps to verify the proof. And in this case, if `proof.Leaf.IsEmpty()` is true by somehow, then the verification should return false.

arlolra: Yeah, that's pretty much what I was getting at above.

Issue #23, opened 2016-07-20
Hash random bytes before putting them on the wire


arlolra:
In several places we do `rand.Read(buf);` where eventually `buf` will be sent to the client.  I'm suggesting we additionally hash `buf = H(buf)` after the read.  Is this a thing we should do?

See https://trac.torproject.org/projects/tor/ticket/17694

/cc @Liamsi

Liamsi:
I briefly consulted @Daeinar on his opinion here: He also agreed that it's a reasonable thing to do. I'll take care of that.

Liamsi:
I searched for `rand` through the project on the master branch. Most occurrences where only in tests. The only relevant occurrence seems to be the tree nonce (at least it's the one I found so far). 
The `TB` seems fine (doesn't reveal bytes from `rand`) same with `prevEpoch`.
Will we make use of the salt field in `userLeafNode` btw? (currently it is only used in a test).

I wonder if missed some other occurrences.

c633:
>  The only relevant occurrence seems to be the tree nonce (at least it's the one I found so far). 

How's about the `prevStrHash` in https://github.com/coniks-sys/coniks-go/blob/master/merkletree/pad.go#L51-L52?

> Will we make use of the salt field in userLeafNode btw? (currently it is only used in a test).

I think yes. This salt could be returned to the client to open the commitment. (cc @masomel)

Liamsi:
> How's about the prevStrHash in https://github.com/coniks-sys/coniks-go/blob/master/merkletree/pad.go#L51-L52?

I thought it was already hashed here (but I was wrong) https://github.com/coniks-sys/coniks-go/blob/master/merkletree/pad.go#L57 Will remove the `else` case (and always do the hash after before calling `NewSTR`. Thanks!

> I think yes. This salt could be returned to the client to open the commitment. (cc @masomel)

Then it will also need be hashed (and hashed again for the commitment).

c633:
> Then it will also need be hashed (and hashed again for the commitment).

Yes, agree.

masomel:
> Then it will also need be hashed (and hashed again for the commitment).

Sounds good to me!

PR #40, opened 2016-07-28
Hash random bytes before putting them on the wire

Liamsi:
As discussed in #23

### DIFF
--- a/merkletree/merkletree.go
+++ b/merkletree/merkletree.go
@@ -2,7 +2,6 @@ package merkletree
 
 import (
 	"bytes"
-	"crypto/rand"
 	"errors"
 
 	"github.com/coniks-sys/coniks-go/crypto"
@@ -26,7 +25,7 @@ type MerkleTree struct {
 
 func NewMerkleTree() (*MerkleTree, error) {
 	root := NewInteriorNode(nil, 0, []bool{})
-	nonce, err := makeRand()
+	nonce, err := crypto.MakeRand()
 	if err != nil {
 		return nil, err
 	}
@@ -95,7 +94,7 @@ func (m *MerkleTree) Get(lookupIndex []byte) *AuthenticationPath {
 func (m *MerkleTree) Set(index []byte, key string, value []byte) error {
 
 	// generate random per user salt
-	salt, err := makeRand()
+	salt, err := crypto.MakeRand()
 	if err != nil {
 		return err
 	}
@@ -219,13 +218,4 @@ func (m *MerkleTree) Clone() *MerkleTree {
 		root:  m.root.Clone(nil).(*interiorNode),
 		hash:  append([]byte{}, m.hash...),
 	}
-}
-
-func makeRand() ([]byte, error) {
-	r := make([]byte, crypto.HashSizeByte)
-	if _, err := rand.Read(r); err != nil {
-		return nil, err
-	}
-	// Do not directly reveal bytes from rand.Read on the wire
-	return crypto.Digest(r), nil
 }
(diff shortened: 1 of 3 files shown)


c633:
LGTM.

masomel:
Same, LGTM.

arlolra:
Since we're doing the same ceremony three times, add a helper?

```
func makeRand() []byte, error {
  rand := make([]byte, crypto.HashSizeByte)
  if _, err := rand.Read(rand); err != nil {
    return nil, err
  }
  // Do not directly reveal bytes from rand.Read on the wire
  return crypto.Digest(rand), nil
}
```

Liamsi:
> Since we're doing the same ceremony three times, add a helper?

We are doing it only twice (the exact same ceremony). But that's already enough for creating the helper I think. But we could hash the (already hashed `rand.Read` output from here: https://github.com/coniks-sys/coniks-go/pull/40/commits/929c98c7adc122c8af9779fbaac7483713b01a68#diff-475ca4a45242c730e6134d970fe64530R59) once again to be consistent with the 2 other calls (and hashing is "cheap" enough).

Do you think this should go into the `crypto` package? Currently, it will only be used in the `merkletree` package

masomel:
> Do you think this should go into the ``crypto`` package? Currently, it will only be used in the ``merkletree`` package

The ``computePrivateIndex()`` isn't in ``crypto`` and only used in ``merkletree``. But I think ``makeRand()`` is a bit more general, so I think putting it in ``crypto`` makes sense.

arlolra:
If you revert the changes you made to pad.go, it's three.  But I guess the double hash is cheap enough that it isn't really an issue.

Liamsi:
> If you revert the changes you made to pad.go, it's three. But I guess the double hash is cheap enough that it isn't really an issue.

OK, but it still unnecessary. Reverted it. Thanks @arlolra !

merkletree/merkletree.go
```diff
@@ -26,10 +25,11 @@ type MerkleTree struct {
 
 func NewMerkleTree() (*MerkleTree, error) {
 	root := NewInteriorNode(nil, 0, []bool{})
-	nonce := make([]byte, crypto.HashSizeByte)
-	if _, err := rand.Read(nonce); err != nil {
+	nonce, err := crypto.MakeRand()
+	if err != nil {
 		return nil, err
 	}
+	nonce = crypto.Digest(nonce)
```
arlolra: Remove :)

merkletree/pad.go
```diff
@@ -46,17 +45,18 @@ func NewPAD(policies Policies, key crypto.SigningKey, len uint64) (*PAD, error)
 
 // if policies is nil, the previous policies will be used
 func (pad *PAD) signTreeRoot(m *MerkleTree, epoch uint64) {
-	var prevStrHash []byte
+	var prevSTRhash []byte
```
arlolra: Maybe just call it `prevHash` to match the arguments to `NewSTR`.

merkletree/merkletree.go
Liamsi: Ooops!

merkletree/pad.go
Liamsi: Makes sense. Done.

arlolra: Except that you did `prevhash`, all lowercase :)

Liamsi: Damm. Sorry for the noise!

arlolra:
Merged in c091b275d7ce823253b591ed263a50004e025962.

@Liamsi Please see https://github.com/coniks-sys/coniks-go/pull/39#issuecomment-236002541

Liamsi:
> @Liamsi Please see #39 (comment)

Thanks. Will rebase next time.

PR #41, opened 2016-07-28
Rename DefaultPolicies to ConiksPolicies

c633:
As discussed in #6

### DIFF
--- a/merkletree/policy.go
+++ b/merkletree/policy.go
@@ -14,15 +14,15 @@ type Policies interface {
 	vrfPrivate() *[vrf.SecretKeySize]byte
 }
 
-type DefaultPolicies struct {
+type ConiksPolicies struct {
 	vrfPrivateKey *[vrf.SecretKeySize]byte
 	epochDeadline TimeStamp
 }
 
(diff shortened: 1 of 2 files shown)


merkletree/policy.go
```diff
@@ -14,23 +14,23 @@ type Policies interface {
 	vrfPrivate() *[vrf.SecretKeySize]byte
 }
 
-type DefaultPolicies struct {
+type ConiskPolicies struct {
```
masomel: typo: ``Conisk`` should be ``Coniks``.

c633: Oops :(

masomel: No problem!

masomel:
I think this is ready to be merged. The additional mandatory policy fields will be added as part of #35 (as we decided in https://github.com/coniks-sys/coniks-go/pull/37#issuecomment-235768526), right?

c633:
> The additional mandatory policy fields will be added as part of #35 (as we decided in #37 (comment)), right?

Right.

PR #43, opened 2016-07-29
Stress tests

### LINKED ISSUE #16: Stress test adding data to the tree

Liamsi:
will resolve #16

### DIFF
--- a/merkletree/pad_test.go
+++ b/merkletree/pad_test.go
@@ -259,10 +259,11 @@ func TestTB(t *testing.T) {
 	tbb = append(tbb, tb.Index...)
 	tbb = append(tbb, tb.Value...)
 
-	// TODO use SigningKeyToPublic from crypto as soon as it is merged:
-	// pub := ed25519.PrivateKey(pad.key[:]).Public().(crypto.SigningKey)
-	// TODO Verify shouldn't use private key (related to above comment)
-	if !crypto.Verify(pad.key, tbb, tb.Signature) {
+	pk, ok := pad.key.Public()
+	if !ok {
+		t.Fatal("Couldn't retrieve public-key.")
+	}
+	if !pk.Verify(tbb, tb.Signature) {
 		t.Fatal("Couldn't validate signature")
 	}
 	// create next epoch and see if the TB is inserted as promised:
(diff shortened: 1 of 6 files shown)


c633:
Should we also resolve #18 ?

masomel:
> Should we also resolve #18 ?

I think we should do this in a separate PR since I expect that we'll have to move code around quite a bit for this.

Liamsi:
> Should we also resolve #18 ?

Some of the code will be re-usable to tackle #18. Will try to keep that in mind.

crypto/crypto.go
```diff
@@ -14,7 +14,7 @@ const (
 )
 
 type SigningKey ed25519.PrivateKey
-type VerifKey ed25519.PublicKey
+type VerifyKey ed25519.PublicKey
```
arlolra: This stuff doesn't look like belongs in this pull.

You should `git rebase -i master` and `d` (drop) that commit.

Liamsi: Should be OK now.

merkletree/pad_test.go
```diff
@@ -235,3 +239,194 @@ func TestPoliciesChange(t *testing.T) {
...
+	pad, err := NewPAD(NewPolicies(3, vrfPrivKey1), signKey, 3)
+	if err != nil {
+		t.Fatal(err)
+	}
+	tb, err := pad.TB(key1, val1)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// TODO shouldn't there be a serialize function?
```
c633: Can we use this function? https://github.com/coniks-sys/coniks-go/blob/master/merkletree/tb.go#L21-L27

merkletree/str.go
```diff
@@ -27,9 +21,6 @@ type SignedTreeRoot struct {
 }
 
 func NewSTR(key sign.PrivateKey, policies Policies, m *MerkleTree, epoch uint64, prevHash []byte) *SignedTreeRoot {
-	if epoch < 0 {
```
c633: Good catch. Thanks!

utils/util_test.go
```diff
@@ -27,3 +28,35 @@ func TestBitsBytesConvert(t *testing.T) {
...
 }
+
+func TestIntToBytes(t *testing.T) {
+	numInt := 42
+	b := IntToBytes(numInt)
+	if int(binary.LittleEndian.Uint32(b)) != numInt {
+		t.Fatal("Conversion to bytes looks wrong!")
+	}
+	// TODO It is possible to call `IntToBytes` with a signed int < 0
+	// isn't that problematic?
```
masomel: Good point. I don't expect the server will use signed ints. Do you think ``IntToBytes()`` should throw an error when it receives a negative value?

Liamsi: I think to let it `panic` or only use `UIntToBytes` would make sense. As far as I can see `node.level` can only grow (or never is a value subtracted from level). Maybe we could even refactor `level` to be unsigned?

merkletree/pad_test.go
Liamsi: Makes sense! Thanks for the pointer.

utils/util_test.go
masomel: Refactoring ``level`` to be unsigned makes a lot of sense to me. The only scenario in which level could decrease is when the tree is re-randomized after the VRF key is changed. But this should still work with an unsigned ``level``.

merkletree/pad_test.go
Liamsi: Now there is even a public `Serialize` method (https://github.com/coniks-sys/coniks-go/blob/master/merkletree/tb.go#L29-L31). I would prefer to use it instead.

utils/util_test.go
Liamsi: BTW unsigned inters can overflow in golang without panic/error or something similar: https://golang.org/ref/spec#Integer_overflow

Liamsi: > Refactoring level to be unsigned makes a lot of sense to me. The only scenario in which level could decrease is when the tree is re-randomized after the VRF key is changed. But this should still work with an unsigned level.

OK, should this go into another issue/PR?

c633: > OK, should this go into another issue/PR?

I think yes.

merkletree/pad_test.go
c633: Makes sense to me!

utils/util_test.go
Liamsi: OK, thanks! Opened https://github.com/coniks-sys/coniks-go/issues/53 for that.

arlolra:
Was this done?

c633:
Oops. I definitely did something terrible. Sorry for the noise.

c633:
Please see #54 . I'm so sorry about that :(

Liamsi:
> Oops. I definitely did something terrible. Sorry for the noise.
> Please see #54 . I'm so sorry about that :(

Please, don't worry. Luckily, it easy to revert using git(hub) as you already did :-)

PR #48, opened 2016-08-02
Refactor crypto subpackage as discussion in #39

c633:
* See https://github.com/coniks-sys/coniks-go/pull/39#discussion_r73082470

PR #49, opened 2016-08-02
Add auth path verification

c633:
This pull should be merged before #48


arlolra:
Thanks!  Merged since this seems uncontroversial, and others have had a look in #39

Issue #53, opened 2016-08-02
Use unsigned integers when dealing with non-negative values


Liamsi:
Discussion started here https://github.com/coniks-sys/coniks-go/pull/43#discussion_r73206002 (refactor `node.level` to an unsigned integer value *or* panic/err if methods that expect a non-negative int receive a negative on; find out which approach is safer/faster/more reasonable).

Also, clarify if and how we want to deal with overflows (also in case of uint(32/64) if we go with unsigned values).

Liamsi:
An interesting read on the signed/unsigned question: http://www.soundsoftware.ac.uk/c-pitfall-unsigned

masomel:
Good read! I did some small tests, and it seems like ``uint32``and ``uint64`` have the same underflow issue as described in the article. Given this, I think it might be safer to panic/err if methods that expect a non-negative int receive a negative one.

PR #54, opened 2016-08-02
Stress tests

c633:
Reopening for discussion here since I messed up #43 .

### DIFF
--- a/merkletree/pad_test.go
+++ b/merkletree/pad_test.go
@@ -259,10 +259,11 @@ func TestTB(t *testing.T) {
 	tbb = append(tbb, tb.Index...)
 	tbb = append(tbb, tb.Value...)
 
-	// TODO use SigningKeyToPublic from crypto as soon as it is merged:
-	// pub := ed25519.PrivateKey(pad.key[:]).Public().(crypto.SigningKey)
-	// TODO Verify shouldn't use private key (related to above comment)
-	if !crypto.Verify(pad.key, tbb, tb.Signature) {
+	pk, ok := pad.key.Public()
+	if !ok {
+		t.Fatal("Couldn't retrieve public-key.")
+	}
+	if !pk.Verify(tbb, tb.Signature) {
 		t.Fatal("Couldn't validate signature")
 	}
 	// create next epoch and see if the TB is inserted as promised:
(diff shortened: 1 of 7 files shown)


utils/util_test.go
```diff
@@ -27,3 +28,35 @@ func TestBitsBytesConvert(t *testing.T) {
...
 	}
 }
+
+func TestIntToBytes(t *testing.T) {
+	numInt := 42
+	b := IntToBytes(numInt)
+	if int(binary.LittleEndian.Uint32(b)) != numInt {
+		t.Fatal("Conversion to bytes looks wrong!")
+	}
+	// TODO It is possible to call `IntToBytes` with a signed int < 0
```
c633: Continue the discussion in https://github.com/coniks-sys/coniks-go/pull/43#discussion_r73206002.

Do you have any explanation for this? I really don't know the reason why Golang developers do this. Even if I try `int(4294967254) == -42`, it still returns false.

Liamsi: I thought this will be discussed (and resolved) in #53 ?

c633: Yup, agree!

utils/util_test.go
```diff
@@ -27,3 +28,35 @@ func TestBitsBytesConvert(t *testing.T) {
...
+	// isn't that problematic?
+	// for instance this will fail:
+	// b := IntToBytes(-42)
+	// if int(binary.LittleEndian.Uint32(b)) != numInt {
```
c633: I think it should be `if int32(binary.LittleEndian.Uint32(b)) != numInt`. If so, the test will pass. I forgot that `int` int 64-bit system is 8 bytes.

c633: I think it should be `if int32(binary.LittleEndian.Uint32(b)) != numInt`. If so, the test will pass. I forgot that `int` int 64-bit system is 8 bytes.

Liamsi: Good catch, thanks! Changed in latest commit.

Liamsi: same with `TestLongToBytes` below.

utils/util_test.go
```diff
@@ -27,3 +28,37 @@ func TestBitsBytesConvert(t *testing.T) {
...
+
+func TestIntToBytes(t *testing.T) {
+	numInt := 42
+	b := IntToBytes(numInt)
+	if int(binary.LittleEndian.Uint32(b)) != numInt {
+		t.Fatal("Conversion to bytes looks wrong!")
+	}
+	numInt = -42
+	b = IntToBytes(numInt)
+	if int32(binary.LittleEndian.Uint32(b)) != int32(numInt) {
```
c633: Can it be simplified to `int32(binary.LittleEndian.Uint32(b)) != numInt`?

Liamsi: Yep, pushed a second ago.

merkletree/pad_test.go
```diff
@@ -235,3 +239,190 @@ func TestPoliciesChange(t *testing.T) {
...
+		t.Fatal("Couldn't retrieve public-key.")
+	}
+	tbb := tb.Serialize(pad.latestSTR.Signature)
+	if !pk.Verify(tbb, tb.Signature) {
+		t.Fatal("Couldn't validate signature")
+	}
+	// create next epoch and see if the TB is inserted as promised:
+	pad.Update(nil)
+
+	ap, err := pad.Lookup(key1)
```
c633: Should we also verify the proof of inclusion here, just to make this test as a sample for complete TB verification?

utils/util_test.go
Liamsi: Wait, I changed another line; but here it's needed or not? `int32(binary.LittleEndian.Uint32(b)) != numInt` won't work because the type of `numInt` is an `int` and the type of `int32(binary.LittleEndian.Uint32(b))` is an `int32`

c633: You're right. Sorry for the noise.

merkletree/pad_test.go
Liamsi: Like in https://github.com/coniks-sys/coniks-go/pull/54/commits/91a572537bfa1ccb8d57d590f2a2604a8cf47782?

merkletree/pad_test.go
```diff
@@ -264,10 +264,23 @@ func TestTB(t *testing.T) {
 	// create next epoch and see if the TB is inserted as promised:
 	pad.Update(nil)
 
-	ap, err := pad.Lookup(key1)
+	ap, err := pad.Lookup(key)
 	if !bytes.Equal(ap.LookupIndex, tb.Index) || !bytes.Equal(ap.Leaf.Value(), tb.Value) {
 		t.Error("Value wasn't inserted as promised")
 	}
+	// step 1. verify VRF index
+	if !bytes.Equal(vrfPrivKey1.Compute([]byte(key)), ap.LookupIndex) {
```
c633: I think we should verify the VRF index in return `TB` first (after verifying the signature), then we could remove this verification since we already compare both indices above.

merkletree/pad_test.go
c633: Yup, I suggest to make some small changes, though.

merkletree/pad_test.go
Liamsi: Right. Thanks for your feedback!

merkletree/pad_test.go
Liamsi: Thanks for your feedback :-) https://github.com/coniks-sys/coniks-go/pull/54/commits/95650ab818b2b1e74de76175c305b8522e5aaf74

c633:
LGTM. Nice work @Liamsi :). Thanks!

merkletree/pad_test.go
```diff
@@ -241,14 +241,14 @@ func TestPoliciesChange(t *testing.T) {
 }
 
 func TestTB(t *testing.T) {
```
masomel: Should ``TestTB()`` be moved to ``tb_test.go``?

Liamsi: Yes, why not. It makes sense (https://github.com/coniks-sys/coniks-go/pull/54/commits/04cf99ad1e002027dfe6feb92dba7ad444543dbb)

merkletree/pad_test.go
Liamsi: Done in https://github.com/coniks-sys/coniks-go/pull/54/commits/95650ab818b2b1e74de76175c305b8522e5aaf74

masomel:
LGTM. Thanks @Liamsi!

PR #59, opened 2016-08-10
Open commitment for verification

### LINKED ISSUE #57: Server isn't opening the commitment
We need to make this change,

```
type ProofNode interface {
	Level() int
	Index() []byte
	Value() []byte
	IsEmpty() bool
-	Commitment() []byte
+	Salt() []byte
}
```

then the client computes the digest with `Salt()`, `Value()`, and the `name` it looked up.

I think ... need to go back to the paper.

c633:
* Resolve #57
* ~~Change the way we compute the commitment: from `commit(key||value)` to `commit(index||value)`~~

### DIFF
--- a/client/cgo/cgotest.go
+++ b/client/cgo/cgotest.go
@@ -1,6 +1,8 @@
 package main
 
 /*
+#include <stdlib.h>
+
 int testVerifyVrf(unsigned char *pk, int pkSize,
     unsigned char *m, int mSize,
     unsigned char *index, int indexSize,
@@ -38,6 +40,17 @@ int testVerifyAuthPath(
         leafIndex, leafIndexSize,
         leafCommitment, leafCommitmentSize,
         isLeafEmpty);
+}
+
+int testVerifyCommitment(
+	unsigned char *salt, int saltSize,
+	char *key, int keySize,
+	unsigned char *value, int valueSize,
+	unsigned char *commitment, int commitmentSize)  {
+	return cgoVerifyCommitment(salt, saltSize,
+		key, keySize,
+		value, valueSize,
+		commitment, commitmentSize);
 }
 
 #cgo CFLAGS: -Wno-implicit-function-declaration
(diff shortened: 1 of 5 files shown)


arlolra:
See the follow up responses in #57

masomel:
LGTM.

client/cgo/cgotest.go
```diff
@@ -1,6 +1,8 @@
 package main
 
 /*
+#include <stdlib.h>
```
arlolra: What necessitated this?

arlolra: Nevermind, the C.free

client/cgo/cgotest.go
```diff
@@ -166,6 +179,17 @@ func testVerifyAuthPath(t *testing.T) {
 	if proof.Leaf.IsEmpty() {
 		isLeafEmpty = 1
 	}
+	// verify commitment
+	key := C.CString(key3)
+	defer C.free(unsafe.Pointer(key))
```
arlolra: Curious, why you're doing this, but above you do `byteSliceToCpchar(message)`?

client/cgo/cgotest.go
```diff
@@ -166,6 +179,17 @@ func testVerifyAuthPath(t *testing.T) {
...
+
+	if v := C.testVerifyCommitment(byteSliceToCpchar(proof.Leaf.Salt()), C.int(len(proof.Leaf.Salt())),
+		key, C.int(len(key3)),
+		byteSliceToCpchar(proof.Leaf.Value()), C.int(len(proof.Leaf.Value())),
+		byteSliceToCpchar(proof.Leaf.Commitment()), C.int(len(proof.Leaf.Commitment()))); v != 1 {
+		t.Fatal("Veify commitment failed")
```
arlolra: typo

merkletree/proof.go
```diff
@@ -88,13 +89,22 @@ func VerifyAuthPath(ap *AuthenticationPath,
 	return true
 }
 
+func VerifyCommitment(salt []byte, key string, value []byte, commitment []byte) bool {
+	got := crypto.Digest(salt, []byte(key), value)
```
arlolra: I remember early on I told you to remove the `crypto.Commitment` function because it was redundant with `Digest`.  I sort of regret that now :(

c633: Since it was fine in that moment. I think we should add `crypto.Commitment` back, for now.

arlolra: > I think we should add crypto.Commitment back, for now.

Please do :)

c633: I'm thinking about this:

```
// package crypto
func Commitment(key string, value []byte) (salt []byte, commit []byte, err error) {
	salt, err = MakeRand()
	if err != nil {
		return nil, nil, err
	}
	return salt, Digest(salt, []byte(key), value), nil
}
```
and make `VerifyCommitment` be a function of `crypto` package.
What do you think?

arlolra: > What do you think?

That's probably fine.

I'm wondering if `salt` is particular to our implementation.  And, instead, we should model it more like,

```
type Commit struct {
  salt []byte
  value []byte
}

func NewCommit(stuff) (*Commit, error) {
  salt, err := MakeRand()
  if err != nil {
    return nil, err
  }
  return Digest(salt, stuff), nil
}

func (c *Commmit) Verify(stuff []byte) bool {
  return bytes.Equal(c.value, Digest(c.salt, stuff)
}
```

Maybe that's overkill?  On the client you'd skip the constructor,

```
&Commitment{salt, value}
```

I don't know.  I think what you suggested is fine ...

arlolra: Btw, let's do this in a separate pull.  What you have here is fine for #57.  Just fix the typo.

arlolra:
Merged in e3034c5fb67fa674d234e3ab943dd45b4385fc3b

c633:
> Merged in e3034c5

I sorry I didn't response last night. I would like to make a small change in 662b63dcb5f06063b9b0cbf8091d7078a90d5cf0. Let me know if I should squash it with e3034c5fb67fa674d234e3ab943dd45b4385fc3b

arlolra:
Looks good.  Just push it, don't squash.

PR #62, opened 2016-08-11
Refactoring: use unsigned int32 instead of int where it makes sense

### LINKED ISSUE #53: Use unsigned integers when dealing with non-negative values

Liamsi:
Small refactoring. Should resolve #53.
I chose `uint32` as the type for leaf nodes' level as its max value should be more then large enough and its size in memory is the same as `int`'s.

### DIFF
--- a/client/cgo/verification.go
+++ b/client/cgo/verification.go
@@ -113,7 +113,7 @@ func cgoVerifyAuthPath(treeHash unsafe.Pointer, treeHashSize C.int,
 	ap.LookupIndex = li
 	ap.PrunedTree = pt
 
-	if merkletree.VerifyAuthPath(ap, leafi, leafc, int(leafLevel), int(isLeafEmpty) == 1, th) {
+	if merkletree.VerifyAuthPath(ap, leafi, leafc, uint32(leafLevel), int(isLeafEmpty) == 1, th) {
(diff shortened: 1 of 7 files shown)


masomel:
LGTM!

c633:
LGTM!

arlolra:
> its size in memory is the same as int's.

Isn't `int`'s size implementation-specific?

Liamsi:
> Isn't int's size implementation-specific?

Sorry, yes of course it is. I wanted to say that `uint32`'s size is smaller (on most platforms) or at most equal to `int`'s. (`int`'s size is 32 or 64 bits `uint32`'s is always 32 bits)

arlolra:
Ah, ok, I was just making sure I hadn't missed something. Thanks.

arlolra:
Merged in e0ec01087e5cff342df57000a6456b2db7fac197

Issue #50, opened 2016-08-02
Allow developers to use their own crypto schemes


c633:
As @arlolra  commented in https://github.com/coniks-sys/coniks-go/pull/48#issuecomment-236802477:
> it might be worth abstracting this further into interfaces which the vrf/ and sign/ packages implement, so that using the PAD with other schemes is straightforward.

It also useful to allow the developers to supply theirs own hash scheme #6.

masomel:
I think this is also part of #6.

c633:
Close since it is a part of #6

masomel:
So I think there are two parts to this issue, right? One is specifying the scheme in the policy, and the other is actually supplying the scheme to the data structure. I think this issue should remain open to address this second part specifically.

PR #63, opened 2016-08-12
Remove unnecessary Policies's method & add new util functions

c633:
This pull is a part of #29. It includes following changes:
* Remove unnecessary method from `merkletree/policy.go`.
* Add new functions to `util` package.
* `sign.GenerateKey` should take a reader as `vrf`.

### DIFF
--- a/client/cgo/cgotest.go
+++ b/client/cgo/cgotest.go
@@ -90,7 +90,7 @@ func testVerifyVrf(t *testing.T) {
 }
 
 func testVerifySignature(t *testing.T) {
-	key, err := sign.GenerateKey()
+	key, err := sign.GenerateKey(nil)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -111,7 +111,7 @@ func testVerifySignature(t *testing.T) {
 }
 
 func testVerifyHashChain(t *testing.T) {
-	signKey, err := sign.GenerateKey()
+	signKey, err := sign.GenerateKey(nil)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -140,7 +140,7 @@ func testVerifyHashChain(t *testing.T) {
 }
 
 func testVerifyAuthPath(t *testing.T) {
-	signKey, err := sign.GenerateKey()
+	signKey, err := sign.GenerateKey(nil)
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -224,7 +224,7 @@ func testVerifyAuthPath(t *testing.T) {
 }
 
 func testVerifyProofOfAbsenceSamePrefix(t *testing.T) {
-	signKey, err := sign.GenerateKey()
+	signKey, err := sign.GenerateKey(nil)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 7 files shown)


c633:
Btw, I'm a little bit uncomfortable when looking at the `sign.go` and `vrf.go` right now. They both use `ed25519` but they have difference `GenerateKey`'s signatures, and difference key type (`sign`'s key type is a slice and `vrf`'s key type is an array). Just a few questions came to my mind:

* Do we need to return the public part in `GenerateKey`?
* What should be returned by `key.Public()`? (a `PublicKey` or a slice of byte?)
* What should be the underlying type of the key: a slice or an array? I'd prefer a slice. 

I suggest we should refactor the `vrf` package to export the same API's signature & key type as `sign` package. What do you think? 
@arlolra  @Liamsi  @masomel

Liamsi:
> Do we need to return the public part in GenerateKey?

If we don't, we assume the key to be of a specific format (like the one created by ed25519). Not returning the public-key here could make #50 more difficult to implement.

> What should be returned by key.Public()? (a PublicKey or a slice of byte?)
> What should be the underlying type of the key: a slice or an array? I'd prefer a slice.

I agree that the underlying type should be a slice instead of a fixed sized array (which again assumes we use a specific key format/size). 
I think we'll need to make this transition anyways if we want to define an interface for signing/VRF/hashing as proposed in #50 (or some related discussion I can't find right now).

c633:
> If we don't, we assume the key to be of a specific format (like the one created by ed25519). Not returning the public-key here could make #50 more difficult to implement.

If we do, `key.Public()` seems to be redundant when I looked at it in the first time. Look at the code, the only use of `key.Public()` right now is for serialization function. Should `key.Public()` return a slice and be renamed to something like `key.PublicPart()`?

Liamsi:
> Look at the code, the only use of key.Public() right now is for serialization function. Should key.Public() return a slice and be renamed to something like key.PublicPart()?

I see your point. Then why not call it `Serialize()` like in other place we want a slice of byte from an object (used for serialization)? `PublicPart` sound like the public key is always part of the (encoding of the) private key. Are you referring to both `Public` (in vrf and sign) functions btw?

Liamsi:
> They both use ed25519 but they have difference GenerateKey's signatures, and difference key type (sign's key type is a slice and vrf's key type is an array)

As a side-note: It would be fairly easy to modify the VRF lib to use sha512 instead of sha3/shake. Then we could use the exact same key type as in sign.go (`type PrivateKey ed25519.PrivateKey`) and `GenerateKey` could also look the same (basically calls `ed25519.GenerateKey`).

arlolra:
> I suggest we should refactor the vrf package to export the same API's signature & key type as sign package. What do you think?

I think this is a fine idea.  Can you open a separate ticket for it?  I'd like to merge this code and I don't want the discussion to get lost.

> If we don't, we assume the key to be of a specific format (like the one created by ed25519). Not returning the public-key here could make #50 more difficult to implement.

Not sure I agree.  If, for whatever reason, the public key cannot be generated from the private key at call time, (`private.Public()`), we can always augment the `PrivateKey` structure to store it at construction.

arlolra:
> Can you open a separate ticket for it?

Sorry, I opened #65

c633:
> Sorry, I opened #65

Thanks!

Issue #15, opened 2016-07-13
Document how data is serialized


arlolra:
Particularly things like `LongToBytes()` should be well specified.

PR #64, opened 2016-08-12
Implement protocol:

c633:
This is a part of #29. This pull includes the implementation of the CONIKS protocol.

* Implement register API

* ~~Implement key lookup API~~

* ~~Implement monitoring API~~

### DIFF
--- a/protocol/directory.go
+++ b/protocol/directory.go
@@ -1,0 +1,95 @@
+package protocol
+
+import (
+	"bytes"
+
+	"github.com/coniks-sys/coniks-go/crypto/sign"
+	"github.com/coniks-sys/coniks-go/crypto/vrf"
+	"github.com/coniks-sys/coniks-go/merkletree"
+)
+
+type ConiksDirectory struct {
+	pad      *merkletree.PAD
+	useTBs   bool
+	tbs      map[string]*merkletree.TemporaryBinding
+	policies merkletree.Policies
+}
+
+func InitDirectory(epDeadline merkletree.TimeStamp, vrfKey vrf.PrivateKey,
+	signKey sign.PrivateKey, dirSize uint64, useTBs bool) *ConiksDirectory {
+	d := new(ConiksDirectory)
+	d.SetPolicies(epDeadline, vrfKey)
+	pad, err := merkletree.NewPAD(d.policies, signKey, dirSize)
+	if err != nil {
+		panic(err)
+	}
+	d.pad = pad
+	d.useTBs = useTBs
+	if useTBs {
+		d.tbs = make(map[string]*merkletree.TemporaryBinding)
+	}
+	return d
+}
+
+func (d *ConiksDirectory) Update() {
+	d.pad.Update(d.policies)
+	// clear issued temporary bindings
+	for key := range d.tbs {
+		delete(d.tbs, key)
+	}
+}
+
+func (d *ConiksDirectory) SetPolicies(epDeadline merkletree.TimeStamp, vrfKey vrf.PrivateKey) {
+	d.policies = merkletree.NewPolicies(epDeadline, vrfKey)
+}
+
+func (d *ConiksDirectory) LatestSTR() *merkletree.SignedTreeRoot {
+	return d.pad.LatestSTR()
+}
+
+// HandleOps validates the request message and then pass it to
+// appropriate operation handler according to the request type.
+func (d *ConiksDirectory) HandleOps(req *Request) (Response, ErrorCode) {
+	switch req.Type {
+	case RegistrationType:
+		if msg, ok := req.Request.(*RegistrationRequest); ok {
+			if len(msg.Username) > 0 && len(msg.Key) > 0 {
+				return d.Register(msg)
+			}
+		}
+	}
+	return nil, ErrorMalformedClientMessage
+}
+
+func (d *ConiksDirectory) Register(req *RegistrationRequest) (
+	*DirectoryProof, ErrorCode) {
+	// check whether the name already exists
+	// in the directory before we register
+	ap, err := d.pad.Lookup(req.Username)
+	if err != nil {
+		return nil, ErrorDirectory
+	}
+	if bytes.Equal(ap.LookupIndex, ap.Leaf.Index) {
+		return NewRegistrationProof(ap, d.LatestSTR(), nil), ErrorNameExisted
+	}
+	if d.useTBs {
+		// also check the temporary bindings array
+		// currently the server allows only one registration/key change per epoch
+		if tb := d.tbs[req.Username]; tb != nil {
+			return NewRegistrationProof(ap, d.LatestSTR(), tb), ErrorNameExisted
+		}
+
+		// insert new data to the directory on-the-fly
+		tb, err := d.pad.TB(req.Username, req.Key)
+		if err != nil {
+			return nil, ErrorDirectory
+		}
+		d.tbs[req.Username] = tb
+		return NewRegistrationProof(ap, d.LatestSTR(), tb), Success
+	} else {
+		if err = d.pad.Set(req.Username, req.Key); err != nil {
+			return nil, ErrorDirectory
+		}
+		return NewRegistrationProof(ap, d.LatestSTR(), nil), Success
+	}
+}
--- a/protocol/directory_test.go
+++ b/protocol/directory_test.go
@@ -1,0 +1,140 @@
+package protocol
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/coniks-sys/coniks-go/crypto/sign"
+	"github.com/coniks-sys/coniks-go/crypto/vrf"
+)
+
+func newDirectory(t *testing.T, useTBs bool) *ConiksDirectory {
+	vrfKey, err := vrf.GenerateKey(nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	signKey, err := sign.GenerateKey(nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// epDeadline merkletree.TimeStamp, vrfKey vrf.PrivateKey,
+	// signKey sign.PrivateKey, dirSize uint64, useTBs bool uint64
+	d := InitDirectory(1, vrfKey, signKey, 10, useTBs)
+	return d
+}
+
+func TestRegisterWithTB(t *testing.T) {
+	// expect return a proof of absence
+	// along with a TB of registering user
+	d := newDirectory(t, true)
+
+	res, err := d.Register(&RegistrationRequest{
+		Username: "alice",
+		Key:      []byte("key")})
+	if res.Type != RegistrationType {
+		t.Fatal("Expect response type", RegistrationType, "got", res.Type)
+	}
+	if err != Success {
+		t.Fatal("Unable to register")
+	}
+	if ap := res.AP; ap == nil || bytes.Equal(ap.LookupIndex, ap.Leaf.Index) {
+		t.Fatal("Expect a proof of absence")
+	}
+	if res.TB == nil {
+		t.Fatal("Expect returned TB is not nil")
+	}
+	if d.tbs["alice"] == nil {
+		t.Fatal("Expect TBs array has registering user")
+	}
+	d.Update()
+	if len(d.tbs) != 0 {
+		t.Fatal("Expect TBs array is empty")
+	}
+}
+
+func TestRegisterExistedUserWithTB(t *testing.T) {
+	d := newDirectory(t, true)
+	_, err := d.Register(&RegistrationRequest{
+		Username: "alice",
+		Key:      []byte("key")})
+	if err != Success {
+		t.Fatal("Unable to register")
+	}
+	// register in the same epoch
+	// expect return a proof of absence
+	// along with a TB of registering user
+	// and error ErrorNameExisted
+	res, err := d.Register(&RegistrationRequest{
+		Username: "alice",
+		Key:      []byte("key")})
+	if err != ErrorNameExisted {
+		t.Fatal("Expect error code", ErrorNameExisted, "got", err)
+	}
+	if ap := res.AP; ap == nil || bytes.Equal(ap.LookupIndex, ap.Leaf.Index) {
+		t.Fatal("Expect a proof of absence")
+	}
+	if res.TB == nil {
+		t.Fatal("Expect returned TB is not nil")
+	}
+
+	d.Update()
+	// register in different epochs
+	// expect return a proof of inclusion
+	// and error ErrorNameExisted
+	res, err = d.Register(&RegistrationRequest{
+		Username: "alice",
+		Key:      []byte("key")})
+	if err != ErrorNameExisted {
+		t.Fatal("Expect error code", ErrorNameExisted, "got", err)
+	}
+	if ap := res.AP; ap == nil || !bytes.Equal(ap.LookupIndex, ap.Leaf.Index) {
+		t.Fatal("Expect a proof of inclusion")
+	}
+	if res.TB != nil {
+		t.Fatal("Expect returned TB is nil")
+	}
+}
+
+func TestRegisterWithoutTB(t *testing.T) {
+	// expect return a proof of absence
+	d := newDirectory(t, false)
+	res, err := d.Register(&RegistrationRequest{
+		Username: "alice",
+		Key:      []byte("key")})
+	if err != Success {
+		t.Fatal("Unable to register")
+	}
+	if ap := res.AP; ap == nil || bytes.Equal(ap.LookupIndex, ap.Leaf.Index) {
+		t.Fatal("Expect a proof of absence")
+	}
+	if res.TB != nil {
+		t.Fatal("Expect returned TB is nil")
+	}
+}
+
+func TestRegisterExistedUserWithoutTB(t *testing.T) {
+	d := newDirectory(t, false)
+	_, err := d.Register(&RegistrationRequest{
+		Username: "alice",
+		Key:      []byte("key")})
+	if err != Success {
+		t.Fatal("Unable to register")
+	}
+
+	d.Update()
+	// expect return a proof of inclusion
+	// and error ErrorNameExisted
+	res, err := d.Register(&RegistrationRequest{
+		Username: "alice",
+		Key:      []byte("key")})
+	if err != ErrorNameExisted {
+		t.Fatal("Expect error code", ErrorNameExisted, "got", err)
+	}
+	if ap := res.AP; ap == nil || !bytes.Equal(ap.LookupIndex, ap.Leaf.Index) {
+		t.Fatal("Expect a proof of inclusion")
+	}
+	if res.TB != nil {
+		t.Fatal("Expect returned TB is nil")
+	}
+}
--- a/protocol/error.go
+++ b/protocol/error.go
@@ -1,0 +1,32 @@
+// Defines constants representing the types
+// of errors that the server may return to a client.
+
+package protocol
+
+import "errors"
+
+type ErrorCode int
+
+const (
+	Success ErrorCode = iota + 10
+	ErrorDirectory
+	ErrorNameExisted
+	ErrorNameNotFound
+	ErrorMalformedClientMessage
+)
+
+var (
+	errorMessages = map[ErrorCode]error{
+		ErrorMalformedClientMessage: errors.New("[coniks] Malformed client request"),
+		ErrorNameExisted:            errors.New("[coniks] Registering identity is already registered"),
+		ErrorNameNotFound:           errors.New("[coniks] Name not found"),
+		ErrorDirectory:              errors.New("[coniks] Directory error"),
+	}
+)
+
+func (e ErrorCode) Error() error {
+	if errorMessages[e] == nil {
+		return errorMessages[ErrorDirectory]
+	}
+	return errorMessages[e]
+}
--- a/protocol/message.go
+++ b/protocol/message.go
@@ -1,0 +1,47 @@
+package protocol
+
+import (
+	m "github.com/coniks-sys/coniks-go/merkletree"
+)
+
+// Defines constants representing the types
+// of messages exchanged by clients and servers.
+const (
+	RegistrationType = iota
+)
+
+// Request messages
+type Request struct {
+	Type    int
+	Request interface{}
+}
+
+type RegistrationRequest struct {
+	Username               string `json:"username"`
+	Key                    []byte `json:"key"`
+	AllowUnsignedKeychange bool   `json:"allow_unsigned_key_change,omitempty"`
+	AllowPublicLookup      bool   `json:"allow_public_lookup,omitempty"`
+}
+
+// Response messages
+type Response interface{}
+
+type ErrorResponse struct {
+	Error ErrorCode
+}
+
+func NewErrorResponse(e ErrorCode) Response {
+	return &ErrorResponse{Error: e}
+}
+
+type DirectoryProof struct {
+	Type int
+	AP   *m.AuthenticationPath
+	STR  *m.SignedTreeRoot
+	TB   *m.TemporaryBinding `json:",omitempty"`
+}
+
+func NewRegistrationProof(ap *m.AuthenticationPath, str *m.SignedTreeRoot,
+	tb *m.TemporaryBinding) *DirectoryProof {
+	return &DirectoryProof{RegistrationType, ap, str, tb}
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,8 @@ script:
   - go test -coverprofile=vrf.coverprofile ./crypto/vrf
   - go test -coverprofile=merkletree.coverprofile ./merkletree
   - go test -coverprofile=utils.coverprofile ./utils
-
+  - go test -coverprofile=protocol.coverprofile ./protocol
+  
 after_success:
   - gover
   - goveralls -coverprofile=gover.coverprofile -service travis-ci
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ The packages in this library implement the various components of the CONIKS syst
 - ``crypto``: Cryptographic algorithms and operations
 - ``merkletree``: Merkle prefix tree and related data structures
 - ``utils``: Utility functions
+- ``protocol``: CONIKS protocols implementation/library
 - ``storage``: DB hooks for storage backend (currently the library supports key-value db only)
 
 ## Disclaimer


c633:
Oops, is there any problem with this pull? @arlolra

arlolra:
Sorry, I had deleted the branch w/o thinking.

protocol/directory.go
```diff
@@ -0,0 +1,109 @@
...
+			return ap, tb, ErrorNameExisted
+		}
+
+		// insert new data to the directory on-the-fly
+		tb, err := d.pad.TB(uname, key)
+		if err != nil {
+			return nil, nil, ErrorInternalServer
+		}
+		d.tbs[uname] = tb
+		return ap, tb, Success
```
arlolra: See https://github.com/coniks-sys/coniks-go/pull/29#discussion_r74539375

protocol/error.go
```diff
@@ -0,0 +1,32 @@
...
+	ErrorNameNotFound
+	ErrorMalformedClientMessage
+)
+
+var (
+	errorMessages = map[ErrorCode]error{
+		ErrorMalformedClientMessage: errors.New("[coniks] Malformed client request"),
+		ErrorNameExisted:            errors.New("[coniks] Registering identity is already registered"),
+		ErrorNameNotFound:           errors.New("[coniks] Name not found"),
+		ErrorInternalServer:         errors.New("[coniks] Internal server error"),
```
arlolra: I'm wondering is Directory would be a better name, since we don't have a server at this point.

protocol/encoding.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	return json.Marshal(tb)
+}
+
+func MarshalSTR(str *merkletree.SignedTreeRoot) ([]byte, error) {
+	policies, err := json.Marshal(str.Policies)
+	if err != nil {
+		return nil, err
+	}
+	return json.Marshal(&struct {
```
arlolra: See https://github.com/coniks-sys/coniks-go/pull/29/files/a34c7568f9b402bb8784f6ce0ea23046b37b74c9#r74498386

arlolra: I'm going to backtrack a little bit.  Maybe I don't know what I'm talking about.

It feels like the encoding of the messages should be part of the server, since that's where the decision about JSON/Protobuf/etc. will happen.  But the message struct should be part of the protocol, and they should define the interface with the directory.

Help me think this through.

protocol/directory.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (d *ConiksDirectory) SetPolicies(epDeadline merkletree.TimeStamp, vrfKey *vrf.PrivateKey) {
+	d.policies = merkletree.NewPolicies(epDeadline, vrfKey)
+}
+
+func (d *ConiksDirectory) LatestSTR() *merkletree.SignedTreeRoot {
+	return d.pad.LatestSTR()
+}
+
+func (d *ConiksDirectory) Register(uname string, key []byte) (
```
arlolra: Shouldn't this take a `RegistrationRequest`?  Isn't that why we're defining that?

```
func (d *ConiksDirectory) Register(req RegistrationRequest) (
```

protocol/directory.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	if err = d.pad.Set(uname, key); err != nil {
+		return nil, nil, ErrorInternalServer
+	}
+
+	return ap, nil, Success
+}
+
+func (d *ConiksDirectory) KeyLookup(uname string) (
```
arlolra:
```
func (d *ConiksDirectory) KeyLookup(req *KeyLookupRequest) (
```

protocol/directory.go
```diff
@@ -0,0 +1,109 @@
...
+		// if not found in the tree, do lookup in tb array
+		if d.useTBs {
+			if tb := d.tbs[uname]; tb != nil {
+				return ap, tb, Success
+			}
+		}
+		return ap, nil, ErrorNameNotFound
+	}
+
+	return ap, nil, Success
```
arlolra: Similarlry, shouldn't it return a `KeyLookupResponse`?

protocol/directory.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		d.tbs[uname] = tb
+		return ap, tb, Success
+	}
+
+	if err = d.pad.Set(uname, key); err != nil {
+		return nil, nil, ErrorInternalServer
+	}
+
+	return ap, nil, Success
```
arlolra: return a `RegistrationResponse`?

protocol/directory.go
c633: Yes, you're right. Don't understand why I have had such a stupid code :(

protocol/error.go
masomel: Yeah, I agree. Also because I think it's best to keep the server and protocol logic separate.

protocol/encoding.go
masomel: > It feels like the encoding of the messages should be part of the server, since that's where the decision about JSON/Protobuf/etc. will happen. But the message struct should be part of the protocol, and they should define the interface with the directory.

I agree with this. I like the idea of defining the interface with the directory using messages. I think the request messages representing the client-to-server messages already do a fair job. What about the server-to-client messages? Right now the message format and the encoding for responses are intermingled. How can we clearly define a general message format for a registration response, or a lookup?

protocol/directory.go
masomel: Hmm, this is an interesting idea. I still think the more natural way to think about a registration and lookup would be to pass in the individual values, instead of the request message, though. I think the directory should remain a higher-level abstraction to the PAD, and it should still be up to the server to handle the message parsing and pass in the parsed values.

protocol/directory.go
masomel: Same as https://github.com/coniks-sys/coniks-go/pull/64#discussion_r74650894

protocol/directory.go
masomel: Same as https://github.com/coniks-sys/coniks-go/pull/64#discussion_r74650894

protocol/directory.go
c633: Since it returns `*RegistrationResponse`, does it make sense to pass in `*RegistrationRequest`?

masomel:
After a quick glance at 873af4b, I'm still not sure how I feel about using the messages as arguments and return values of the directory functions. I think this change makes the directory data structure less general. I still like @arlolra's idea of using the messages as the interface to the directory, but I was imagining messages to be more of a layer of abstraction above the directory, which the server then combines internally by parsing the messages into the correct inputs for the directory.

protocol/encoding.go
c633: > I like the idea of defining the interface with the directory using messages

Could you please explain this for me? Thanks!

masomel: Sure! I'm thinking of the messages as a layer of abstraction above the directory, so that a client only needs to "think" in terms of operations they want the server to perform. For example, in the case of a registration, all the client needs to do is send the username, key (and user policies?) to the server, which then handles these inputs and passes them on to the directory. I hope this makes sense. 

Like I said above, I think the request messages already do a god job of representing the operations. But I think the response messages from the server to the client don't define a clear interface yet. I think part of that is that they contain a lot of pieces of data that the client has to parse. Maybe it might be helpful to think of the responses more in terms of client operations as well?

masomel:
> Yup, I'm going to ask you about this. @masomel and @Liamsi did agree that we should return the AP/TB along with the error. But I'm thinking about the privacy here. It seems like we have to come back to this discussion. 

My main argument for returning the proofs (AP/TB) along with the error is for transparency. I don't like the idea of allowing the server to return messages without proofs sometimes, I think these could be used to equivocate again and there would be no proof. I think one solution could be to scrub the proof of inclusion of the sensitive data before sending it back. And only sending back the full proof of inclusion when the request is a lookup. What do you think?

protocol/encoding.go
c633: > Sure! I'm thinking of the messages as a layer of abstraction above the directory, so that a client only needs to "think" in terms of operations they want the server to perform. For example, in the case of a registration, all the client needs to do is send the username, key (and user policies?) to the server, which then handles these inputs and passes them on to the directory. I hope this makes sense.
Like I said above, I think the request messages already do a god job of representing the operations. 

Thanks! I got it. 

> But I think the response messages from the server to the client don't define a clear interface yet.

I'm not so sure about it. Shouldn't moving `encoding.go` to the server make sense? Because `encoding.go` contains mostly protocol's messages marshaling (Though the original idea was that we would add more encoding/decoding functions into this file when we support (say) Protobuf/etc.)

arlolra:
> I was imagining messages to be more of a layer of abstraction above the directory, which the server then combines internally by parsing the messages into the correct inputs for the directory.

Then that sounds like the messages shouldn't be part of the protocol.

> I think this change makes the directory data structure less general.

I'm not so sure.  All it does it force you to use a structure, rather than spread out the data over the functions' signature.  It's more verbose, yes, but no less general.

However, what we gain is a specification (by implementation) of messages exchanged, (sort of like what I was asking in #15), which leaves the marshalling (JSON, Protobuf, etc.) to the transport layer.

I've been thinking about the protocol package as the spec.

c633:
> I think one solution could be to scrub the proof of inclusion of the sensitive data before sending it back. And only sending back the full proof of inclusion when the request is a lookup.

I see it just takes an extra query. Is there any other difference? 

> But I'm thinking about the privacy here

Maybe I was overthinking about it.

> What do you think?

Yes, I agree that the server should always returns messages with proofs.

masomel:
> Then that sounds like the messages shouldn't be part of the protocol.

I don't think so. I agree, the protocol package should be like a spec, so it should define protocol messages abstractly.

> I'm not so sure. All it does it force you to use a structure, rather than spread out the data over the functions' signature. It's more verbose, yes, but no less general.

I mean, sure, the function now returns a single value as opposed to 3 or 4. But now what's definitely more verbose is the `Register` function because it now also has to initialize the response struct. I just think we're mixing message handling with core operations by having the directory accept/return messages, instead of the "raw" data.

> what we gain is a specification (by implementation) of messages exchanged, (sort of like what I was asking in #15), which leaves the marshalling (JSON, Protobuf, etc.) to the transport layer.

As I said above, I don't disagree that marshaling should be part of the transport layer (i.e. encoding should be moved to the server package). I think it's fine to think of messages as a spec of the data exchanged between the server and the client. But I think the directory performing the core CONIKS operations should be independent of how the client-server messages are defined.

protocol/encoding.go
masomel: >> But I think the response messages from the server to the client don't define a clear interface yet.
> 
> I'm not so sure about it. Shouldn't moving encoding.go to the server make sense? Because encoding.go contains mostly protocol's messages marshaling (Though the original idea was that we would add more encoding/decoding functions into this file when we support (say) Protobuf/etc.)

Maybe I misunderstood something, but I think whether the response messages define a clear interface and whether we move encoding to the server package are two separate issues. I think the issue that @arlolra was bringing up originally was that `encoding.go` also contains a lot of functions to marshal core data structures, which aren't protocol messages themselves, and it shouldn't be part of the protocol how the data structures are encoded to be sent to the client over the network. In other words, I think the idea would be for `encoding.go` to only contain encoding/decoding functions for message structs, and then it would also be fine to add functions for other encodings, like you suggested.

The clearer interface I was referring to is defining response messages more in terms of client-side operations (e.g. monitoring). But I don't know how much sense this makes...

arlolra:
> I just think we're mixing message handling with core operations by having the directory accept/return messages, instead of the "raw" data.

I don't know, `Directory` is an abstraction of over the "raw" merkletree, and if I was looking at the protocol from a high level, I would say, a registration request is received by the server and calls the directory register function with it.

I guess what I don't see is how to use the defined messages if they aren't part of the interface to the directory.

But, I should say, I'm fine with backing off from this idea.  We seems to be in agreement to move marshalling to the server and, otherwise, the protocol is looking pretty good.  So, if this is feeling like a digression, don't worry about it.

masomel:
> I don't know, Directory is an abstraction of over the "raw" merkletree, and if I was looking at the protocol from a high level, I would say, a registration request is received by the server and calls the directory register function with it.
>
> I guess what I don't see is how to use the defined messages if they aren't part of the interface to the directory.

I totally agree that the `Directory` abstracts away the Merkle tree. I can see what you mean now, that high-level view of the protocol makes sense. I was thinking that the server would explicitly set the abstract messages, and then pass them through the encoder before sending them out on the wire. I also still think that moving that logic out of the `Directory` would make the code there a bit cleaner. But at the same time, it's quite possible I'm overthinking all this and trying to make the code more modular than is necessary.

> I'm fine with backing off from this idea.

Your high-level view of the protocol actually makes sense to me, so I'm fine with going with this idea!

masomel:
>> I think one solution could be to scrub the proof of inclusion of the sensitive data before sending it back. And only sending back the full proof of inclusion when the request is a lookup.
>
> I see it just takes an extra query. Is there any other difference? 

I don't think scrubbing the proof of inclusion would require an extra query; couldn't you just set `Proof.Leaf.Value() = nil` in `Directory.Register()` in the `ErrorNameExists` case before returning the AP?

protocol/directory.go
```diff
@@ -54,56 +54,86 @@ func (d *ConiksDirectory) LatestSTR() *merkletree.SignedTreeRoot {
...
 	// in the directory before we register
-	ap, err := d.pad.Lookup(uname)
+	ap, err := d.pad.Lookup(req.Username)
 	if err != nil {
-		return nil, nil, ErrorInternalServer
+		return nil, ErrorDirectory
 	}
 	if bytes.Equal(ap.LookupIndex, ap.Leaf.Index()) {
-		return ap, nil, ErrorNameExisted
+		return &RegistrationResponse{
```
masomel: Could this be replaced by a setter function? This will clean up the code a bit.

protocol/directory.go
```diff
@@ -54,56 +54,86 @@ func (d *ConiksDirectory) LatestSTR() *merkletree.SignedTreeRoot {
...
+		return nil, ErrorDirectory
 	}
 
 	if !bytes.Equal(ap.LookupIndex, ap.Leaf.Index()) {
 		// if not found in the tree, do lookup in tb array
 		if d.useTBs {
-			if tb := d.tbs[uname]; tb != nil {
-				return ap, tb, Success
+			if tb := d.tbs[req.Username]; tb != nil {
+				return &KeyLookupResponse{
```
masomel: Similarly here, I think this should be replaced with a setter to make the code a bit more concise.

protocol/message.go
```diff
@@ -0,0 +1,50 @@
+package protocol
+
+import "github.com/coniks-sys/coniks-go/merkletree"
+
+// Defines constants representing the types
+// of messages exchanged by clients and servers.
+const (
+	RegistrationType = 0
+	CommitmentType   = 1
```
masomel: What is `CommitmentType` used for?

protocol/message.go
```diff
@@ -0,0 +1,50 @@
...
+	STR  *merkletree.SignedTreeRoot
+	AP   *merkletree.AuthenticationPath
+	TB   *merkletree.TemporaryBinding
+}
+
+type KeyLookupRequest struct {
+	Username string `json:"username"`
+}
+
+type KeyLookupResponse struct {
```
masomel: For `RegistrationResponse` and `KeyLookupResponse`, I'm wondering if we should call these something like `RegistrationProof` and `LookupProof` instead, to be a bit more accurate.

protocol/message.go
c633: It's just a placeholder for monitoring function. I added it to make this lib align with the java implementation. It looks a little bit awkward if there is no type with value 1.

protocol/message.go
c633: Yup, totally makes sense!

protocol/directory.go
c633: I'm not so sure. Eventually, this's just a constructor. A setter function doesn't change the code so much, I think.

c633:
> I don't think scrubbing the proof of inclusion would require an extra query; couldn't you just set Proof.Leaf.Value() = nil in Directory.Register() in the ErrorNameExists case before returning the AP?

You're right. But my questions is does the client just need to send a key lookup request to obtain the full proof of inclusion (if he wants)? So why do we need this extra step?

protocol/message.go
masomel: Ah ok. I actually need to change this in the Java implementation since commitments is an old term we used for STRs. But I don't think there's a separate STR message in this implementation, at least for now, right? So I think you could remove it and make `LookupType = 1`so you don't have the awkwardness.

masomel:
> But my questions is does the client just need to send a key lookup request to obtain the full proof of inclusion (if he wants)? So why do we need this extra step?

You're absolutely right. But I think if we can potentially prevent a privacy leak by making it (a tiny bit) more cumbersome for malicious clients to obtain the private information, we should :) The server should also rate-limit the requests it receives from the same client anyway to prevent enumeration attacks and DoS attacks. But this is a separate issue.

c633:
> But I think if we can potentially prevent a privacy leak by making it (a tiny bit) more cumbersome for malicious clients to obtain the private information, we should :)

Thank you for the explanation!

protocol/message.go
c633: > But I don't think there's a separate STR message in this implementation, at least for now, right?

How's about the monitoring function? Does the client still need to obtain some missed STRs to verify the hash chain anyways? Maybe I can rename it to `ConsistencyCheckType`?

c633:
> I totally agree that the Directory abstracts away the Merkle tree. 

If we want to completely abstract away the Merkletree, and also move the encoding to the server, I think we should have changes as in 425b5edcffc0d4de63c46e8ba0817b719e46e7f4.

masomel:
> If we want to completely abstract away the Merkletree, and also move the encoding to the server, I think we should have changes as in 425b5ed.

Those changes make a lot of sense to me, looks good!

protocol/message.go
masomel: > How's about the monitoring function? Does the client still need to obtain some missed STRs to verify the hash chain anyways? Maybe I can rename it to `ConsistencyCheckType`?

You're right, the client might have to "catch up" when monitoring. But it needs to do full lookups, not just verify the hash chain, to make sure that its public key/data blob remained consistent during the missed epochs. So I still think that monitoring would be covered by `KeyLookupType`.

protocol/directory.go
masomel: You're right. I also realized later that this should be a constructor. You don't think refactoring these calls into constructors would make the code less verbose?

protocol/message.go
```diff
@@ -46,5 +46,16 @@ type KeyLookupRequest struct {
 type KeyLookupProof struct {
 	Type int
 	AP   *merkletree.AuthenticationPath
-	TB   *merkletree.TemporaryBinding
+	TB   *merkletree.TemporaryBinding `json:",omitempty"`
```
masomel: `KeyLookupProof` should also contain the STR.

protocol/message.go
```diff
@@ -46,5 +46,16 @@ type KeyLookupRequest struct {
...
+}
+
+type ConsistencyCheckRequest struct {
```
masomel: This is just like a `KeyLookupRequest` except with an `Epoch` added. Couldn't we reuse `KeyLookupRequest` for the consistency checks and annotate `Epoch` with `omitempty`?

protocol/message.go
```diff
@@ -46,5 +46,16 @@ type KeyLookupRequest struct {
...
+	Username string `json:"username"`
+	Epoch    uint64 `json:"epoch"`
+}
+
+type ConsistencyCheckProof struct {
```
masomel: With the STR added to `KeyLookupProof`, this essentially becomes the same struct (except without the optional TB field). I think we could merge these structs, but keep the `ConsistencyCheckProof` name since I think it reflects better how clients operate on this message.

c633: I'm not so sure about it. I think `ConsistencyCheckProof` shouldn't include TB field, even if the server's using TB. I'd like to have 2 separate struct to make thing more clarify. Hopefully it makes sense to you.

protocol/message.go
c633: Yes, I agree!

masomel:
@c633 It seems the coveralls server is down right now (I tried several hours ago in coniks-java, and also an hour ago, and it's still failing), and it's causing Travis to fail, too :(

c633:
> @c633 It seems the coveralls server is down right now (I tried several hours ago in coniks-java, and also an hour ago, and it's still failing), and it's causing Travis to fail, too :(

Yes, I see.

Btw, I think this pull is already for another review. 
/cc @arlolra @masomel @Liamsi

protocol/message.go
Liamsi: Just another idea: You could refactor the common fields into yet another struct (not sure how to name it) and imbed it into both. In that way you won't repeat the same struct but still have a clear separation of concerns.

protocol/directory.go
```diff
@@ -0,0 +1,109 @@
...
+type ConiksDirectory struct {
+	pad      *merkletree.PAD
+	useTBs   bool
+	tbs      map[string]*merkletree.TemporaryBinding
+	policies merkletree.Policies
+}
+
+func InitDirectory(epDeadline merkletree.TimeStamp, vrfKey *vrf.PrivateKey,
+	signKey sign.PrivateKey, dirSize uint64,
+	useTBs bool, capacity uint64) *ConiksDirectory {
```
Liamsi: Nitpicking: If capacity is only used for TBs the name of the argument could reflect that (initialTBcapacity might be too long, but sth like this)

Liamsi: Or we could even drop the capacity completely (as it is "just" an optimization).

c633: I decided to drop it in 582e08ebe5fc9385bc8722a2fbed994d9b29db76

protocol/directory.go
```diff
@@ -0,0 +1,190 @@
...
+	d.policies = merkletree.NewPolicies(epDeadline, vrfKey)
+}
+
+func (d *ConiksDirectory) LatestSTR() *merkletree.SignedTreeRoot {
+	return d.pad.LatestSTR()
+}
+
+// HandleOps validates the request message and then pass it to
+// appropriate operation handler according to the request type.
+func (d *ConiksDirectory) HandleOps(req *Request) (Response, ErrorCode) {
```
arlolra: I like this, but wonder if it's better suited in the server/ ?

protocol/directory.go
```diff
@@ -0,0 +1,190 @@
...
+	if err != nil {
+		return nil, ErrorDirectory
+	}
+	if bytes.Equal(ap.LookupIndex, ap.Leaf.Index) {
+		return &RegistrationProof{
+			Type: RegistrationType,
+			AP:   ap,
+			STR:  d.LatestSTR(),
+			TB:   nil,
+		}, ErrorNameExisted
```
arlolra: Does this look better as?

```
  return NewRegistrationProof(ap, d.LatestSTR(), nil), ErrorNameExisted
```

protocol/directory.go
c633: The main reason for this is to prevent invalid requests from being passed into the handlers. Since the `server` package is just a reference implementation, so I think it'd be better if the library provided an API for validation.

protocol/directory.go
c633: Yup, @masomel suggested this, too. Thank you all!

protocol/directory.go
arlolra: Fair enough

protocol/message.go
```diff
@@ -0,0 +1,74 @@
...
+	Username               string `json:"username"`
+	Key                    []byte `json:"key"`
+	AllowUnsignedKeychange bool   `json:"allow_unsigned_key_change,omitempty"`
+	AllowPublicLookup      bool   `json:"allow_public_lookup,omitempty"`
+}
+
+type RegistrationProof struct {
+	Type int
+	AP   *m.AuthenticationPath
+	STR  *m.SignedTreeRoot
```
arlolra: `*ConsistencyCheckProof`

c633: Thanks!

protocol/directory.go
```diff
@@ -0,0 +1,147 @@
...
+		return NewKeyLookupProof(ap, d.LatestSTR(), nil), ErrorNameNotFound
+	}
+	return NewKeyLookupProof(ap, d.LatestSTR(), nil), Success
+}
+
+// ConsistencyCheck uses the same request struct
+// as KeyLookup but with a different `Request`'s `Type`.
+// The caller must check for Epoch field when unmarshalling the request
+// before passing it into this method.
+func (d *ConiksDirectory) ConsistencyCheck(req *KeyLookupRequest) (
```
arlolra: Hmm, how is this different from a `KeyLookup`?  Put another way, is a consistency check supposed to be ignoring TBs?

Also, `KeyLookup` be ignore the `Epoch` field entirely, as it is now?

I think these two funcs should be merged, and make better use of the fact that `Epoch` is optional.

arlolra:
Now that the policy branch has merged (that this is pull wants to merge into), if another commit merges to master and you want to include it here, you need to:

 - Merge master into policy
 - Push policy to github
 - Rebase protocol on policy
 - Push protocol

Sorry this is kind of cumbersome.  Another way to handle this is, once policy merges, close the pulls that depend on it and open new ones against master.  But that's probably not necessary.

protocol/message.go
c633: This makes me think about some changes in 7e807e5821ed0b73622108b2b1b746fc3a42f8ca. What do you think?

protocol/directory.go
c633: I think `KeyLookup` should be performed in the latest STR, and a consistency check shouldn't include a TB in its response (since it's supposed to not be performed in the same registering epoch).  
I'd like to have 2 separate methods for separate functions, just to make sure that the developers wouldn't misuse them.

arlolra: I think the only distinction between the two is on the client side, whether the client is looking up its own key versus its contact's.

Also, I don't see why you couldn't ask for your own TB and make sure the server is handing out the same one it gave you while registering, and if not, blow the whistle before the next epoch.  That's a type of consistency, though it can easily be defeated by targeted responses.

Let that be a reminder that the client needs to check the consistency of not only the TB it's been given when registering, but also all the TBs it receives on lookup (not sure if you're doing that already).

protocol/message.go
```diff
@@ -0,0 +1,64 @@
...
+
+type ErrorResponse struct {
+	Error ErrorCode
+}
+
+func NewErrorResponse(e ErrorCode) Response {
+	return &ErrorResponse{Error: e}
+}
+
+type DirectoryProof struct {
```
arlolra: Nice!

protocol/message.go
masomel: > I think `ConsistencyCheckProof` shouldn't include TB field, even if the server's using TB. I'd like to have 2 separate struct to make thing more clarify.

I see. Is the idea that `ConsistencyCheckProofs` are specifically used when the client needs to catch up on missed epochs, which is why the TB isn't needed?

c633: Yes, it is. However, @arlolra  and I have done some changes in 7e807e5821ed0b73622108b2b1b746fc3a42f8ca which maybe solved this issue.

protocol/directory.go
masomel: > I think these two funcs should be merged, and make better use of the fact that `Epoch` is optional.

I agree. This is what I was suggesting above in https://github.com/coniks-sys/coniks-go/pull/64#discussion_r74690651 as well.

> the client needs to check the consistency of not only the TB it's been given when registering, but also all the TBs it receives on lookup

I don't quite understand what you mean here. Do you mean that the client should be checking if other clients are seeing the same TB it received upon registering?

c633: > That's a type of consistency

Yes, I agree that the user can ask for his own TB. But in that case, I would like to use the key lookup instead. 

> also all the TBs it receives on lookup

Is that check the proof of absence verification? I don't see the difference between the responses of the lookup and consistency check in this case. 

As I understand, the consistency check is specifically used to verify the hash chain (i.e., the STR in different epochs), right?

arlolra: There are two different TB cases on the client side,

* Alice registers and receives a TB.  At the next epoch, Alice checks that the consistency of her key and if it's good, she can discard the TB since she no longer needs to hold that proof.  One of the things I said above is that she can also do a lookup before the next epoch and make sure the server returns the same TB she received at registration.
* Bob requests Alice's key before the next epoch and receives a TB for Alice.  Bob needs to retain the TB for Alice until the next epoch, and then perform a consistency check for Alice's key in the epoch.

So, client need to retain all the TBs it receives.  Again, maybe we're already doing this, I just wanted to be sure.

masomel: > As I understand, the consistency check is specifically used to verify the hash chain (i.e., the STR in different epochs), right? 

Consistency really involves three things for a client: ensuring that the key hasn't changed unexpectedly between epochs, verifying the hash chain, and also checking for equivocation by comparing your STR with other auditors/clients.

c633: > So, client need to retain all the TBs it receives

Yes, I agree with you in this point. But I don't understand why Bob needs to "perform a consistency check for Alice's key in the epoch"? If he does, when will he do it?

arlolra: > the client probably doesn't know what is the latest epoch before sending request

Isn't that why `Epoch` is optional?

c633: > ensuring that the key hasn't changed unexpectedly between epochs

Yes, that is my point. I mean the consistency check is specifically used in between/different epochs, right? Aww, sorry for my bad expression.

c633: > Isn't that why Epoch is optional?

I'm not so sure, since the default value of `Epoch` is 0, my issue here is I don't know how to know when `Epoch` is supplied in the request.

masomel: > she can also do a lookup before the next epoch and make sure the server returns the same TB she received at registration.

I don't know how useful this information would be. Unless the server is being overtly malicious as decides to send her a bad TB, she won't detect a difference when she looks up her own TB. I think it may be more helpful for clients to exchange TBs or something like that if we want to detect TB equivocation.

> client need to retain all the TBs it receives

Not yet in ctypes-otr, though @c633 and I discussed that we'll need to do this https://github.com/arlolra/ctypes-otr/pull/82#discussion_r73722153

arlolra: > I don't know how useful this information would be.

That's why I said above,

> That's a type of consistency, though it can easily be defeated by targeted responses.

I guess my point was that it shouldn't be a reason for a different API.

> Not yet in ctypes-otr, though @c633 and I discussed that we'll need to do this

Great!

masomel: > Yes, that is my point. I mean the consistency check is specifically used in between/different epochs, right?

Right. But the client has the latest STR stored, which contains the more recent epoch, and the client knows that the epochs are a counter.

> If he does, when will he do it? 

Bob can simply use the epoch number from the latest STR he stored, no? The clients should always monitor themselves first before doing anything else, so Bob will have the most up-to-date STR before he does any checks for other users.

arlolra: > > Isn't that why Epoch is optional?

> I'm not so sure, since the default value of Epoch is 0, my issue here is I don't know how to know when Epoch is supplied in the request.

Aha!  That's a good point.

In the code you already have,

```
+			// we don't expect to have a consistency check at epoch 0
+			if len(msg.Username) > 0 && msg.Epoch > 0 {
```

We can say `0` means latest ... but is that unsatisfying.  Going to think this for a bit.

masomel: > I guess my point was that it shouldn't be a reason for a different API.

I definitely agree on this.

> I don't know how to know when `Epoch` is supplied in the request. 

Maybe I'm missing something, but couldn't the client just omit it in the request? The server can then assume that if it's omitted, the client is requesting the latest version.

arlolra: > Maybe I'm missing something, but couldn't we annotate Epoch with omitempty as well? ~~couldn't the client just omit it in the request?~~ The server can then assume that if it's omitted, the client is requesting the latest version.

We're doing that already .The problem is the struct still has a field that golang initiates to the default value for its type.

masomel: > We're doing that already .The problem is the struct still has a field that golang initiates to the default value for its type.

I see, thanks. I just went to read more about this, too.

masomel: > We can say 0 means latest ... but is that unsatisfying. Going to think this for a bit.

Not sure this is a good idea, but we could use infinity to mean latest.

arlolra:
```
+           // we don't expect to have a consistency check at epoch 0
+           if len(msg.Username) > 0 && msg.Epoch > 0 {
```

Then again, I don't see why we're disallowing that.  You should be able to confirm your key was absent when the directory was constructed.

> Not sure this is a good idea, but we could use infinity to mean latest.

How do you express infinity in JSON though?

We probably need a `Latest` field, which makes me think we're circling back to explicit types for these request.  Seems like Huy had it right to begin with :)

Maybe we could change `ConsistencyCheck` to `KeyMonitor` and require an `Epoch` for it, which would imply TBs are nil for the request. And `KeyLookup` would never take a `Epoch`, as it's currently implemented.  So, distinct request structs.

Sorry for the long Sunday discussion everyone.

Liamsi: Trying to catch up on the discussion here:
> Then again, I don't see why we're disallowing that. You should be able to confirm your key was absent when the directory was constructed.

I agree on that. I would remove the 

> How do you express infinity in JSON though?

We could use another default (instead of 0) and call that infinity (for instance the maximum value of `uint64`). Not sure if this is a good idea, though. 

> We probably need a Latest field, which makes me think we're circling back to explicit types for these request. Seems like Huy had it right to begin with :)

I would also say that it can't harm to have separate/explicit types for different requests (and now it seems even necessary). If the structs turn out too similar one can still refactor out the common fields into a 3rd unexported struct.

c633: > Then again, I don't see why we're disallowing that. You should be able to confirm your key was absent when the directory was constructed.

Why don't we just do a lookup in this case?

masomel: >> How do you express infinity in JSON though?
>
> We could use another default (instead of 0) and call that infinity (for instance the maximum value of uint64).

Fair point. I guess, using a default max value is more what I had in mind.

> Maybe we could change ConsistencyCheck to KeyMonitor and require an Epoch for it, which would imply TBs are nil for the request. And KeyLookup would never take a Epoch, as it's currently implemented. So, distinct request structs.

I think it makes sense to try to make a distinction between lookup requests and monitoring requests, but I'm not sure epoch vs no epoch and TB vs no TB is the only distinction. As we defined monitoring (the operation), it's a self-check that you should be doing every single epoch (or catch up if you go offline). What if Bob goes offline for a couple epochs and needs to verify a message I sent him while he was offline? I don't think "monitor" should be the right request in this case since he's only interested in ensuring that my key was in the tree at that time, and that the old STR is consistent with the latest STR he saw. So I still think that a lookup with an epoch would make sense for such cases, and the server returns a nil TB and a range of STRs if its not the latest epoch. A monitoring request could include a range of epochs (i.e. start and end epoch) and could return all auth paths and STRs for the given range (this will also save the client so many individual requests for each missed epoch). What do you guys think?

I think one related question that we haven't addressed is whether we'll require Bob to check my key for all missed epochs if he wants to *send* me a message in the current epoch. In our original design, we argued that only the owner of the key could really know if the key had changed unexpectedly given that unsigned key changes are an option. So we decided that all clients trust that everyone has monitored their key and will blow the whistle if any inconsistency is detected. I think this simplifies checking in the client, but I agree that it's not perfect.

c633: Interesting!
I removed all key lookup & consistency check code (just keep the registration part). I also opened new issue #70 for our discussion.

arlolra: >> You should be able to confirm your key was absent when the directory was constructed.

> Why don't we just do a lookup in this case?

Because, what if the current epoch is 5? We're looking for proof that **at epoch 0**, a key was absent (which they all should be).  I don't think that's a misuse of any operation, though it may not be a popular request.

> What do you guys think?

Yup, `KeyMonitor` was a poor choice of name on my part.  Sorry about that.  And the details for the monitoring requests sound good.  Maybe a better name would have been `KeyLookup` and `KeyLookupInEpoch`, which do have the `Epoch` / `TB` distinctions.

protocol/directory_test.go
```diff
@@ -0,0 +1,141 @@
...
+		t.Fatal(err)
+	}
+	signKey, err := sign.GenerateKey(nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// epDeadline merkletree.TimeStamp, vrfKey *vrf.PrivateKey,
+	// signKey sign.PrivateKey, dirSize uint64,
+	// useTBs bool, capacity uint64
```
arlolra: No more capacity here

arlolra:
> I removed all key lookup & consistency check code (just keep the registration part). I also opened new issue #70 for our discussion. So we can review the other part of this pull and merge them into the master branch first.

What you have here LGTM.  Unless others object, I'm for merging it and continuing the discussion there and elsewhere.

Liamsi:
> What you have here LGTM. Unless others object, I'm for merging it and continuing the discussion there and elsewhere.

Nope, as already stated by mail: LGTM, too.

arlolra:
Merged in 57413362b0ee76d3218ceb720c7bebc221adb6f1

Issue #61, opened 2016-08-10
Bring back crypto.Commitment()


arlolra:
See the discussion at https://github.com/coniks-sys/coniks-go/pull/59#discussion-diff-74298824

My bad advice to remove it in https://github.com/coniks-sys/coniks-go/pull/1/files/1f04f1ebb6b0986505ff1b9ae6b8e1a76e78370b#r67399418

arlolra:
Fixed by 8c09142804c6a09de93c7aee0398756c56dca565

PR #66, opened 2016-08-12
A stab at #61

arlolra:
A rough sketch.  Really not sure how I feel about it.  Opening for comments though.

### DIFF
--- a/client/cgo/verification.go
+++ b/client/cgo/verification.go
@@ -111,8 +111,16 @@ func cgoVerifyAuthPath(treeHash unsafe.Pointer, treeHashSize C.int,
 	ap.TreeNonce = tn
 	ap.LookupIndex = li
 	ap.PrunedTree = pt
+	ap.Leaf = &merkletree.ProofNode{
+		Level:      uint32(leafLevel),
+		Salt:       nil,
+		Index:      leafi,
+		Value:      nil,
+		IsEmpty:    int(isLeafEmpty) == 1,
+		Commitment: leafc,
+	}
 
-	if merkletree.VerifyAuthPath(ap, leafi, leafc, uint32(leafLevel), int(isLeafEmpty) == 1, th) {
+	if ap.VerifyAuthPath(th) {
 		return 1
 	}
 	return 0
--- a/client/cgo/verification.go
+++ b/client/cgo/verification.go
@@ -119,7 +119,7 @@ func cgoVerifyAuthPath(treeHash unsafe.Pointer, treeHashSize C.int,
 		Commitment: &crypto.Commit{nil, leafc},
 	}
 
-	if ap.VerifyAuthPath(th) {
+	if ap.Verify(th) {
 		return 1
 	}
 	return 0
--- a/client/cgo/verification.go
+++ b/client/cgo/verification.go
@@ -113,11 +113,10 @@ func cgoVerifyAuthPath(treeHash unsafe.Pointer, treeHashSize C.int,
 	ap.PrunedTree = pt
 	ap.Leaf = &merkletree.ProofNode{
 		Level:      uint32(leafLevel),
-		Salt:       nil,
 		Index:      leafi,
 		Value:      nil,
 		IsEmpty:    int(isLeafEmpty) == 1,
-		Commitment: leafc,
+		Commitment: &crypto.Commit{nil, leafc},
 	}
 
 	if ap.VerifyAuthPath(th) {
@@ -139,7 +138,9 @@ func cgoVerifyCommitment(salt unsafe.Pointer, saltSize C.int,
 	keyS := C.GoStringN(key, keySize)
 	valueB := C.GoBytes(value, valueSize)
 	commitmentB := C.GoBytes(commitment, commitmentSize)
-	if merkletree.VerifyCommitment(saltB, keyS, valueB, commitmentB) {
+
+	commit := &crypto.Commit{saltB, commitmentB}
+	if commit.Verify([]byte(keyS), valueB) {
 		return 1
 	}
 	return 0
--- a/merkletree/node.go
+++ b/merkletree/node.go
@@ -22,9 +22,8 @@ type userLeafNode struct {
 	node
 	key        string
 	value      []byte
-	salt       []byte
 	index      []byte
-	commitment []byte
+	commitment *crypto.Commit
 }
 
 type emptyNode struct {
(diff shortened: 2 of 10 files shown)


merkletree/node.go
```diff
@@ -131,9 +130,8 @@ func (n *userLeafNode) Clone(parent *interiorNode) MerkleNode {
 		},
 		key:        n.key,
 		value:      n.value,
-		salt:       n.salt,
 		index:      append([]byte{}, n.index...), // make a copy of index
-		commitment: n.commitment,
+		commitment: n.commitment.Clone(),
```
c633: Why is it just `commitment: n.commitment`?

arlolra: `pNode.commitment.Salt = nil`

arlolra:
There's a much less intrusive proposal in https://github.com/coniks-sys/coniks-go/pull/59#discussion_r74302407 which maybe I prefer.  I just wanted to see what this looked like.

c633:
I prefer this implementation indeed, mostly because of `NewCommit(stuff ...[]byte)` and `commit.Verify(stuff ...[]byte)`. It also aligns with other crypto `Verify` methods.

merkletree/proof.go
```diff
@@ -65,7 +64,7 @@ func authPathHash(ap *AuthenticationPath,
 
 // VerifyAuthPath should be called after the vrf index is verified successfully
 func VerifyAuthPath(ap *AuthenticationPath,
-	leafIndex, leafCommitment []byte, leafLevel uint32, isLeafEmpty bool,
+	leafIndex []byte, leafCommitment *crypto.Commit, leafLevel uint32, isLeafEmpty bool,
```
c633: Could `leafCommitment` just be `[]byte` instead? And in the caller we could pass `leaf.Commitment().Value`, so we don't need a construction step as in https://github.com/coniks-sys/coniks-go/blob/fde67948f80829f9571c8cfb0e52ed90ee7f7532/client/cgo/verification.go#L116 (I feel little awkward when this constructor takes `nil` argument)

merkletree/node.go
arlolra: But I guess that begs the questions, is suppressing the salt necessary?

merkletree/proof.go
arlolra: I agree this is distasteful; see what I can do.

masomel:
I think https://github.com/coniks-sys/coniks-go/blob/commit/merkletree/proof_test.go#L71, L91, L129 should be using `proof.Leaf.Commitment.Value`.

merkletree/proof.go
arlolra: > And in the caller we could pass `leaf.Commitment().Value`

Ah, right, so the reason I did that was because `leaf.Commitment()` is `nil` for empty nodes, so you get a nil pointer dereference.

client/cgo/verification.go
```diff
@@ -112,8 +142,15 @@ func cgoVerifyAuthPath(treeHash unsafe.Pointer, treeHashSize C.int,
 	ap.TreeNonce = tn
 	ap.LookupIndex = li
 	ap.PrunedTree = pt
+	ap.Leaf = &leafNode{
```
arlolra: Kind of doubled down on the `nil`s here, because doing this makes the signature to the verification functions a lot cleaner, and clearer.

And, the important point, is that I think we'll see that as we finish up the protocol, the client is going to be receiving messages that unmarshall like this anyways.

masomel: > Kind of doubled down on the nils here, because doing this makes the signature to the verification functions a lot cleaner, and clearer.

Can you please elaborate on this? I might be missing something here, but I'm not sure how setting the `leafNode`'s `value` and `salt` to `nil` make the signature to the verification functions easier. Also, these two values would only be `nil` in only some cases, no?

arlolra: Well, if you set `ap.Leaf`, you can do,

```
-if merkletree.VerifyAuthPath(ap, leafi, leafc, uint32(leafLevel), int(isLeafEmpty) == 1, th) {
+if merkletree.VerifyAuthPath(ap, th) {
```

> Also, these two values would only be nil in only some cases, no?

Not if all you're doing with the leaf is calling `VerifyAuthPath`.  But that's sort of what I was getting at above in my "important point".  These client functions will probably evolve, as we flesh out the messages, to be called with a more complete leaf.

I'm thinking that eventually `cgoVerifySignature`, `cgoVerifyCommitment`, will be combined in more elaborate `cgoDoTheClientBusiness` type thing.

merkletree/node.go
c633: What do you think @masomel ?

client/cgo/verification.go
c633: > the client is going to be receiving messages that unmarshall like this anyways.

It's maybe a dumb question but do we still need to marshal the messages for transmissions?

masomel: I got that defining the leaf in the client is what simplifies the verification function signatures. I was confused about your statement that doubling down on `nil`s makes the signature simpler. Anyways, I didn't realize the `leafNode` you've set in the client is just a placeholder. The client functions will certainly evolve as we flesh out the messages. 

> eventually cgoVerifySignature, cgoVerifyCommitment, will be combined in more elaborate cgoDoTheClientBusiness type thing.

Yeah, sounds about right.

client/cgo/verification.go
```diff
@@ -111,8 +111,15 @@ func cgoVerifyAuthPath(treeHash unsafe.Pointer, treeHashSize C.int,
...
+	ap.Leaf = &merkletree.ProofNode{
+		Level:      uint32(leafLevel),
+		Index:      leafi,
+		Value:      nil,
+		IsEmpty:    int(isLeafEmpty) == 1,
+		Commitment: &crypto.Commit{nil, leafc},
+	}
 
-	if merkletree.VerifyAuthPath(ap, leafi, leafc, uint32(leafLevel), int(isLeafEmpty) == 1, th) {
+	if ap.VerifyAuthPath(th) {
```
c633: Should we just call it `ap.Verify`?

c633:
@arlolra How do you feel about f20f991bcb2d0cf8b2aff6001349f26b41897753?

arlolra:
> How do you feel about f20f991?

Looks great!  Thanks for catching my mistake.

So, I'll leave it up to you all as to whether we like this direction or not.

c633:
LGTM!

PR #67, opened 2016-08-12
Align sign and vrf package

### LINKED ISSUE #65: Align GenerateKey signatures in vrf/sign
See https://github.com/coniks-sys/coniks-go/pull/63#issuecomment-239392423

Liamsi:
- change underlying type for vrf keys to slice of bytes instead of arrays
- GenerateKey in `vrf` and `sign` package have equivalent signatures
- in `vrf` and `sign` `Public()` returns a vrf/sign `PublicKey` type
- deleted `Serialize()` ~~returns a slice of byte~~

Should resolve #65

### DIFF
--- a/crypto/sign/sign.go
+++ b/crypto/sign/sign.go
@@ -24,15 +24,15 @@ func GenerateKey(rnd io.Reader) (PrivateKey, error) {
 	return PrivateKey(sk), err
 }
 
-func (key *PrivateKey) Sign(message []byte) []byte {
-	return ed25519.Sign(ed25519.PrivateKey(*key), message)
+func (key PrivateKey) Sign(message []byte) []byte {
+	return ed25519.Sign(ed25519.PrivateKey(key), message)
 }
 
-func (key *PrivateKey) Public() (PublicKey, bool) {
-	pk, ok := ed25519.PrivateKey(*key).Public().(ed25519.PublicKey)
+func (key PrivateKey) Public() (PublicKey, bool) {
+	pk, ok := ed25519.PrivateKey(key).Public().(ed25519.PublicKey)
 	return PublicKey(pk), ok
 }
 
-func (pk *PublicKey) Verify(message, sig []byte) bool {
-	return ed25519.Verify(ed25519.PublicKey(*pk), message, sig)
+func (pk PublicKey) Verify(message, sig []byte) bool {
+	return ed25519.Verify(ed25519.PublicKey(pk), message, sig)
 }
--- a/crypto/vrf/vrf.go
+++ b/crypto/vrf/vrf.go
@@ -37,20 +37,19 @@ const (
 	ProofSize        = 32 + 32 + intermediateSize
 )
 
-type PrivateKey [PrivateKeySize]byte
-type PublicKey [PublicKeySize]byte
+type PrivateKey []byte
+type PublicKey []byte
 
 // GenerateKey creates a public/private key pair. rnd is used for randomness.
 // If it is nil, `crypto/rand` is used.
-func GenerateKey(rnd io.Reader) (pk *PublicKey, sk *PrivateKey, err error) {
+func GenerateKey(rnd io.Reader) (sk PrivateKey, err error) {
 	if rnd == nil {
 		rnd = rand.Reader
 	}
-	sk = new(PrivateKey)
-	pk = new(PublicKey)
+	sk = make([]byte, 64)
 	_, err = io.ReadFull(rnd, sk[:32])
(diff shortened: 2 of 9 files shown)


crypto/sign/sign.go
```diff
@@ -33,6 +33,10 @@ func (key *PrivateKey) Public() (PublicKey, bool) {
 	return PublicKey(pk), ok
 }
 
+func (pk *PublicKey) Serialize() []byte {
```
c633: This must be my bad. This function is being used nowhere and should be removed.

crypto/vrf/vrf.go
```diff
@@ -60,15 +59,19 @@ func GenerateKey(rnd io.Reader) (pk *PublicKey, sk *PrivateKey, err error) {
...
+}
+
+func (sk *PrivateKey) Public() (PublicKey, bool) {
+	pk, ok := ed25519.PrivateKey(*sk).Public().(ed25519.PublicKey)
+	return PublicKey(pk), ok
 }
 
-func (sk *PrivateKey) Public() []byte {
-	return ed25519.PrivateKey(sk[:]).Public().(ed25519.PublicKey)
+func (pk *PublicKey) Serialize() []byte {
```
c633: Same as https://github.com/coniks-sys/coniks-go/pull/67/files#r74655028

crypto/vrf/vrf_test.go
```diff
@@ -29,23 +30,28 @@ func TestHonestComplete(t *testing.T) {
 	}
 }
 
-func TestConvertSecretKeyToPublicKey(t *testing.T) {
-	pk, sk, err := GenerateKey(nil)
+func TestConvertPrivateKeyToPublicKey(t *testing.T) {
```
c633: I think we can also remove this test if we remove `Serialize`.

crypto/sign/sign.go
Liamsi: No problem. I'll also remove the one in sign.go.

crypto/vrf/vrf_test.go
Liamsi: Do you think it doesn't make sense to test if `Public()` succeeds and the returned bytes match `sk[32:]`? I would prefer to have a test for `Public` and this were the only two checks that came to my mind.

c633: > Do you think it doesn't make sense to test if Public() succeeds and the returned bytes match sk[32:]

Yes, it makes sense when I'm thinking about #50. Sorry for the noise :(

crypto/vrf/vrf.go
```diff
@@ -60,15 +59,15 @@ func GenerateKey(rnd io.Reader) (pk *PublicKey, sk *PrivateKey, err error) {
...
 
-func (sk *PrivateKey) Public() []byte {
-	return ed25519.PrivateKey(sk[:]).Public().(ed25519.PublicKey)
+func (sk *PrivateKey) Public() (PublicKey, bool) {
+	pk, ok := ed25519.PrivateKey(*sk).Public().(ed25519.PublicKey)
+	return PublicKey(pk), ok
 }
 
-func (sk *PrivateKey) expandSecret() (x, skhr *[32]byte) {
+func (sk PrivateKey) expandSecret() (x, skhr *[32]byte) {
```
arlolra: Why the change here (and throughout) away from pointers?

Liamsi: Because the Private/PublicKeys are now slices of bytes which (even if passed by value/used as value receiver) already behave more like a pointer (which makes the pointer receiver obsolete, only the "header" gets copied). Also, it made it easier to treat it as a slice of byte (otherwise we would need to add a `*` everywhere where we do a byte operation). But I can change it back to pointer receivers.

Liamsi: I think, we should also align this in sign and vrf packages, too. As explained above pointers bring no benefit here and I would prefer values but pointers won't harm and have no real downside (some extra `*` here and there).

arlolra: My preference for pointers here is that I've made the mistake of trying to mutate the value, not remembering that these methods are syntactic sugar, and wondering why my structure wasn't updating.  But that's just me not being very bright.

I'm fine with it either way, under the condition that we change them everywhere to be the same.

c633:
> I'm fine with it either way, under the condition that we change them everywhere to be the same.

I updated in a230decdadf05853a07395517ed9b46a16f1e998.

This pull looks good to me.

arlolra:
LGTM

PR #68, opened 2016-08-13
Account verification protocol documentation

### LINKED ISSUE #56: Write account verification protocol spec
We'd like a more formal specification of this protocol. Related to #33.

masomel:
Should resolve #56.


c633:
Can you merge this pull with #33 ?

bots/doc.go
```diff
@@ -0,0 +1,69 @@
...
+register the username and the public key with the CONIKS server.
+
+The registration proxy is connected to reserved user account with the
+same third-party service, allowing the CONIKS client to send messages
+to the proxy using the service's communication system. Therefore, in
+order to register a new username-to-key mapping, the client must send
+the request to the proxy's reserved third-party account.
+
+Privacy is a key feature of CONIKS, so the only piece of third-party
+account information that the CONIKS client and proxy store (or even see?),
```
Liamsi: The CONIKS client also sees (stores and generates) the (private-)key. Maybe, we should leave away the client in this sentence?

bots/doc.go
```diff
@@ -0,0 +1,69 @@
...
+
+We are most concerned with an attacker attempting to impersonate
+a user of a third-party service (e.g. Twitter, IRC)
+by registering a name-to-key mapping for this user with the CONIKS server.
+
+Requiring that the legitimate user connect to her third-party account
+using the correct credentials serves as a proof of ownership of
+the account, and we assume that obtaining this user's credentials is a
+sufficiently large burden to deter most adversaries from mounting the
+impersonation attack.
```
Liamsi: Shouldn't we explicitly mention that (because of above threat-model) we do not check if the user owns the corresponding private-key? (That's something I would expect the registration bot also to do. But maybe it's just me.)

c633: Yeah, I was thinking about the unknown key share attack. I'm thinking that the user would sign his Twitter handle and send the signature along with the request. What do you think?

Liamsi: > I'm thinking that the user would sign his Twitter handle and send the signature along with the request. 

I would approve of that idea. IIRC the problems with that approach was that we eventually would have to deal with a set of different key-types/encodings etc, right?

c633: > IIRC the problems with that approach was that we eventually would have to deal with a set of different key-types/encodings etc, right?

Yes.

Liamsi: Maybe we can just mention that although it is not in our threat-model, we try to mitigate this kind of attack where feasible and where it makes sense (like in the TorMessenger integration).

arlolra: See "§4.3.1 Visibility of Public Keys".  The server needs to consider the "key" to be an opaque blob for the strict policy to work.

Also, in Tor Messenger, I thought we decided on only storing the fingerprint.

bots/doc.go
masomel: This is true, I mention that the client generates the key pair on L14. This paragraph is supposed to address questions about the third-party account information. But I think we should mention what the client *does* store to be completely transparent.

masomel:
> Can you merge this pull with #33 ?

I wanted to keep this separate from #33 while both pulls are still under review to keep the pulls more manageable. Is it ok if I merge this with #33 once this pull has been reviewed or when #33 is ready to be merged, whichever happens first?

bots/doc.go
masomel: > Shouldn't we explicitly mention that (because of above threat-model) we do not check if the user owns the corresponding private-key? (That's something I would expect the registration bot also to do. But maybe it's just me.)

@Liamsi You mean the registration bot doesn't need to check if the user owns the private key because we assume that receiving a registration request from that user/third-party account with the corresponding public key is enough proof of ownership?

> Yeah, I was thinking about the unknown key share attack. I'm thinking that the user would sign his Twitter handle and send the signature along with the request. What do you think?
>
> Maybe we can just mention that although it is not in our threat-model, we try to mitigate this kind of attack where feasible and where it makes sense (like in the TorMessenger integration).

I would also be more comfortable having a solution to this attack, but given what @arlolra points out about supporting the strict key visibility policy, I'm not sure how we can mitigate this threat. 

> Also, in Tor Messenger, I thought we decided on only storing the fingerprint.

We did, though I want to keep this document as general as possible. I think how we've adapted this protocol for Tor Messenger should be part of the Tor Messenger CONIKS documentation.

masomel:
I was just reading about OAuth-signed HTTP requests, @c633 we already use these, right? I'm assuming the client also sends OAuth-signed requests? I'd like to show we provide stronger guarantees than just "trusting the connection between the third-party account and client".

c633:
> I was just reading about OAuth-signed HTTP requests, @c633 we already use these, right?

It's implemented in the twitter module. I think we already use it, but @arlolra would have the exact answer.

bots/doc.go
```diff
@@ -1,69 +1,135 @@
...
+1. The user registers her third-party account with her communication
+client, thereby authorizing the client and the integrated CONIKS client
+to connect to her third-party account.
+
+2. The CONIKS client generates a key pair for the new user account
+and stores the public and private key locally.
+
+3. On behalf of the connected account, the client sends a registration
+request
+    req = (username, opaque public key data blob)
```
c633: I think we can remove the `username` part, and let the bot insert this field into the request (as my current implementation).

bots/doc.go
```diff
@@ -1,69 +1,135 @@
...
+to the designated account proxy using the service's communication
+protocol.
 
-2. The client sends a registration request containing the generate public key to the proxy's account using the service's communication protocol using the added username.
+4. The proxy receives the request at its reserved account via
+the service's protocol, and verifies that the sender
+is an authorized client.
 
-3. The proxy verifies that the sender of the request is a legitimate account, and sends a registration request to the CONIKS server.
+5. The proxy digitally signs the verified registration request,
```
c633: Could we change this into something like:
"The proxy sends the verified registration request to its designated CONIKS server. This request can be verified by a digital signature signed by the the proxy or sent through a local secure connection between the proxy and the CONIKS server."

bots/doc.go
masomel: Hmm, `req` was more for demonstration purposes, since the bot inserting the field into the request seems more like an implementation detail.

bots/doc.go
masomel: I decided to add this information as part of the system model instead, starting at https://github.com/coniks-sys/coniks-go/blob/reg-bot-doc/bots/doc.go#L64. What do you think?

c633: It looks great!

masomel:
Do you have any comments on this, @arlolra @Liamsi, before I merge this into #33?

arlolra:
Sorry, I haven't had a chance to read it yet.  Give me a day before timing out.  (Also, I'm guessing you just mean doc.go here?  It'd be helpful to rebase these changes so that only your commits show up here in the UI).  Thanks!

bots/doc.go
```diff
@@ -0,0 +1,139 @@
...
+
+Limitations
+
+While bots provides strong security and privacy guarantees against
+modest attackers, our protocol is not reboust against motivated,
+resourceful attackers; these may be able to obtain the necessary
+credentials to authorize a malicious CONIKS client and impersonate a
+target user. Additionally, we avoid exposing the user's public key to
+the registration proxy, which favors user privacy, but this also limits
+the proxy's ability to verify explcitly the ownership of this public key
```
arlolra: typo, explicitly

arlolra:
This is great.  The one thing I would add to the limitations is that

bots/doc.go
```diff
@@ -0,0 +1,139 @@
...
+
+Supported Third-Party Services
+
+The current implementation supports third-party account verification of
+Twitter accounts. The account proxy connects to its reserved Twitter
+account using OAuth, and it only accepts CONIKS registration
+requests received as Twitter direct messages sent from CONIKS clients
+that have been authorized by a legitimate Twitter account.
+
+Limitations
```
arlolra: The limitations should emphasize the differences between first- and third-party registration, no?  And I don't see obtaining user credentials as any different in those scenarios.  Also, presumably the bots are run by the same people (third-party) running the keyserver, so I think that's a general, third-party running your keyserver, limitation and not specific to registration.

The main thing here is that we don't know, in the case of equivocation or whatnot, who is being malicious, the keyserver or the communication server.

arlolra:
This is great; thanks for writing it.  One little quibble with the limitations.

(And sorry for the close just now, clicked the wrong button.)

bots/doc.go
```diff
@@ -0,0 +1,139 @@
...
+
+While bots provides strong security and privacy guarantees against
+modest attackers, our protocol is not reboust against motivated,
```
Liamsi: typo, s/reboust/robust

bots/doc.go
masomel: > The limitations should emphasize the differences between first- and third-party registration, no? 

Hmm, it seems like we're flipping the first party and third party. From the key server's point of view, the first party would be a communication service like Tor Messenger running the key server, and the third party would be Twitter (the identity provider). But I do agree that we should emphasize the differences between the two. I think the place to do that, though, is in the system model section.

> And I don't see obtaining user credentials as any different in those scenarios.

Yeah, you're right. 

> presumably the bots are run by the same people (third-party) running the keyserver, so I think that's a general, third-party running your keyserver, limitation and not specific to registration.

I'm assuming, you're referring to the "proxy cannot verify the ownership of the key" limitation here.

> The main thing here is that we don't know, in the case of equivocation or whatnot, who is being malicious, the keyserver or the communication server.

This is definitely true, but I'm a bit confused as to how this issue relates to the registration bot. The registration bot's purpose isn't to detect equivocation.

arlolra: > Hmm, it seems like we're flipping the first party and third party.

Sounds like it, sorry.  In my comment I meant,

- First party: keyserver is run by identity provider
- Third party: keyserver is run by communication service (or really anyone) on behalf of the identify provider

> I'm assuming, you're referring to the "proxy cannot verify the ownership of the key" limitation here.

No, verifying the ownership of the key is limited by the `allowsPublicVisibility` flag, and seems the same in the two cases.

I was talking about, "we avoid exposing the user's public key to the registration proxy" (which is similarly affected by that flag, but let's ignore that for now).  Is that even true though, we aren't encrypting the data?  What I meant above is that since the proxy and keyserver are both currently being run by the same person, it's not really an additional limitation.

> This is definitely true, but I'm a bit confused as to how this issue relates to the registration bot.

I guess because this is where the breakdown happens?  In the third-party case (my terminology above), we have no proof that the identity provider requested the registration.  If we did, then we'd be able to determine who was being malicious.  But also because I wasn't sure where else we had planned to write this up.

masomel:
Still working on this, sorry for the delay. I will merge this directly into master.

bots/doc.go
masomel: > - First party: keyserver is run by identity provider
- Third party: keyserver is run by communication service (or really anyone) on behalf of the identify provider

Ok, gotcha. I think, though, that the bot is mostly relevant in the third-party case, no? I'm not sure that a first-party would run a bot.

> verifying the ownership of the key is limited by the allowsPublicVisibility flag, and seems the same in the two cases.

Right.

>  I was talking about, "we avoid exposing the user's public key to the registration proxy" (which is similarly affected by that flag, but let's ignore that for now).

Ah I see, but the limitation I was pointing out wasn't the we avoid the exposure of the public key, but rather that "this also limits the proxy's ability to verify explicitly the ownership of this public key". But given the use of the `allowsPublicVisibility` flag, this might not be feasible anyway. I think the real reason why this is a limitation isn't so much that the data isn't exposed, but rather because the bot can't make any assumptions on what the key data looks like. 

> Is that even true though, we aren't encrypting the data?

What I meant by "avoid exposing the public key" is that the client only sends the fingerprint, but this is an implementation detail in the client and because of the fact that the bot doesn't make any assumptions about the key data, I don't think this point should be mentioned. But I don't think we're encrypting the data sent to the bot, unless https://github.com/c633/ctypes-otr/blob/coniks-registration/chrome/content/coniks/coniks.js#L207 encrypts the data. And AFAIK, the Twitter OAuth API only signs the requests.

>  In the third-party case (my terminology above), we have no proof that the identity provider requested the registration.  If we did, then we'd be able to determine who was being malicious. But also because I wasn't sure where else we had planned to write this up.

Ah, this is the difference between the first- and third-party registration you were saying I should describe? It's true, the bot doesn't generate cryptographic proof that the identity provider made the request (I think we had briefly talked about this at some point, about somehow using the TLS handshake as cryptographic proof of the identity provider's involvement). I think the registration bot documentation is a good place for writing this up since I think this is mostly problematic during registration with a third-party keyserver.

masomel:
This pull is ready for another round of reviews. I haven't figured out how to fix the failing Travis builds yet.

arlolra:
Just run `go fmt ./...`

bots/doc.go
```diff
@@ -0,0 +1,191 @@
...
+communication services on top of their existing communication services.
+As a result, these third-party communication services must also provide
+encryption key management for their users.
+
+CONIKS is a key management and verification system which ensures that 
+the public keys maintained for users in CONIKS key directories remain 
+consistent over time and across different vantage points in the system. 
+But CONIKS only ensures the consistency of online identities, and makes 
+no guarantees about the persons who own the usernames registered in the
+key directories. This this less of an issue for secure communication
```
arlolra: s/This this/This is/

bots/doc.go
```diff
@@ -0,0 +1,191 @@
...
+Users: In order to use CONIKS-backed secure communication services that 
+do not provide new identities, users must authorize the service
+to connect to their first-party account.
+bots leverages this authorization in its account verification protocol,
+and trusts that the service's clients use a secure authorization procotol
+such as OAuth.
+
+CONIKS clients: Users run the third-party secure communication client 
+software with an integrated CONIKS client. The CONIKS client software 
+stores a for each of the user's registered
```
arlolra: stores a?

arlolra:
LGTM

masomel:
> Just run go fmt ./...

Ah good point. Not sure what I changed that's causing it to fail now. Anyways, thanks for reminding me!

Issue #45, opened 2016-07-30
Logging


arlolra:
Some form of sanitized logs will probably be useful for operating the server, at various severity levels.

c633:
I propose to use [zap](https://github.com/uber-go/zap) since it has nice structured log messages, supports verbose, JSON serialized log messages, ... They also claimed that `zap` provides the best performance compare with other log packages (https://github.com/uber-go/zap#performance). 
I will open a pull request for this after #33 is merged.

Liamsi:
Never heard about zap before (also suprised that uber has so many golang libs opensourced).
It looks a little verbose while coding but the performance looks very impressive. Good choice.

c633:
Use `zap.SugaredLogger` for now. See its APIs document here: https://github.com/uber-go/zap/blob/master/README.md#quick-start

```
logger.Infow(message, 
             key, value, 
             ...)
```
outputs:

`{"msg":message,"key":value}`

and 
```
logger.Infof("sprintf formatted %s", "string")
```
outputs:
 `"msg: sprintf formatted string"`

PR #71, opened 2016-08-15
Implement reference key server

c633:
* Implement registration function

### DIFF
--- a/keyserver/README.md
+++ b/keyserver/README.md
@@ -1,0 +1,42 @@
+# CONIKS Server implementation in Golang
+
+## Usage
+```
+⇒  go install github.com/coniks-sys/coniks-go/keyserver/coniksserver
+⇒  coniksserver -h
+ _______  _______  __    _  ___  ___   _  _______
+|       ||       ||  |  | ||   ||   | | ||       |
+|       ||   _   ||   |_| ||   ||   |_| ||  _____|
+|       ||  | |  ||       ||   ||      _|| |_____
+|      _||  |_|  ||  _    ||   ||     |_ |_____  |
+|     |_ |       || | |   ||   ||    _  | _____| |
+|_______||_______||_|  |__||___||___| |_||_______|
+
+Usage:
+  coniksserver [command]
+
+Available Commands:
+  init        Create a configuration file and generate all keys
+  run         Run a CONIKS server instance
+
+Flags:
+  -h, --help   help for coniksserver
+
+Use "coniksserver [command] --help" for more information about a command.
+```
+
+Run the server
+```
+⇒  mkdir coniks; cd coniks
+⇒  coniksserver init -c # creates all files including a self-signed ssl keys/cert
+⇒  coniksserver run -p  # run & write down the process ID into coniks.pid
+```
+
+You can reload the server's policies while it's running by editing the `config.toml` file
+and possibly replace `vrf.priv` with a new key, then run
+```
+⇒  kill -USR2 `cat coniks.pid`
+```
+
+## Disclaimer
+Please keep in mind that this CONIKS server implementation is under active development. The repository may contain experimental features that aren't fully tested.
--- a/keyserver/coniksserver/cmd/init.go
+++ b/keyserver/coniksserver/cmd/init.go
@@ -1,0 +1,105 @@
+package cmd
+
+import (
+	"bytes"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
+	"strconv"
+
+	"github.com/BurntSushi/toml"
+	"github.com/coniks-sys/coniks-go/crypto/sign"
+	"github.com/coniks-sys/coniks-go/crypto/vrf"
+	"github.com/coniks-sys/coniks-go/keyserver"
+	"github.com/coniks-sys/coniks-go/keyserver/testutil"
+	"github.com/coniks-sys/coniks-go/utils"
+	"github.com/spf13/cobra"
+)
+
+// initCmd represents the init command
+var initCmd = &cobra.Command{
+	Use:   "init",
+	Short: "Create a configuration file and generate all keys",
+	Long:  `Create a configuration file and generate all keys for signing and VRF`,
+	Run: func(cmd *cobra.Command, args []string) {
+		dir := cmd.Flag("dir").Value.String()
+		mkConfig(dir)
+		mkSigningKey(dir)
+		mkVrfKey(dir)
+
+		cert, err := strconv.ParseBool(cmd.Flag("cert").Value.String())
+		if err == nil && cert {
+			testutil.CreateTLSCert(dir)
+		}
+	},
+}
+
+func init() {
+	RootCmd.AddCommand(initCmd)
+	initCmd.Flags().StringP("dir", "d", ".", "Location of directory for storing generated files")
+	initCmd.Flags().BoolP("cert", "c", false, "Generate self-signed ssl keys/cert with sane defaults")
+}
+
+func mkConfig(dir string) {
+	file := path.Join(dir, "config.toml")
+	var conf = keyserver.ServerConfig{
+		DatabasePath:        "coniks.db",
+		LoadedHistoryLength: 1000000,
+		TLS: &keyserver.TLSConnection{
+			LocalAddress:  "127.0.0.1:3001",
+			PublicAddress: "0.0.0.0:3000",
+			TLSCertPath:   "server.pem",
+			TLSKeyPath:    "server.key",
+		},
+		Policies: &keyserver.ServerPolicies{
+			EpochDeadline: 60,
+			VRFKeyPath:    "vrf.priv",
+			SignKeyPath:   "sign.priv",
+		},
+	}
+
+	var confBuf bytes.Buffer
+
+	e := toml.NewEncoder(&confBuf)
+	err := e.Encode(conf)
+	if err != nil {
+		log.Println(err)
+		return
+	}
+	util.WriteFile(file, confBuf)
+}
+
+func mkSigningKey(dir string) {
+	file := path.Join(dir, "sign.priv")
+	if _, err := os.Stat(file); err == nil {
+		log.Printf("%s already exists\n", file)
+		return
+	}
+	sk, err := sign.GenerateKey(nil)
+	if err != nil {
+		log.Print(err)
+		return
+	}
+	if err := ioutil.WriteFile(file, sk[:], 0600); err != nil {
+		log.Print(err)
+		return
+	}
+}
+
+func mkVrfKey(dir string) {
+	file := path.Join(dir, "vrf.priv")
+	if _, err := os.Stat(file); err == nil {
+		log.Printf("%s already exists\n", file)
+		return
+	}
+	sk, err := vrf.GenerateKey(nil)
+	if err != nil {
+		log.Print(err)
+		return
+	}
+	if err := ioutil.WriteFile(file, sk[:], 0600); err != nil {
+		log.Print(err)
+		return
+	}
+}
--- a/keyserver/coniksserver/cmd/run.go
+++ b/keyserver/coniksserver/cmd/run.go
@@ -1,0 +1,65 @@
+package cmd
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"os/signal"
+	"path"
+	"strconv"
+
+	"github.com/coniks-sys/coniks-go/keyserver"
+	"github.com/spf13/cobra"
+)
+
+// runCmd represents the run command
+var runCmd = &cobra.Command{
+	Use:   "run",
+	Short: "Run a CONIKS server instance",
+	Long: `Run a CONIKS server instance
+
+This will look for config files with default names (config.toml) 
+in the current directory if not specified differently.
+	`,
+	Run: func(cmd *cobra.Command, args []string) {
+		config := cmd.Flag("config").Value.String()
+		pid, _ := strconv.ParseBool(cmd.Flag("pid").Value.String())
+		// ignore the error here since it is handled by the flag parser.
+		if pid {
+			writePID()
+		}
+		run(config)
+	},
+}
+
+func init() {
+	RootCmd.AddCommand(runCmd)
+	runCmd.Flags().StringP("config", "c", "config.toml", "Path to server configuration file")
+	runCmd.Flags().BoolP("pid", "p", false, "Write down the process id to coniks.pid in the current working directory")
+}
+
+func run(confPath string) {
+	conf, err := keyserver.LoadServerConfig(confPath)
+	if err != nil {
+		log.Fatal(err)
+	}
+	serv := keyserver.NewConiksServer(conf)
+
+	// run the server until receiving an interrupt signal
+	serv.Run(conf.TLS)
+	ch := make(chan os.Signal, 1)
+	signal.Notify(ch, os.Interrupt)
+	<-ch
+	serv.Shutdown()
+}
+
+func writePID() {
+	pidf, err := os.OpenFile(path.Join(".", "coniks.pid"), os.O_CREATE|os.O_WRONLY, 0666)
+	if err != nil {
+		log.Printf("Cannot create coniks.pid: %v", err)
+		return
+	}
+	if _, err := fmt.Fprint(pidf, os.Getpid()); err != nil {
+		log.Printf("Cannot write to pid file: %v", err)
+	}
+}
--- a/keyserver/coniksserver/coniksserver.go
+++ b/keyserver/coniksserver/coniksserver.go
@@ -1,0 +1,7 @@
+package main
+
+import "github.com/coniks-sys/coniks-go/keyserver/coniksserver/cmd"
+
+func main() {
+	cmd.Execute()
+}
--- a/keyserver/listener.go
+++ b/keyserver/listener.go
@@ -1,0 +1,114 @@
+package keyserver
+
+import (
+	"crypto/tls"
+	"io"
+	"log"
+	"net"
+	"time"
+
+	. "github.com/coniks-sys/coniks-go/protocol"
+)
+
+func (server *ConiksServer) listenForRequests(ln *net.TCPListener, handler func(msg []byte) ([]byte, error)) {
+	defer ln.Close()
+	go func() {
+		<-server.stop
+		ln.SetDeadline(time.Now())
+	}()
+
+	for {
+		select {
+		case <-server.stop:
+			server.waitCloseConn.Wait()
+			return
+		default:
+		}
+		conn, err := ln.Accept()
+		if err != nil {
+			if opErr, ok := err.(*net.OpError); ok && opErr.Timeout() {
+				continue
+			}
+			log.Printf("accept client: %s", err)
+			continue
+		}
+		conn = tls.Server(conn, server.tlsConfig)
+		server.waitCloseConn.Add(1)
+		go func() {
+			server.acceptClient(conn, handler)
+			server.waitCloseConn.Done()
+		}()
+	}
+}
+
+func (server *ConiksServer) acceptClient(conn net.Conn, handler func(msg []byte) ([]byte, error)) {
+	defer conn.Close()
+
+	conn.SetDeadline(time.Now().Add(5 * time.Second))
+
+	// handle request
+	buf := make([]byte, 4096)
+	for {
+		n, err := conn.Read(buf)
+		if err != nil {
+			if err != io.EOF {
+				log.Printf("client read %v: %v", conn.RemoteAddr(), err)
+			}
+			return
+		}
+
+		res, err := handler(buf[:n])
+		if err != nil {
+			log.Printf("client handle %v: %v", conn.RemoteAddr(), err)
+		}
+
+		n, err = conn.Write([]byte(res))
+		if err != nil {
+			log.Printf("client write %v: %v", conn.RemoteAddr(), err)
+			return
+		}
+	}
+}
+
+func malformedClientMsg(err error) ([]byte, error) {
+	// check if we're just propagating a message
+	if err == nil {
+		err = ErrorMalformedClientMessage.Error()
+	}
+	response := NewErrorResponse(ErrorMalformedClientMessage)
+	res, e := MarshalResponse(response)
+	if e != nil {
+		panic(e)
+	}
+	return res, err
+}
+
+func (server *ConiksServer) makeHandler(acceptableTypes map[int]bool) func(msg []byte) ([]byte, error) {
+	return func(msg []byte) ([]byte, error) {
+		// get request message
+		req, err := UnmarshalRequest(msg)
+		if err != nil {
+			return malformedClientMsg(err)
+		}
+		if !acceptableTypes[req.Type] {
+			log.Printf("unacceptable message type: %q", req.Type)
+			return malformedClientMsg(ErrorMalformedClientMessage.Error())
+		}
+
+		switch req.Type {
+		default:
+			server.Lock()
+		}
+		response, e := server.dir.HandleOps(req)
+		switch req.Type {
+		default:
+			server.Unlock()
+		}
+
+		res, err := MarshalResponse(response)
+		if err != nil {
+			panic(err)
+		}
+		return res, e.Error()
+	}
+}
--- a/keyserver/server.go
+++ b/keyserver/server.go
@@ -1,0 +1,224 @@
+package keyserver
+
+import (
+	"crypto/tls"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net"
+	"os"
+	"os/signal"
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/BurntSushi/toml"
+	"github.com/coniks-sys/coniks-go/crypto/sign"
+	"github.com/coniks-sys/coniks-go/crypto/vrf"
+	"github.com/coniks-sys/coniks-go/merkletree"
+	"github.com/coniks-sys/coniks-go/protocol"
+	"github.com/coniks-sys/coniks-go/storage/kv"
+	"github.com/coniks-sys/coniks-go/storage/kv/leveldbkv"
+	"github.com/coniks-sys/coniks-go/utils"
+)
+
+type ServerConfig struct {
+	configFilePath      string
+	DatabasePath        string          `toml:"database"`
+	LoadedHistoryLength uint64          `toml:"loaded_history_length"`
+	Policies            *ServerPolicies `toml:"policies"`
+	TLS                 *TLSConnection  `toml:"tls"`
+}
+
+type TLSConnection struct {
+	PublicAddress string `toml:"public_address"` // address:port
+	LocalAddress  string `toml:"local_address"`
+	TLSCertPath   string `toml:"cert"`
+	TLSKeyPath    string `toml:"key"`
+}
+
+type ServerPolicies struct {
+	EpochDeadline merkletree.TimeStamp `toml:"epoch_deadline"`
+	VRFKeyPath    string               `toml:"vrf_key_path"`
+	SignKeyPath   string               `toml:"sign_key_path"` // it should be a part of policies, see #47
+	vrfKey        vrf.PrivateKey
+	signKey       sign.PrivateKey
+}
+
+type ConiksServer struct {
+	sync.RWMutex
+	dir *protocol.ConiksDirectory
+
+	stop          chan struct{}
+	waitStop      sync.WaitGroup
+	waitCloseConn sync.WaitGroup
+
+	db kv.DB // TODO: it is a placeholer for issue #37
+
+	configFilePath string
+	reloadChan     chan os.Signal
+	epochTimer     *time.Timer
+	tlsConfig      *tls.Config
+}
+
+func LoadServerConfig(file string) (*ServerConfig, error) {
+	var conf ServerConfig
+	if _, err := toml.DecodeFile(file, &conf); err != nil {
+		return nil, fmt.Errorf("Failed to load config: %v", err)
+	}
+
+	// load signing key
+	signPath := util.ResolvePath(conf.Policies.SignKeyPath, file)
+	signKey, err := ioutil.ReadFile(signPath)
+	if err != nil {
+		return nil, fmt.Errorf("Cannot read signing key: %v", err)
+	}
+	if len(signKey) != sign.PrivateKeySize {
+		return nil, fmt.Errorf("Signing key must be 64 bytes (got %d)", len(signKey))
+	}
+
+	// load VRF key
+	vrfPath := util.ResolvePath(conf.Policies.VRFKeyPath, file)
+	vrfKey, err := ioutil.ReadFile(vrfPath)
+	if err != nil {
+		return nil, fmt.Errorf("Cannot read VRF key: %v", err)
+	}
+	if len(vrfKey) != vrf.PrivateKeySize {
+		return nil, fmt.Errorf("VRF key must be 64 bytes (got %d)", len(vrfKey))
+	}
+
+	conf.configFilePath = file
+	conf.Policies.vrfKey = vrfKey
+	conf.Policies.signKey = signKey
+	// also update path for db & TLS cert files
+	conf.DatabasePath = util.ResolvePath(conf.DatabasePath, file)
+	conf.TLS.TLSCertPath = util.ResolvePath(conf.TLS.TLSCertPath, file)
+	conf.TLS.TLSKeyPath = util.ResolvePath(conf.TLS.TLSKeyPath, file)
+
+	return &conf, nil
+}
+
+func NewConiksServer(conf *ServerConfig) *ConiksServer {
+	// open db
+	kvdb := leveldbkv.OpenDB(conf.DatabasePath)
+
+	// create server instance
+	server := new(ConiksServer)
+	server.dir = protocol.InitDirectory(
+		conf.Policies.EpochDeadline,
+		conf.Policies.vrfKey,
+		conf.Policies.signKey,
+		conf.LoadedHistoryLength,
+		true)
+	server.stop = make(chan struct{})
+	server.db = kvdb
+	server.configFilePath = conf.configFilePath
+	server.reloadChan = make(chan os.Signal, 1)
+	signal.Notify(server.reloadChan, syscall.SIGUSR2)
+	server.epochTimer = time.NewTimer(time.Duration(conf.Policies.EpochDeadline) * time.Second)
+
+	return server
+}
+
+func (server *ConiksServer) Run(tc *TLSConnection) {
+	server.waitStop.Add(1)
+	go func() {
+		server.EpochUpdate()
+		server.waitStop.Done()
+	}()
+
+	// server listener
+	cer, err := tls.LoadX509KeyPair(tc.TLSCertPath, tc.TLSKeyPath)
+	if err != nil {
+		panic(err)
+	}
+
+	server.tlsConfig = &tls.Config{Certificates: []tls.Certificate{cer}}
+
+	addr, err := net.ResolveTCPAddr("tcp", tc.PublicAddress)
+	if err != nil {
+		panic(err)
+	}
+
+	// server public port
+	publicLn, err := net.ListenTCP("tcp", addr)
+	if err != nil {
+		panic(err)
+	}
+
+	addr, err = net.ResolveTCPAddr("tcp", tc.LocalAddress)
+	if err != nil {
+		panic(err)
+	}
+
+	// server registration port
+	localLn, err := net.ListenTCP("tcp", addr)
+	if err != nil {
+		panic(err)
+	}
+
+	// acceptable types for public connection
+	publicTypes := make(map[int]bool)
+	server.waitStop.Add(1)
+	go func() {
+		server.listenForRequests(publicLn, server.makeHandler(publicTypes))
+		server.waitStop.Done()
+	}()
+
+	// acceptable types for local connection
+	localTypes := make(map[int]bool)
+	localTypes[protocol.RegistrationType] = true
+	server.waitStop.Add(1)
+	go func() {
+		server.listenForRequests(localLn, server.makeHandler(localTypes))
+		server.waitStop.Done()
+	}()
+
+	server.waitStop.Add(1)
+	go func() {
+		server.updatePolicies()
+		server.waitStop.Done()
+	}()
+}
+
+func (server *ConiksServer) EpochUpdate() {
+	for {
+		select {
+		case <-server.stop:
+			return
+		case <-server.epochTimer.C:
+			server.Lock()
+			server.dir.Update()
+			server.epochTimer.Reset(time.Duration(server.dir.EpochDeadline()) * time.Second)
+			server.Unlock()
+		}
+	}
+}
+
+func (server *ConiksServer) Shutdown() error {
+	close(server.stop)
+	server.waitStop.Wait()
+	return server.db.Close()
+}
+
+func (server *ConiksServer) updatePolicies() {
+	for {
+		select {
+		case <-server.stop:
+			return
+		case <-server.reloadChan:
+			// read server policies from config file
+			conf, err := LoadServerConfig(server.configFilePath)
+			if err != nil {
+				log.Println(err)
+				// error occured while reading server config
+				// simply abort the reloading policies process
+				return
+			}
+			server.Lock()
+			server.dir.SetPolicies(conf.Policies.EpochDeadline, conf.Policies.vrfKey)
+			server.Unlock()
+			log.Println("Policies reloaded!")
+		}
+	}
+}
--- a/keyserver/server_test.go
+++ b/keyserver/server_test.go
@@ -1,0 +1,239 @@
+package keyserver
+
+import (
+	"bytes"
+	"encoding/json"
+	"path"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/coniks-sys/coniks-go/crypto/sign"
+	"github.com/coniks-sys/coniks-go/crypto/vrf"
+	"github.com/coniks-sys/coniks-go/keyserver/testutil"
+	"github.com/coniks-sys/coniks-go/merkletree"
+	. "github.com/coniks-sys/coniks-go/protocol"
+	"github.com/coniks-sys/coniks-go/storage/kv"
+	"github.com/coniks-sys/coniks-go/utils"
+)
+
+func startServer(t *testing.T, kvdb kv.DB, epDeadline merkletree.TimeStamp, policiesPath string) (*ConiksServer, func()) {
+	dir, teardown := testutil.CreateTLSCertForTest(t)
+
+	signKey, err := sign.GenerateKey(nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	vrfKey, err := vrf.GenerateKey(nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	conf := &ServerConfig{
+		DatabasePath:        path.Join(dir, "coniks.db"),
+		LoadedHistoryLength: 100,
+		TLS: &TLSConnection{
+			PublicAddress: testutil.PublicConnection,
+			LocalAddress:  testutil.LocalConnection,
+			TLSCertPath:   path.Join(dir, "server.pem"),
+			TLSKeyPath:    path.Join(dir, "server.key"),
+		},
+		Policies: &ServerPolicies{
+			EpochDeadline: epDeadline,
+			vrfKey:        vrfKey,
+			signKey:       signKey,
+		},
+	}
+
+	server := NewConiksServer(conf)
+	server.db = kvdb
+	server.Run(conf.TLS)
+	return server, func() {
+		server.Shutdown()
+		teardown()
+	}
+}
+
+func TestServerStartStop(t *testing.T) {
+	util.WithDB(func(db kv.DB) {
+		_, teardown := startServer(t, db, 60, "")
+		defer teardown()
+	})
+}
+
+func TestServerReloadPoliciesWithError(t *testing.T) {
+	util.WithDB(func(db kv.DB) {
+		server, teardown := startServer(t, db, 1, "")
+		defer teardown()
+		syscall.Kill(syscall.Getpid(), syscall.SIGUSR2)
+		if server.dir.EpochDeadline() != 1 {
+			t.Fatal("Expect the server's policies not change")
+		}
+		// just to make sure the server's still running normally
+		timer := time.NewTimer(1 * time.Second)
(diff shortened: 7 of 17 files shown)


keyserver/coniksserver/main.go
```diff
@@ -0,0 +1,146 @@
...
+func main() {
+	var genkey boolStringFlag
+	flag.Var(&genkey, "genkey", "Generate signing/VRF key")
+	genConfig := flag.Bool("genconfig", false, "Generate server configurations file")
+	dir := flag.String("dir", ".", "Directory location for storing generated files")
+	confPath := flag.String("config", "config.toml", "Path to the configuration file")
+
+	flag.Parse()
+
+	switch true {
```
arlolra: Can remove the true here.

keyserver/coniksserver/main.go
```diff
@@ -0,0 +1,146 @@
...
+	return sf.value
+}
+
+func (sf *boolStringFlag) IsBoolFlag() bool {
+	return true
+}
+
+func main() {
+	var genkey boolStringFlag
+	flag.Var(&genkey, "genkey", "Generate signing/VRF key")
```
arlolra: The default value, "genkey", doesn't do anything.  Should this be "sign,vrf"?

c633: Does `./main.go -gensignkey -genvrfkey` make sense?

arlolra: Hmm, `boolStringFlag` doesn't seem to be working out the way you wanted it to ... at least, the value always seems to be true, and not a string.

Should probably just make change it to,
```
genkey := flag.String("genkey", "sign,vrf", "Generate signing/VRF key")
```

keyserver/coniksserver/main.go
```diff
@@ -45,10 +45,10 @@ func main() {
 
 	flag.Parse()
 
-	switch true {
+	switch {
 	case *genConfig:
 		mkConfig(*dir)
-		return
+		fallthrough
```
arlolra: `fallthrough` doesn't seem right

keyserver/README.md
```diff
@@ -0,0 +1,43 @@
...
+  -dir string
+    	Directory location for storing generated files (default ".")
+  -genconfig
+    	Generate server configurations file
+  -genkey
+    	Generate signing/VRF key
+```        
+Generate new config file with proper format
+```
+go run keyserver/coniksserver/main.go -genconfig
```
arlolra: These two are already covered in the usage output.

keyserver/coniksserver/main.go
c633: > genkey := flag.String("genkey", "sign,vrf", "Generate signing/VRF key")

But I want `genkey` is an optional flag, and there is no way to check if `genkey` is set or not if we use the default flag types, as I know. 

> boolStringFlag doesn't seem to be working out the way you wanted it to

Yes, that's why I have filtered the input in https://github.com/coniks-sys/coniks-go/blob/e96f8c80ec64b4edd0a02fdc368e2ac07e61f89d/keyserver/coniksserver/main.go#L103-L110.

keyserver/coniksserver/main.go
c633: I think we could do `./main -genconfig -genkey=sign,vrf`. Does it make sense?

keyserver/README.md
```diff
@@ -0,0 +1,43 @@
+# CONIKS Server implementation in Golang
+
+## Usage
+```
+⇒  go install keyserver/coniksserver/main.go
```
arlolra: Having a bin called `main` is not idea.

We could rename main.go, or ask people to either,

```
cd keyserver/coniksserver/
go build
./coniksserver -h
```

or, 

```
go install github.com/coniks-sys/coniks-go/keyserver/coniksserver
coniksserver -h
```

keyserver/coniksserver/main.go
arlolra: No, `fallthrough` just falls through, it doesn't check the second condition.

keyserver/coniksserver/main.go
arlolra: > But I want genkey is an optional flag, and there is no way to check if genkey is set or not if we use the default flag types, as I know.

I see ... I think the way you've documented it now is good. Thanks!

keyserver/coniksserver/main.go
```diff
@@ -0,0 +1,146 @@
...
+func (sf *boolStringFlag) String() string {
+	return sf.value
+}
+
+func (sf *boolStringFlag) IsBoolFlag() bool {
+	return true
+}
+
+func main() {
+	var genkey boolStringFlag
```
arlolra: `boolStringFlag` is no longer used.

keyserver/coniksserver/main.go
c633: Oops :( Thanks!

keyserver/coniksserver/main.go
arlolra: If you want to keep it, go back and look at,

```
keyTypes = strings.Split(keys, ",")
log.Println(keyTypes)
```

keyserver/coniksserver/main.go
c633: I just tried with `coniksserver -genconfig` and it generated the config file only. So I think it worked as expected?

arlolra: Only because genkey doesn't have a default value, right.  The code still gets executed but on the empty string, so nothing happens.

I guess it's ok with the fallthrough then, but maybe leave a comment saying that we're relying on that. Or not.

keyserver/README.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+Generation of self-signed(x509) public key (PEM-encodings `.pem`) based on the private (`.key`)
+```
+openssl req -new -x509 -sha256 -key server.key -out server.pem -days 3650
+```
+
+Run the server
+```
+coniksserver -config="/path/to/config.toml"
```
arlolra: I wonder if this should be a path to `dir`?

keyserver/coniksserver/coniksserver.go
```diff
@@ -0,0 +1,148 @@
...
+	genConfig := flag.Bool("genconfig", false, "Generate server configurations file")
+	dir := flag.String("dir", ".", "Directory location for storing generated files")
+	confPath := flag.String("config", "config.toml", "Path to the configuration file")
+
+	flag.Parse()
+
+	switch {
+	case *genConfig:
+		mkConfig(*dir)
+		fallthrough
```
arlolra:
```
// We fallthrough to allow passing both flags,
// and rely on the default for genkey being the empty string.
```

keyserver/coniksserver/main.go
c633: Yup, I got it. Thanks!

keyserver/coniksserver/coniksserver.go
c633: Hmm, I have assigned a default value for `genkey` (https://github.com/coniks-sys/coniks-go/blob/cdaf4bf18b7f801149e6766f7239a301ec0c2365/keyserver/coniksserver/coniksserver.go#L42) and changed the switch a little bit. I think it's ok but not the best solution :(.

keyserver/README.md
arlolra: I know, I was just thinking about,

```
cd keyserver/coniksserver/
mkdir config
./coniksserver -dir config -genconfig -genkey "sign,vrf"
./conkisserver -dir config   // to run
```

But, ya, nevermind.

c633: Yeah, sounds good!

keyserver/coniksserver/coniksserver.go
```diff
@@ -0,0 +1,153 @@
...
+
+	switch {
+	case *genConfig && genkey.set:
+		mkConfig(*dir)
+		mkKey(genkey.value, *dir)
+		return
+	case *genConfig:
+		mkConfig(*dir)
+		return
+	case genkey.set:
```
arlolra:
```
λ (registration *) ./coniksserver -genkey "sign"
2016/08/15 11:58:06 {true sign,vrf}
```

`boolStringFlag` still doesn't seem to be working.

keyserver/server.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func LoadServerConfig(path string) (*ServerConfig, error) {
+	var conf ServerConfig
+	if _, err := toml.DecodeFile(path, &conf); err != nil {
+		return nil, fmt.Errorf("Failed to load config: %v", err)
+	}
+
+	// load signing key
+	signKey, err := ioutil.ReadFile(conf.Policies.SignKeyPath)
```
arlolra: `conf.Policies.SignKeyPath` needs to be resolved relative to the `path`.

keyserver/coniksserver/coniksserver.go
c633: `boolStringFlag` is a bool type flag, so its syntax is `-genkey=sign,vrf`. The reason is I want to have a call like `./coniksserver -genkey` with default value. I'm not sure if we want to change it to string type flag.

keyserver/server.go
c633: > conf.Policies.SignKeyPath needs to be resolved relative to the path.

Since we also have a path for the key-value db, and I think it shouldn't be a relative to the `-dir` path, so `conf.Policies.SignKeyPath` and `conf.Policies.VRFKeyPath` should be a relative path to the binary. What do you think?

arlolra: I was thinking,

```
signPath := conf.Policies.SignKeyPath
if !filepath.IsAbs(signPath) {
	signPath = filepath.Join(filepath.Dir(path), signPath)
}
```

> Since we also have a path for the key-value db

Can you elaborate?  I think it should be relative to `config.toml`, since that's where they're written.  And `config.toml` is expected to be in `dir` now, no?

c633: >Can you elaborate? 

If it is (e.g.) `-dir config`, I don't think that we also want to save the db to the `config` directory.

> And config.toml is expected to be in dir now, no?

Yes, it is.

keyserver/coniksserver/coniksserver.go
arlolra: Ok, but right now it doesn't support generating the keys individually.  So there's no point in doing all the string split orchestration.

keyserver/server.go
arlolra: > I don't think that we also want to save the db to the config directory.

Fair enough.  How about resolving the db relative to the bin, and the keys relative to the config dir?  Otherwise, `-dir` should not apply to *generating* the keys.

c633: >How about resolving the db relative to the bin, and the keys relative to the config dir? 

Yup, it makes sense.

keyserver/coniksserver/coniksserver.go
c633: Did you try `-genkey=sign`. I think it should work.

arlolra: > Did you try -genkey=sign. I think it should work.

Well that's embarrassing.  Yup, that works.  You've even documented it correctly.  Although, it's slightly unintuitive that `./coniksserver -genkey "sign"` doesn't work, but at least I see why now.

Ok, sorry for being slow here.  What you have is fine.

keyserver/server.go
arlolra: How does that reconcile with,

> Otherwise, -dir should not apply to generating the keys.

I'm going to push back and say that everything (keys, config, etc.) should be relative to `dir` and that the default for `dir` is `.`

That gives you this,

```
mkdir coniks/
coniksserver -dir coniks -genkey -genconfig
coniksserver -dir coniks  // run
```

but it also gives you,

```
coniksserver -genkey -genconfig
coniksserver
```

which seems to be what you want (the default being the current working directory, and not the path to the bin itself).

With the code I supplied above, you can always put an absolute path in the `config.toml` for various files.

c633: Yup, I have had some confusions here. Should go to bed :( Thanks for your suggestions. I will do it!

arlolra: Go to bed!

I'll leave some more comments on the rest of the code for you to pickup tomorrow.

keyserver/server.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Cannot read signing key: %v", err)
+	}
+	if len(signKey) != sign.PrivateKeySize {
+		return nil, fmt.Errorf("Signing key must be 64 bytes (got %d)", len(signKey))
+	}
+
+	// load VRF key
+	vrfPath := conf.Policies.VRFKeyPath
+	if !filepath.IsAbs(vrfPath) {
```
arlolra: Maybe a helper for this boilerplate?

```
func resolvePath(file, other string) other string {
  if !filename.IsAbs(other) {
    file = filepath.Join(filepath.Dir(file), other)
  }
} 

vrfPath := util.resolvePath(file, conf.Policies.VRFKeyPath)
```

keyserver/README.md
```diff
@@ -0,0 +1,35 @@
+# CONIKS Server implementation in Golang
+
+## Usage
+```
+⇒  go install github.com/coniks-sys/coniks-go/keyserver/coniksserver
+⇒  coniksserver -h
+Usage of coniksserver:
+  -config string
+    	Path to the configuration file (default "config.toml")
```
arlolra: We're sort of obsoleted `-config` now with `-dir`

keyserver/server_test.go
```diff
@@ -0,0 +1,235 @@
...
+		},
+		Policies: &ServerPolicies{
+			EpochDeadline: epDeadline,
+			vrfKey:        vrfKey,
+			signKey:       signKey,
+		},
+	}
+
+	server := NewConiksServer(conf)
+	server.db = kvdb
```
arlolra: Maybe `NewConiksServer` could take an optional `db`?

keyserver/server.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	// server registration port
+	botLn, err := tls.Listen("tcp", tc.LocalAddress, config)
+	if err != nil {
+		panic(err)
+	}
+
+	server.waitStop.Add(1)
+	go server.listenForRequests(clientLn, server.handleClientMessage)
```
arlolra: Any reason not to prefer a pattern like?

```
go func() {
  server.waitStop.Add(1)
  server.listenForRequests(clientLn, server.handleClientMessage)
  server.waitStop.Done()
}
```

It's a little clearer to see that the `Done()` eventually gets called.

c633: Great! Thanks!

keyserver/README.md
c633: I'd like to have an option to specify the config file name, so that the developers don't have to use our default file name.

keyserver/server_test.go
c633: Maybe it's unnecessary, since this's just a modification for testing.

keyserver/README.md
```diff
@@ -0,0 +1,35 @@
...
+Generation of self-signed(x509) public key (PEM-encodings `.pem`) based on the private (`.key`)
+```
+⇒  openssl req -new -x509 -sha256 -key server.key -out server.pem -days 3650
+```
+
+Run the server
+```
+⇒  mkdir coniks/
+⇒  coniksserver -dir coniks -genkey -genconfig
+⇒  coniksserver -dir coniks  // run
```
Liamsi: Here is yet another (late) proposal: Wouldn't it be nice if for instance `coniksserver init` created all necessary files/generates all keys in the current working directory (or in a default directory). Then above commands could boil down to:
```
 ⇒  mkdir coniks; cd coniks
 ⇒  coniksserver init // creates all files, maybe even the self-signed ssl keys/cert with sane defaults
 ⇒  coniksserver run // looks for config files wit default names in the current directory if not specified differently
``` 
Just another idea. I like how for instance `git` or `go` use "subcommands" to make the commandline tools better structured and hence easier to use. But I also like what we have now.

c633: Interesting! Maybe I'll go with `cobra` package.

Liamsi: Cool, that was quick. [cobra](https://github.com/spf13/cobra) looks indeed very nice. I'm also slightly more familiar with [cli](https://github.com/urfave/cli). Never compared both. BUt I guess any of those equally suit our needs.

keyserver/coniksserver/cmd/init.go
```diff
@@ -0,0 +1,105 @@
...
+	Long:  `Create a configuration file and generate all keys for signing and VRF`,
+	Run: func(cmd *cobra.Command, args []string) {
+		dir := cmd.Flag("dir").Value.String()
+		mkConfig(dir)
+		mkSigningKey(dir)
+		mkVrfKey(dir)
+
+		cert, err := strconv.ParseBool(cmd.Flag("cert").Value.String())
+		if err == nil && cert {
+			testutil.CreateTLSCert(dir)
```
Liamsi: If `CreateTLSCert` won't be used only in tests anymore we should move it from `testutil` into a general `util` package.

c633: Yes, I was also thinking about it. But `CreateTLSCert` should be used for development environment only (i.e., `go test` and local deployment), so I'm not sure moving it into `util` makes sense.

keyserver/coniksserver/cmd/run.go
```diff
@@ -0,0 +1,50 @@
...
+	Long: `Run a CONIKS server instance
+
+This will look for config files with default names (config.toml) 
+in the current directory if not specified differently.
+	`,
+	Run: func(cmd *cobra.Command, args []string) {
+		dir := cmd.Flag("dir").Value.String()
+		config := cmd.Flag("config").Value.String()
+
+		confPath := path.Join(dir, config)
```
arlolra: I see you changed `config` to just a filename now.  Maybe leaving it as a path is ok, and checking for absolute here?

c633: > Maybe leaving it as a path is ok,

Yeah, it makes sense.
> checking for absolute here?

Why do we need this check?

arlolra: I was thinking you'd leave `dir` around, but, yeah, since everything is relative to the `config.toml` now, we're good.  Thanks.

protocol/message.go
```diff
@@ -35,13 +35,18 @@ func NewErrorResponse(e ErrorCode) Response {
...
 
 func NewRegistrationProof(ap *m.AuthenticationPath, str *m.SignedTreeRoot,
-	tb *m.TemporaryBinding) *DirectoryProof {
-	return &DirectoryProof{RegistrationType, ap, str, tb}
+	tb *m.TemporaryBinding, err ErrorCode) *DirectoryProof {
+	df := &DirectoryProof{RegistrationType, ap, str, tb, ""}
+	if err != Success {
+		df.Error = err.Error().Error()
+	}
+	return df
```
arlolra: Maybe `return df, err` to reduce some redundancy above.

c633: The purpose of `Error` field in `DirectoryProof` is discussed here https://github.com/coniks-sys/coniks-go/pull/29#discussion_r72494892.

arlolra: I'm saying that above, you're doing,

``
return NewRegistrationProof(ap, d.LatestSTR(), nil, Success), Success
```

but if you `return df, err` here, you can do,

```
return NewRegistrationProof(ap, d.LatestSTR(), nil, Success)
```

Comment (c633) [inline protocol/message.go]: Ahh, I see. Sorry for my bad :(

File: keyserver/handler.go [hunk 19]
```diff
@@ -0,0 +1,15 @@
+package keyserver
+
+import . "github.com/coniks-sys/coniks-go/protocol"
+
+func (server *ConiksServer) handleRegistration(reg *RegistrationRequest) (Response, error) {
+	if len(reg.Username) == 0 || len(reg.Key) == 0 {
```
arlolra: This entire file seems redundant with `HandleOps`.  Shouldn't you be using that instead?

c633: Yes, I still wonder if whether we want to perform some extra tasks in the handles in the future. Otherwise, it's completely redundant.

c633: @masomel Do you think we will have some other tasks, related to the server implementation, in these handles?

keyserver/listener.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// handle request
+	switch req.Type {
+	default:
+		log.Printf("unknown message type: %q", req.Type)
+		return malformedClientMsg(), ErrorMalformedClientMessage.Error()
+	}
+}
+
+func (server *ConiksServer) handleBotMessage(msg []byte) ([]byte, error) {
```
arlolra: What do you think of this ... combine `handleBotMessage` and `handleClientMessage` into one func, but you make another function that binds the `req.Type`s it accepts.

```
func makeHandler(acceptableTypes []int)  func(msg []byte) ([]byte, error) {
  return func(msg []byte) { ... }
}
```

c633: Is that what you mean? cfe5e95b9cc2bd0403f69f6313d65fec1bb35319

keyserver/listener.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (server *ConiksServer) makeHandler(acceptableTypes map[int]bool) func(msg []byte) ([]byte, error) {
+	return func(msg []byte) ([]byte, error) {
+		// get request message
+		req, err := UnmarshalRequest(msg)
+		if err != nil {
+			return malformedClientMsg(), err
+		}
+
```
arlolra: Maybe hoist the type check,

```
if !acceptableTypes[req.Type] {
  log.Printf("unacceptable message type: %q", req.Type)
  return malformedClientMsg(), ErrorMalformedClientMessage.Error()
}
```

Then you can keep the switch on req.Type and the default error.

keyserver/listener.go
arlolra: Yup, thanks!

keyserver/listener.go
```diff
@@ -0,0 +1,115 @@
...
+		switch true {
+		case req.Type == RegistrationType && acceptableTypes[RegistrationType]:
+			response, e := server.dir.HandleOps(req)
```
arlolra: Didn't you have `server.Lock()` here before?

keyserver/listener.go
```diff
@@ -0,0 +1,126 @@
...
+		req, err := UnmarshalRequest(msg)
+		if err != nil {
+			return malformedClientMsg(err)
+		}
+		if !acceptableTypes[req.Type] {
+			log.Printf("unacceptable message type: %q", req.Type)
+			return malformedClientMsg(ErrorMalformedClientMessage.Error())
+		}
+
+		switch true {
```
arlolra: Is this switch even necessary?  `HandleOps` has one already :)

keyserver/listener.go
c633: Yeah, I was thinking that I can simplify the handle to 
```
if !acceptableTypes[req.Type] {
			log.Printf("unacceptable message type: %q", req.Type)
			return malformedClientMsg(ErrorMalformedClientMessage.Error())
		}
		response, e := server.dir.HandleOps(req)
		res, err := json.Marshal(response)
		if err != nil {
			panic(e)
		}
		return res, e.Error()
```
 It's sad :(

keyserver/listener.go
c633: I think yes (https://github.com/coniks-sys/coniks-go/pull/71#discussion_r74984970)

keyserver/listener.go
arlolra: Yes, that's the body that I'd like to see.

> In case of KeyLookup and KeyLookupInEpoch we probably just need to use server.RLock() instead of server.Lock(). It's sad :(

I think adding switches here to apply locks might be ok.

```
switch req.Type {
  default:
    server.Lock()
}
response, e := server.dir.HandleOps(req)
switch req.Type {
  default:
    server.UnLock()
}
```

Either that, or you move the mutex to the Directory, since that has knowledge of what actually needs the locking (or, even the Merkletree).

c633: Wow, such a brilliant idea.

keyserver/listener.go
```diff
@@ -0,0 +1,127 @@
...
+	"io"
+	"log"
+	"net"
+	"time"
+
+	. "github.com/coniks-sys/coniks-go/protocol"
+)
+
+func (server *ConiksServer) listenForRequests(ln net.Listener, handler func(msg []byte) ([]byte, error)) {
+	defer ln.Close()
```
arlolra: `defer ln.Close()` isn't necessary since `<-server.stop` is the only way to return from this function.

c633: Right! Thanks!

keyserver/listener.go
```diff
@@ -0,0 +1,127 @@
...
+		if err != nil {
+			select {
+			case <-server.stop:
+				return
+			default:
+				log.Printf("accept client: %s", err)
+				continue
+			}
+		}
+		server.waitStop.Add(1)
```
arlolra: Is this wait necessary, since we're already waiting on `ln.Close()`?  In other words, will any response get through after a call to `ln.Close()`?

keyserver/listener.go
```diff
@@ -0,0 +1,127 @@
...
+	"encoding/json"
+	"io"
+	"log"
+	"net"
+	"time"
+
+	. "github.com/coniks-sys/coniks-go/protocol"
+)
+
+func (server *ConiksServer) listenForRequests(ln net.Listener, handler func(msg []byte) ([]byte, error)) {
```
arlolra: I think this would all be clearer as,

```
defer ln.Close()
for {
  switch {
    <-server.stop
    return
  }
  conn, err := ln.Accept()
  if err != nil {
    log.Printf("accept client: %s", err)
    continue
  }
  ...
}
```

arlolra: You can then add a second wait group just for the accepted requests and call Wait on that group after <-server.stop so that they all get handled cleanly, then return to close the connection.

keyserver/listener.go
arlolra: The difference between the above suggestion and here is that you're calling close before sending a response to the already accepted clients.  I don't know enough about internal of `Close` though.

keyserver/listener.go
c633: I still think that `<-server.stop` should be handled in a go routine. And your suggestion on a second wait group for the accepted requests seems a safe/good way to go.

keyserver/listener.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func (server *ConiksServer) acceptClient(conn net.Conn, handler func(msg []byte) ([]byte, error)) {
+	defer conn.Close()
+	closed := make(chan struct{})
+	defer close(closed)
+	go func() {
+		select {
+		case <-server.stop:
+			conn.Close()
```
arlolra: Oh, hmm, yeah, I think we should be a little nicer and give the whole 5 seconds for this connection to finish before shutting down.

With the above suggestions,  you can remove all the references to <-server.stop from this method.

keyserver/listener.go
arlolra: > I still think that <-server.stop should be handled in a go routine.

Can you explain why?

c633: I'm sorry. It shouldn't, it just my personal preference.

arlolra: So like this?

```
var stopServer bool
go func() {
  <-server.stop
  stopServer = true
}
for {
  if stopServer {
    return
  }
  ...
```

c633: I didn't notice that the you removed `<-server.stop` from `ln.Accept()`. Sorry for the noise.
Is this your suggestion? 2e182419b8b77ece38e599f4cc4a49b2c04dbf70

arlolra: Yup, exactly.  Thanks!

arlolra:
Ok, this LGTM now.  I'm going to give some time for others before cavalierly merging.  Thanks Huy!

c633:
Thanks for all your comments/suggestions! They're so great!

keyserver/listener.go
```diff
@@ -0,0 +1,106 @@
...
+		default:
+			server.Lock()
+		}
+		response, e := server.dir.HandleOps(req)
+		switch req.Type {
+		default:
+			server.Unlock()
+		}
+
+		res, err := json.Marshal(response)
```
arlolra: Sorry, one last thing, move this to a function in encoding.go called MarshallRequest, so that all json calls are isolated there.

c633: Yes.

keyserver/listener.go
```diff
@@ -0,0 +1,106 @@
...
+
+	for {
+		select {
+		case <-server.stop:
+			server.waitCloseConn.Wait()
+			return
+		default:
+		}
+
+		conn, err := ln.Accept()
```
arlolra: Oh, hmm, the loop is stuck waiting on this Accept now :(

Thinking ...

c633: A work around for a TCP connection is 
```
    go func() {
		<-server.stop
		ln.(*net.TCPListener).SetDeadline(time.Now())
	}()
```
but we're using TLS, so ... I'm also thinking about it.

arlolra: Let's set a deadline,

```
ln.SetDeadline(time.Now().Add(5 * time.Second))
```

then check the error for a timeout.

c633: There is no `SetDeadline` for `tls.listener`...

c633: What do you think about https://github.com/coniks-sys/coniks-go/blob/4ab5e299ac9f02d350c891cbf30a120362243fbd/keyserver/listener.go#L14-L30?

keyserver/listener.go
```diff
@@ -0,0 +1,109 @@
...
+func (server *ConiksServer) listenForRequests(ln net.Listener, handler func(msg []byte) ([]byte, error)) {
+	defer ln.Close()
+	go func() {
+		<-server.stop
+		server.waitCloseConn.Wait()
+		ln.Close()
+	}()
+
+	for {
+		conn, err := ln.Accept()
```
arlolra: We can use a tcp listener and then call,

```
conn = tls.Server(conn, server.tlsConfig) 
```

keyserver/listener.go
```diff
@@ -0,0 +1,109 @@
...
+		if err != nil {
+			select {
+			case <-server.stop:
+				return
```
arlolra: This return invokes the deferred close though, and even if you remove that, it clears the wait group on return.

keyserver/listener.go
arlolra: Wrote this up quickly: 7f42aa13f189d30b9d3cafd43da76ae33be70fe1

c633: Thanks! I merged it in c6fcdf492c062c9563d846a713a41c6052b699f5.

keyserver/listener.go
```diff
@@ -0,0 +1,110 @@
...
+	for {
+		select {
+		case <-server.stop:
+			server.waitCloseConn.Wait()
+			return
+		default:
+		}
+		ln.SetDeadline(time.Now().Add(5 * time.Second))
+		conn, err := ln.Accept()
+		conn = tls.Server(conn, server.tlsConfig)
```
arlolra: Sorry, move this below the err check.

masomel:
I'm still reviewing this pull, and was going through every single commit. Please don't merge this yet.

keyserver/README.md
```diff
@@ -0,0 +1,46 @@
...
+|     II  .7I;. ,I.C KTocoCNS.    | against malicious or compromised
+|    T7  1t,  ;:,A,.I;BCZCZCmW,   | communication providers and surveillance
+|   7I .2:  7T: ,7i:,7mpODUDOBW   | by storing users' encryption keys in
+|   S  Z, ,C;   Zm T.;SNGpGpmB@   | tamper-evident and publicly auditable key
+|   c; Z, D.   oIm;.I:SBmmmKN@2   | directories on the server side.
+|    U. U T   2.,:i Z7@NANB@@B    | This allows messaging clients to verify the
+|     o;,I:0  ;,,,,IGBW@W@W@o     | identity of users automatically, and
+|      ,;;oZp7.;C7E@@W@W@N7       | prevents malicious/compromised servers
+|           ,1o:T@@O7:            | from hijacking secure communications
+.---------------------------------. without getting caught.
```
masomel: What does this image represent?

keyserver/coniksserver/cmd/init.go
```diff
@@ -0,0 +1,105 @@
...
+	if _, err := os.Stat(file); err == nil {
+		log.Printf("%s already exists\n", file)
+		return
+	}
+	sk, err := sign.GenerateKey(nil)
+	if err != nil {
+		log.Print(err)
+		return
+	}
+	if err := ioutil.WriteFile(file, sk[:], 0600); err != nil {
```
masomel: Do we think file mode 0600 will be enough to protect the keys?

keyserver/server.go
```diff
@@ -0,0 +1,224 @@
...
+	localTypes[protocol.RegistrationType] = true
+	server.waitStop.Add(1)
+	go func() {
+		server.listenForRequests(localLn, server.makeHandler(localTypes))
+		server.waitStop.Done()
+	}()
+
+	server.waitStop.Add(1)
+	go func() {
+		server.updatePolicies()
```
masomel: Sorry if I missed an earlier discussion on this, but it seems odd to me that ``updatePolicies()`` would be part of `server.Run()`.

keyserver/coniksserver/cmd/init.go
masomel: I should clarify that I'm not opposed to this for now, but I do wonder if it'll be sufficient on the long run.

arlolra:
> was going through every single commit.

Sorry for clearing the history, I can restore it if you want :(

keyserver/handler.go
masomel: Sorry I missed this when I was checking through the comments earlier (so many comments...). It seems you decided to get rid of `handler.go`, which I think is fine. The only server-related tasks I can think of right now that we may need to perform in the handlers are possibly reading from the DB. But even those, I think, may be abstracted away in the directory, no?

masomel:
Don't worry about it :) It's much easier to review a single commit and I had only made one small comment (which I've added back in).

arlolra:
Phew

keyserver/README.md
arlolra: I think it's an ascii onion.  But, yeah, we should probably stick to CONIKS branding.

keyserver/server.go
arlolra: Can you say a little more?  I don't think we've talked about this yet.

keyserver/listener.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+		response, e := server.dir.HandleOps(req)
+		switch req.Type {
+		default:
+			server.Unlock()
+		}
+
+		res, err := MarshalResponse(response)
+		if err != nil {
+			panic(e)
```
masomel: I think the panic should be on `err` not `e`, no? Otherwise, if the registration request resulted in an error as well, we'll end up panicking on that instead.

keyserver/listener.go
```diff
@@ -0,0 +1,108 @@
...
+		n, err := conn.Read(buf)
+		if err != nil {
+			if err != io.EOF {
+				log.Printf("client read %v: %v", conn.RemoteAddr(), err)
+			}
+			return
+		}
+
+		res, err := handler(buf[:n])
+		if err != nil {
```
masomel: Much like in https://github.com/coniks-sys/coniks-go/pull/71/commits/89cfe98c5ee06764e6855cdfbbc148a3d2f6f3ee#r75038395, we'll end up catching errors like `ErrorNameExisted', right? Do we want to be logging these?

protocol/directory.go
```diff
@@ -70,13 +74,13 @@ func (d *ConiksDirectory) Register(req *RegistrationRequest) (
 		return nil, ErrorDirectory
```
masomel: Shouldn't this return an `ErrorResponse` along with the error?

keyserver/listener.go
```diff
@@ -0,0 +1,108 @@
...
+		default:
+			server.Lock()
+		}
+		response, e := server.dir.HandleOps(req)
+		switch req.Type {
+		default:
+			server.Unlock()
+		}
+
+		res, err := MarshalResponse(response)
```
masomel: What'll happen here if `response == nil`, as would be the case when a `ErrorDirectory` occurs in `directory.Register()`? Also see https://github.com/coniks-sys/coniks-go/pull/71/commits/89cfe98c5ee06764e6855cdfbbc148a3d2f6f3ee#r75040455

keyserver/listener.go
masomel: After going through the flow of operations, I think ``handler`` should only return an error if an internal error occurred.

protocol/directory.go
```diff
@@ -70,13 +74,13 @@ func (d *ConiksDirectory) Register(req *RegistrationRequest) (
...
 	}
 	if bytes.Equal(ap.LookupIndex, ap.Leaf.Index) {
-		return NewRegistrationProof(ap, d.LatestSTR(), nil), ErrorNameExisted
+		return NewRegistrationProof(ap, d.LatestSTR(), nil, ErrorNameExisted)
```
masomel: As I said in https://github.com/coniks-sys/coniks-go/pull/71/commits/89cfe98c5ee06764e6855cdfbbc148a3d2f6f3ee#r75041201, I don't think we should be returning the error in this case since ``ErrorNameExisted`` will be propagated all the way up `listener.acceptClient()`, which I think should only be concerned about internal server errors (and maybe panic on those).

keyserver/listener.go
arlolra: `panic` seems very wrong here.

keyserver/README.md
c633: Yes, it's Tor logo ascii. I hope you don't get mad at this :(

keyserver/coniksserver/cmd/init.go
c633: I'm not so sure but the permission of private keys in `.ssh` is also `600`. @arlolra : Will we have some better secure key management in production stage?

keyserver/listener.go
c633: > I think the panic should be on err not e, no?

Yes, it should panic on `err`. 

> panic seems very wrong here.

I think if there is any error here, it must be our bug?

keyserver/listener.go
c633: > Do we want to be logging these?

I think we should log these errors, too. There must have something wrong if there is a `ErrorNameExisted` in our current implementation, I think.

protocol/directory.go
c633: Yes, it should be handled in `HandleOps`. I fixed it in 5c4c4b02c0284c33a4dfd3440576c1d65cb7c823.

protocol/directory.go
c633: Maybe using the prefix `Error` here is not right choice but all our predefined `ErrorXxx`s are used for logging and returning to the client only, that is `ErrorNameExisted` and `ErrorDirectory` are treated as the same.

keyserver/server.go
masomel: Sure! I was under the impression that any server policy update was really a configuration update for which the server needs to be shut down and restarted, so I'm not sure I know how/when it makes sense to change the server's policies online. But I could just be missing something about how ``updatePolicies()`` works...

keyserver/listener.go
masomel: > I think if there is any error here, it must be our bug? 

Not necessarily a bug? I'm not sure. But definitely internal server error (e.g. `ErrorDirectory`)

> `panic` seems very wrong here.

@arlolra do you think we shouldn't be panicking here?

keyserver/server.go
c633: I think I can explain a little bit here. The server's `reloadChan` will listen for a signal from the user (`kill -USR2 pid`), so what `updatePolicies()` does is to receive this signal and re-read the policies from the config file.

keyserver/listener.go
c633: > But definitely internal server error (e.g. ErrorDirectory)

No, it is `panic(err)` which is an error returned by `json.Marshal`. That's why I said it's maybe our bug.

keyserver/listener.go
masomel: > There must have something wrong if there is a ErrorNameExisted in our current implementation, I think.

Logging these errors makes sense, but I don't think it's necessarily true that `ErrorNameExisted` is a problem in our implementation. It's completely possible for one user to try to register a name that already exists in the directory.

OTOH, if we support multiple devices/application instances, we need to handle the case in which a second device registers a second key for an existing name (this is related to #17).

c633: Yup, you're right. I typed faster than my thought. I still think that it makes sense to let `handler` return all the errors up to `listener.Accept()`(for logging purpose).

keyserver/listener.go
masomel: > No, it is panic(err) which is an error returned by json.Marshal. That's why I said it's maybe our bug. 

Oh yeah you're right, it's definitely a bug in this case because the `response` might be malformed. My previous response more for the general case.

keyserver/server.go
masomel: Ah! Ok, I did not get that. I'm still trying to wrap my head around the `waitStop` and the go func stuff. This makes much more sense now. @c633 thanks for the explanation! This means that that the server admin can change the config file while the server is running and then just `kill -USR2 pid` to update it? That's really cool! This should be specified in the README.

c633: > This means that that the server admin can change the config file while the server is running and then just kill -USR2 pid to update it?

Yes, it is. I updated the README.

keyserver/listener.go
c633: @arlolra : What do you think about:

```
--- a/keyserver/listener.go
+++ b/keyserver/listener.go

 func (server *ConiksServer) listenForRequests(ln *net.TCPListener, handler func(msg []byte) ([]byte, error)) {
 	defer ln.Close()
+	go func() {
+		<-server.stop
+		ln.SetDeadline(time.Now())
+	}()
 	for {
 		select {
 		case <-server.stop:
			server.waitCloseConn.Wait()
			return
		default:
 		}
-		ln.SetDeadline(time.Now().Add(5 * time.Second))
 		conn, err := ln.Accept()
 		if err != nil {
 			if opErr, ok := err.(*net.OpError); ok && opErr.Timeout() {

```

arlolra: Looks good

keyserver/README.md
```diff
@@ -0,0 +1,56 @@
...
+|   rri:iM995HM@MBMMMBMMMBMBMBB@BM5SS9M5,iis,  |
+|   :s;i:rMGX5@BMMMM@B@B@B@@@B@MX2SX9MX::irr   |
+|    :ri::iSG9B@BBM@GGGMMMGG9X222X9M9s,:irs    |
+|     riii:,r2@B@B@B9252525252SS99Xs:,irsr     |
+|    ;riiiii:,:s5MB@M9GGGGXXSXS9H:,::rrs:      |
+|    sri;irrrii,,,::iirrriX9XS9GS,;rsr:        |
+|   :srrrsrr:irrrrii:i:::,9MGMM@Mii:           |
+|   2sr::,      ,:iirrrrriMMGSsr:              |
+|  ,,                                          |
+\----------------------------------------------/
```
arlolra: That's Tor Messenger specific.  We should keep this to CONIKS only.  Since there's no logo, what if you just did "CONIKS" in ascii art?

Pick your favourite http://patorjk.com/software/taag/#p=display&f=Rectangles&t=CONIKS

```
                                     
 _____ _____ _____ _____ _____ _____ 
|     |     |   | |     |  |  |   __|
|   --|  |  | | | |-   -|    -|__   |
|_____|_____|_|___|_____|__|__|_____|
                                     
```

c633: Yup, I probably should remove it :D. Sorry for the noise :D

keyserver/README.md
```diff
@@ -0,0 +1,56 @@
...
+```
+⇒  mkdir coniks; cd coniks
+⇒  coniksserver init -c // creates all files including a self-signed ssl keys/cert
+⇒  coniksserver run     // run
+```
+
+You can reload the server's policies while it's running by editing the `config.toml` file
+and possibly replace `vrf.priv` with a new key, then run
+```
+⇒  kill -USR2 pid
```
arlolra: It might be nice if `coniksserver run` took a flag to spit out the `pid` in a `coniks.pid`, or something.

keyserver/coniksserver/cmd/init.go
arlolra: Probably, but I'm not sure about the specifics.

We should maybe just ensure the key is only required for reading at process launch, and then can be safely removed from its location.

keyserver/listener.go
arlolra: Sorry, I hastily through in the above comment.  Don't mind me here.

> it's definitely a bug in this case because the response might be malformed

Sounds right.

Thanks for digging into the error propagation so thoroughly.

keyserver/README.md
c633: Maybe I thought that' not a big deal, so I gave it a try in d144f97cbe7f0ea300954560108ddea0ff5f9427.

arlolra: Thanks, but where's the ascii CONIKS :(

protocol/directory.go
masomel: Hmm ok. It seemed to me like `ErrorDirectory` happens when the `pad.Lookup` returns an error, right, which would indicate an internal server error, right? We should definitely return an error about this to the client, but I think we should treat such an error differently on the server side.

keyserver/listener.go
masomel: Ok, that's fair. But I think the listener should additionally check for the type of error and alert the server admin about internal errors.

keyserver/README.md
```diff
@@ -42,14 +28,14 @@ Use "coniksserver [command] --help" for more information about a command.
 Run the server
 ```
 ⇒  mkdir coniks; cd coniks
-⇒  coniksserver init -c // creates all files including a self-signed ssl keys/cert
-⇒  coniksserver run     // run
+⇒  coniksserver init -c # creates all files including a self-signed ssl keys/cert
+⇒  coniksserver run -p  # run & write down the process ID into coniks.pid
```
masomel: Nice!

protocol/directory.go
c633: You're right. That's my intention for #45.

masomel: Yeah, this definitely makes sense.

masomel:
LGTM :)

PR #72, opened 2016-08-15
Implement protocol key lookup & monitoring function

### LINKED ISSUE #70: Key lookup & Consistency check
Continue the discussion at https://github.com/coniks-sys/coniks-go/pull/64#discussion_r74705314

c633:
* Should resolve #70

### DIFF
--- a/keyserver/server_test.go
+++ b/keyserver/server_test.go
@@ -2,6 +2,7 @@ package keyserver
 
 import (
 	"bytes"
+	"encoding/base64"
 	"encoding/json"
 	"path"
 	"syscall"
@@ -16,6 +17,27 @@ import (
 	"github.com/coniks-sys/coniks-go/storage/kv"
 	"github.com/coniks-sys/coniks-go/utils"
 )
+
+var registrationMsg = `
+{
+    "type": 0,
+    "request": {
+        "username": "alice@twitter",
+        "key": [0,1,2],
+        "allow_unsigned_key_change": true,
+        "allow_public_lookup": true
+    }
+}
+`
+
+var keylookupMsg = `
+{
+    "type": 1,
+    "request": {
+        "username": "alice@twitter"
+    }
+}
+`
 
 func startServer(t *testing.T, kvdb kv.DB, epDeadline merkletree.TimeStamp, policiesPath string) (*ConiksServer, func()) {
 	dir, teardown := testutil.CreateTLSCertForTest(t)
@@ -80,18 +102,8 @@ func TestBotSendsRegistration(t *testing.T) {
 	util.WithDB(func(db kv.DB) {
 		_, teardown := startServer(t, db, 60, "")
 		defer teardown()
-		reg := `
-{
-	"type": 0,
-	"request": {
-		"username": "alice@twitter",
-		"key": [0,1,2],
-        "allow_unsigned_key_change": true,
-        "allow_public_lookup": true
-	}
-}
-`
-		rev, err := testutil.NewClient(t, testutil.LocalConnection, []byte(reg))
+
+		rev, err := testutil.NewClient(t, testutil.LocalConnection, []byte(registrationMsg))
 		if err != nil {
 			t.Fatal(err)
 		}
@@ -120,18 +132,8 @@ func TestSendsRegistrationFromOutside(t *testing.T) {
 	util.WithDB(func(db kv.DB) {
 		_, teardown := startServer(t, db, 60, "")
 		defer teardown()
-		reg := `
-{
-	"type": 0,
-	"request": {
-		"username": "alice@twitter",
-		"key": [0,1,2],
-        "allow_unsigned_key_change": true,
-        "allow_public_lookup": true
-	}
-}
-`
-		rev, err := testutil.NewClient(t, testutil.PublicConnection, []byte(reg))
+
+		rev, err := testutil.NewClient(t, testutil.PublicConnection, []byte(registrationMsg))
 		if err != nil {
 			t.Fatal(err)
 		}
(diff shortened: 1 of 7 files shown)


keyserver/server_test.go
```diff
@@ -237,3 +238,249 @@ func TestRegisterDuplicateUserInDifferentEpoches(t *testing.T) {
...
+		if response.AP == nil {
+			t.Fatal("Expect a proof of inclusion")
+		}
+		if response.TB != nil {
+			t.Fatal("Expect returned TB is nil")
+		}
+	})
+}
+
+func TestKeyLookupInEpoch(t *testing.T) {
```
c633: This test failed when running in my machine. The reason is `conn.Read()` in the client side didn't read all written bytes. My current `go` version is `go version go1.7 darwin/amd64`. Do you have any idea why this happened? @arlolra @Liamsi

Liamsi: In go 1.7 the tls behaviour changed: https://golang.org/doc/go1.7#minor_library_changes (scroll to crypto/tls)
I experimented with `DynamicRecordSizingDisabled` to enforce the "old" 1.6 behaviour but the `Read` in `testutil.NewClient` still doesn't read enough data. In general it would be right to not rely on the on the underlying implementation of `net.Conn` and to make sure that `Read` and `Write` have received and written the expected amount of bytes.

c633: Wow, thanks!

> I experimented with DynamicRecordSizingDisabled to enforce the "old" 1.6 behaviour

Where did you set this field? I set this field to true in the server's tls config, and `Read` in `testutil.NewClient` was able to read all the buffer.

protocol/directory.go
```diff
@@ -102,3 +121,66 @@ func (d *ConiksDirectory) Register(req *RegistrationRequest) (
 		return NewRegistrationProof(ap, d.LatestSTR(), nil, Success)
 	}
 }
+
+func (d *ConiksDirectory) KeyLookup(req *KeyLookupRequest) (
+	*DirectoryProof, ErrorCode) {
+	ap, err := d.pad.Lookup(req.Username)
+	if err != nil {
+		return nil, ErrorDirectory
```
masomel: I think this should return a `NewErrorResponse()`, no?

protocol/directory.go
```diff
@@ -102,3 +121,66 @@ func (d *ConiksDirectory) Register(req *RegistrationRequest) (
...
+
+func (d *ConiksDirectory) KeyLookupInEpoch(req *KeyLookupInEpochRequest) (
+	*DirectoryProofInEpoch, ErrorCode) {
+	var strs []*merkletree.SignedTreeRoot
+	startEp := req.Epoch
+	endEp := d.LatestSTR().Epoch
+
+	ap, err := d.pad.LookupInEpoch(req.Username, startEp)
+	if err != nil {
+		return nil, ErrorDirectory
```
masomel: Same as L129.

protocol/directory.go
```diff
@@ -102,3 +121,66 @@ func (d *ConiksDirectory) Register(req *RegistrationRequest) (
...
+	var aps []*merkletree.AuthenticationPath
+	startEp := req.StartEpoch
+	endEp := req.EndEpoch
+	if endEp > d.LatestSTR().Epoch {
+		endEp = d.LatestSTR().Epoch
+	}
+	for ep := startEp; ep <= endEp; ep++ {
+		ap, err := d.pad.LookupInEpoch(req.Username, ep)
+		if err != nil {
+			return nil, ErrorDirectory
```
masomel: Same as L129 and L152.

protocol/directory_test.go
```diff
@@ -138,3 +139,231 @@ func TestRegisterExistedUserWithoutTB(t *testing.T) {
 		t.Fatal("Expect returned TB is nil")
 	}
 }
+
+func TestKeyLookupWithTB(t *testing.T) {
+	d := newDirectory(t, true)
+	_, err := d.Register(&RegistrationRequest{
```
masomel: I think it might be worth keeping the response message here to make sure the server responds with the same TB for the following lookup, as a sanity check.

protocol/directory_test.go
```diff
@@ -138,3 +139,231 @@ func TestRegisterExistedUserWithoutTB(t *testing.T) {
...
+	}
+	if ap := res.AP; ap == nil || bytes.Equal(ap.LookupIndex, ap.Leaf.Index) {
+		t.Fatal("Expect a proof of absence")
+	}
+	if res.TB != nil {
+		t.Fatal("Expect returned TB is nil")
+	}
+}
+
+func TestKeyLookupWithoutTB(t *testing.T) {
```
masomel: How is this test different from `TestKeyLookupWithTB`? Also, the lookup in same epoch in that test above would also return an `ErrorNameNotFound`, right?

masomel: Ah, nevermind, `useTBs` is set to false here. But I still think that the test above should return the `ErrorNameNotFound`.

protocol/directory_test.go
```diff
@@ -138,3 +139,231 @@ func TestRegisterExistedUserWithoutTB(t *testing.T) {
...
+		t.Fatal("Expect a proof of absence")
+	}
+	if res.TB != nil {
+		t.Fatal("Expect returned TB is nil")
+	}
+
+	d.Update()
+	// lookup in different epochs
+	// expect a proof of absence
+	res, err = d.KeyLookup(&KeyLookupRequest{Username: "bob"})
```
masomel: This part of the test seems redundant to me since it's no different than in the analogous `WithTB` test. But maybe that's exactly what you're testing?

protocol/directory_test.go
```diff
@@ -138,3 +139,231 @@ func TestRegisterExistedUserWithoutTB(t *testing.T) {
...
+	d := newDirectory(t, false)
+	_, err := d.Register(&RegistrationRequest{
+		Username: "alice",
+		Key:      []byte("key")})
+	if err != Success {
+		t.Fatal("Unable to register")
+	}
+
+	d.Update()
+	savedSTR := d.LatestSTR().Signature
```
masomel: It's fine to only save the STR signature here since it's a test, but in the client we should make sure to save a few more STR fields like the latestEpoch and the epochDeadline.

protocol/directory_test.go
```diff
@@ -138,3 +139,231 @@ func TestRegisterExistedUserWithoutTB(t *testing.T) {
...
+	for i := 2; i < N; i++ {
+		d.Update()
+	}
+
+	// missed from epoch 2
+	res, err := d.ConsistencyCheck(&ConsistencyCheckRequest{"alice", uint64(2), uint64(11)})
```
masomel: I'm a bit confused about the `uint64(11)` part. If `N = 10` at the beginning of the test and the update loop goes to `i < N` meaning that the loop ends when `N = 9`, how does and end epoch of 11 work? (The loop in `d.ConsistencyCheck()` retrieves all STRs up to the end epoch inclusive, but the latest epoch we have is 10).

protocol/directory_test.go
```diff
@@ -138,3 +139,231 @@ func TestRegisterExistedUserWithoutTB(t *testing.T) {
...
+	if res.Type != ConsistencyCheckType {
+		t.Fatal("Expect response type", ConsistencyCheckType, "got", res.Type)
+	}
+	if err != Success {
+		t.Fatal("Unable to perform key lookup in epoch", 2)
+	}
+	if len(res.AP) != 10-2 || len(res.STR) != 10-2 {
```
masomel: Why not compute `numEpochs = 10-2` only once?

protocol/directory.go
c633: It should be handled in `HandleOps`.

protocol/message.go
```diff
@@ -8,6 +8,9 @@ import (
 // of messages exchanged by clients and servers.
 const (
 	RegistrationType = iota
+	KeyLookupType
+	KeyLookupInEpochType
+	ConsistencyCheckType
```
masomel: I think `MonitoringType` might be a better name for `ConsistencyCheckType` since that's really the only time a client should send this type of request.

protocol/message.go
```diff
@@ -39,10 +57,39 @@ type DirectoryProof struct {
...
+}
+
+type DirectoryProofInEpoch struct {
+	Type  int
+	AP    *m.AuthenticationPath
+	STR   []*m.SignedTreeRoot
+	Error ErrorCode
+}
+
+type DirectoryProofs struct {
```
masomel: I think we might be able to consolidate `DirectoryProofInEpoch` and `DirectoryProofs` and just have the client make sure that the length of the `aps` struct is of length 1 when doing a `KeyLookupInEpoch`.

protocol/directory.go
masomel: Ah ok. But currently, such an error is only handled in the registration case in `HandleOps`.

keyserver/server_test.go
Liamsi: Glad I could help :-) I only set `DynamicRecordSizingDisabled` on the client side which obviously isn't enough... 
I still think we shouldn't rely that a single call of Read/Write reads/writes all the data we expect. As we saw this is pretty much implementation specific (even if it works now).

c633: We can check for Write but for Read, it seems we have to send the length of the buffer first?

protocol/directory_test.go
c633: Yep, it makes sense.

protocol/directory_test.go
c633: Yes, probably redundant.

protocol/message.go
c633: How about `KeyMonitoringType`?

protocol/directory_test.go
c633: >  how does and end epoch of 11 work?

This test is to make sure that the returned response does not have more than 'latestEpoch - startEpoch` STR(s) even if `endEpoch > latestEpoch`. How does it sound?

protocol/message.go
c633: I did it in ea4a0cc092ea99798d5f5ecbae6807ac38b5a505. I also added new `Limit` field into the `KeyLookupInEpochRequest`. What do you think?
/cc @arlolra @Liamsi

protocol/directory.go
c633: Thanks! I fixed it in ec0d1b8ca099f06b4fa4a7d916a35ed2675e0f77.

keyserver/server_test.go
Liamsi: True, or we could read until `io.EOF`, or use `ioutil.ReadAll`.

c633: I tried both way, but it turned out that the Read actually stopped if the connection is timed out. (We are using `for` loop in the server side).

protocol/directory.go
c633: > I think this should return a NewErrorResponse(), no?

You're right. It totally makes sense now. Thanks! Fixed in 5b780addc637e26562b30eab53b59673e473b3c8.

protocol/directory_test.go
masomel: I see, it definitely makes sense to test this. But it wasn't clear that this is what this part of the test is checking. I'm thinking this test should have two parts since checking that the server returns a valid hashchain given a range of epochs is maybe a different test than checking that the number of STRs returned is correct.

protocol/message.go
masomel: I have a question about the new `Limit` field. Could you please elaborate on what you mean by "control the response length"?

c633: >  How would a client that doesn't know how many epochs have passed use the Limit?

Hmm, I think the client should lookup for range from lookup epoch to its current latest epoch (the epoch of current saved STR). Even so, the `Limit` field seems to make no sense here. I originally thought that if the client want to do a key lookup in epoch, say, a month ago, then I'm not sure how big the response's size is?

Let me think more about it ...

protocol/directory_test.go
c633: It makes a lot of sense. Thanks!

protocol/message.go
masomel: > I think the client should lookup for range from lookup epoch to its current latest epoch (the epoch of current saved STR).

I think this works since we can assume that the current saved epoch will always be greater than the lookup epoch, because the client should monitor/catch up as soon as the account logs in.

> I originally thought that if the client want to do a key lookup in epoch, say, a month ago, then I'm not sure how big the response's size is?

This is definitely a good point. For such a long request, we should definitely have some way of dealing with a large response size. Maybe `Limit` could be used to indicate the maximum size for a batched response?

> Let me think more about it ... 

Definitely!

c633: I moved our discussion to #70.

masomel:
This pull LGTM. I think we can merge this, and add the chunked response optimization we've been discussing in #70 later.

keyserver/server_test.go
```diff
@@ -17,6 +18,27 @@ import (
 	"github.com/coniks-sys/coniks-go/utils"
 )
 
+var registrationMsg = `
+{
+	"type": 0,
+	"request": {
+		"username": "alice@twitter",
+		"key": [0,1,2],
+        "allow_unsigned_key_change": true,
```
arlolra: tabs

protocol/directory.go
```diff
@@ -59,9 +59,27 @@ func (d *ConiksDirectory) HandleOps(req *Request) (Response, ErrorCode) {
 		if msg, ok := req.Request.(*RegistrationRequest); ok {
 			if len(msg.Username) > 0 && len(msg.Key) > 0 {
 				res, e := d.Register(msg)
-				if res == nil {
-					return NewErrorResponse(e), e
-				}
+				return res, e
```
arlolra: `return d.Register(msg)`

protocol/directory.go
```diff
@@ -59,9 +59,27 @@ func (d *ConiksDirectory) HandleOps(req *Request) (Response, ErrorCode) {
...
+			}
+		}
+	case KeyLookupType:
+		if msg, ok := req.Request.(*KeyLookupRequest); ok {
+			if len(msg.Username) > 0 {
+				res, e := d.KeyLookup(msg)
+				return res, e
```
arlolra: same

protocol/directory.go
```diff
@@ -59,9 +59,27 @@ func (d *ConiksDirectory) HandleOps(req *Request) (Response, ErrorCode) {
...
+			}
+		}
+	case KeyLookupInEpochType:
+		if msg, ok := req.Request.(*KeyLookupInEpochRequest); ok {
+			if len(msg.Username) > 0 {
+				res, e := d.KeyLookupInEpoch(msg)
+				return res, e
```
arlolra: same

protocol/directory.go
```diff
@@ -59,9 +59,27 @@ func (d *ConiksDirectory) HandleOps(req *Request) (Response, ErrorCode) {
...
+			}
+		}
+	case MonitoringType:
+		if msg, ok := req.Request.(*MonitoringRequest); ok {
+			if len(msg.Username) > 0 && msg.StartEpoch <= msg.EndEpoch {
+				res, e := d.Monitor(msg)
 				return res, e
```
arlolra: same

protocol/directory.go
```diff
@@ -91,14 +109,77 @@ func (d *ConiksDirectory) Register(req *RegistrationRequest) (
...
+	if !bytes.Equal(ap.LookupIndex, ap.Leaf.Index) {
+		// if not found in the tree, do lookup in tb array
+		if d.useTBs {
+			if tb := d.tbs[req.Username]; tb != nil {
+				return NewKeyLookupProof(ap, d.LatestSTR(), tb, Success)
+			}
+		}
+		return NewKeyLookupProof(ap, d.LatestSTR(), nil, ErrorNameNotFound)
+	}
+	return NewKeyLookupProof(ap, d.LatestSTR(), nil, Success)
```
arlolra: Just a suggestion.  Little clearer like,

```
if bytes.Equal(ap.LookupIndex, ap.Leaf.Index) {
  return NewKeyLookupInEpochProof(ap, strs, Success)
}
// if not found in the tree, do lookup in tb array
if d.useTBs {
  if tb := d.tbs[req.Username]; tb != nil {
    return NewKeyLookupProof(ap, d.LatestSTR(), tb, Success)
  }
}
return NewKeyLookupProof(ap, d.LatestSTR(), nil, ErrorNameNotFound)
```

You'd want to switch up the end of `KeyLookupInEpoch` to be the same though.

protocol/directory.go
```diff
@@ -91,14 +109,77 @@ func (d *ConiksDirectory) Register(req *RegistrationRequest) (
...
+}
+
+func (d *ConiksDirectory) KeyLookupInEpoch(req *KeyLookupInEpochRequest) (
```
arlolra: How or why is this used?  Isn't it the same the `Monitor` request, minus the auth paths?

c633: `KeyLookupInEpoch` returns the auth path in the looking up epoch only, while `Monitor` returns a range of auth paths from [start, end]. Is that what you're asking?

arlolra: When would you use it?

protocol/directory.go
c633: Thanks!

protocol/directory.go
c633: Do you mean the `KeyLookupInEpoch`? 

Is it this?
> Because, what if the current epoch is 5? We're looking for proof that at epoch 0, a key was absent (which they all should be)

arlolra: > Do you mean the KeyLookupInEpoch?

Yeah, what request is the client making that requires that?  I only see the need for `KeyLookup` and `Monitor`, but maybe I'm missing something.

arlolra: In other words, all the functionality of `KeyLookupInEpoch` is covered by `Monitor` now.

arlolra: So, can we get rid of it?

c633: > What request is the client making that requires that?
>> What if Bob goes offline for a couple epochs and needs to verify a message I sent him while he was offline?

> In other words, all the functionality of KeyLookupInEpoch is covered by Monitor now.

I think the main difference is that `KeyLookupInEpoch` would be used for looking up your contacts' key in past epochs, whereas `Monitor` would be used for monitoring your own binding. 

Maybe we want to discuss this in #70 ?

arlolra: That still sounds like a `Monitor` request.

> Maybe we want to discuss this in #70 ?

Sure, let's move it there.

I've fallen behind on that thread anyways so a good excuse to catch up.

PR #75, opened 2016-08-21
CloseWrite to indicate end

arlolra:
This should prevent the need for `DynamicRecordSizingDisabled`

I don't think you intended to keep long living tcp connections.  If we want them, we probably need a protocol above that layer, even if it's just sending the length to read in the first byte.

### DIFF
--- a/keyserver/listener.go
+++ b/keyserver/listener.go
@@ -1,6 +1,7 @@
 package keyserver
 
 import (
+	"bytes"
 	"crypto/tls"
 	"io"
 	"log"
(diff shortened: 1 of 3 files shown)


keyserver/listener.go
```diff
@@ -32,7 +33,6 @@ func (server *ConiksServer) listenForRequests(ln *net.TCPListener, handler func(
 			log.Printf("accept client: %s", err)
 			continue
 		}
-		conn = tls.Server(conn, server.tlsConfig)
```
c633: Could we keep this and remove the L47?

keyserver/listener.go
```diff
@@ -43,30 +43,25 @@ func (server *ConiksServer) listenForRequests(ln *net.TCPListener, handler func(
...
+	}
 
-		res, err := handler(buf[:n])
-		if err != nil {
-			log.Printf("client handle %v: %v", conn.RemoteAddr(), err)
-		}
+	res, err := handler(buf.Bytes())
+	if err != nil {
+		log.Printf("client handle %v: %v", conn.RemoteAddr(), err)
+		// FIXME: what're we doing with the error here?
```
c633: I return `err` here for logging, it should be the same error with `res.Error`.

arlolra: I see ... that's non-obvious.  I'll add a comment.

keyserver/listener.go
arlolra: We could ... I did this to match the client (and also because I was called `c.CloseWrite()` after the write here, but that wasn't necessary since we're closing the entire connection).

But, out of curiosity, why do you prefer that?

keyserver/listener.go
arlolra: It might be clearer to not return an error and instead inspect the res.Error code to see if it isn't success.

keyserver/listener.go
c633: > But, out of curiosity, why do you prefer that?

Since I also handle an unix socket connection here (https://github.com/coniks-sys/coniks-go/blob/63ef04220a778a4707329a701625dbdd16470684/keyserver/listener.go#L40-L42).

arlolra: Ok, sure.

keyserver/listener.go
c633: I still prefer to return the err here instead of inspecting the res.Error. Because I want to avoid type assertion as much as possible.

arlolra: Don't the responses satisfy an interface already? Change,

```
type Response interface{}
```
to
```
type Response interface{
  Error ErrorCode
}
```

c633: Aha, yup. I will need to think more before replying ...

c633:
This pull LGTM.

keyserver/listener.go
arlolra: I see.  `logMsg` would be clearer, yes.  Also, if the type was `error` that'd be good too.

arlolra:
Feel free to merge it then.

PR #78, opened 2016-09-01
Remove Policies interface

c633:
Remove `Policies` interface and use `ConiksPolicies` struct instead.

### DIFF
--- a/merkletree/pad.go
+++ b/merkletree/pad.go
@@ -26,7 +26,7 @@ type PAD struct {
 
 // NewPAD creates new PAD consisting of an array of hash chain
 // indexed by the epoch and its maximum length is len
-func NewPAD(policies Policies, signKey sign.PrivateKey, len uint64) (*PAD, error) {
+func NewPAD(policies *Policies, signKey sign.PrivateKey, len uint64) (*PAD, error) {
 	if policies == nil {
 		panic(ErrorNilPolicies)
 	}
@@ -37,15 +37,14 @@ func NewPAD(policies Policies, signKey sign.PrivateKey, len uint64) (*PAD, error
 	if err != nil {
 		return nil, err
 	}
-	pad.policies = policies
+	pad.policies = *policies
 	pad.snapshots = make(map[uint64]*SignedTreeRoot, len)
 	pad.loadedEpochs = make([]uint64, 0, len)
 	pad.updateInternal(nil, 0)
 	return pad, nil
 }
 
-// if policies is nil, the previous policies will be used
-func (pad *PAD) signTreeRoot(m *MerkleTree, epoch uint64) {
+func (pad *PAD) signTreeRoot(epoch uint64) {
 	var prevHash []byte
 	if pad.latestSTR == nil {
 		var err error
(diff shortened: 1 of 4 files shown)


arlolra:
Looks good, but are there any other policies besides CONIKS policies?  If not, maybe just call this `Policies`?

c633:
Yes, it makes more sense.

merkletree/str.go
```diff
@@ -21,10 +21,10 @@ type SignedTreeRoot struct {
 	PreviousEpoch   uint64
 	PreviousSTRHash []byte
 	Signature       []byte
-	Policies        Policies
+	Policies        *Policies
```
arlolra: This is a late night comment, so take it for what it's worth.  Is there any chance that changing this to a pointer is going to lead to the Policies in the STR being mutated?

c633: I think it's still safe by far. Changing this to pointer absolutely makes the Policies being mutated but it's the same for other fields of slice type.

arlolra: Sort of ... there's an expectation that policies are to be able to be altered between epochs.  With this change, we're asking developers to be aware that they need to allocate a new Policies structure each time, rather than manipulate the one they may be retaining since startup. It just feels like a copy is less error prone.

After writing that sentence, I looked at directory.go and saw that we're doing this right,
https://github.com/coniks-sys/coniks-go/blob/master/protocol/directory.go#L43
but still feel uncomfortable with the pointer.  Maybe others have an opinion?

Also, I think we have a bug #80

c633: > With this change, we're asking developers to be aware that they need to allocate a new Policies structure each time, rather than manipulate the one they may be retaining since startup

I don't think that we're asking developers to allocate a new Policies struct each time. They just need to allocate a new struct when the policies change, much like when we using the interface. 
@masomel @Liamsi : What do you think?

Liamsi: Currently, the policies are accessed in the `merkletree` package only. Couldn't we even make the field unexported? So we hide the inner workings and still use a pointer here (which was introduced for performance reasons, I guess?)

c633: Unfortunately, we need to export Policies's fields for json marshalling.

Liamsi: Thank for the clarification. Then I'm also slightly in favour of making this a copy instead of a pointer. Simply because it is slightly more explicit.

Liamsi:
LGTM (no strong opinion on this discussion: https://github.com/coniks-sys/coniks-go/pull/78#discussion_r77117504)

merkletree/str.go
masomel: I think I agree with @arlolra and @Liamsi, a copy seems safer than a pointer.

c633: I fixed it in ceb0a58cdd139cb212c07dccabfb2f07be51520a. But a part of me still feels uncomfortable when considering the memory use between a copy and a pointer, in this case.

Maybe another solution is to ask developers to be aware that they need to allocate a new Policies structure each time, in the documentations.

arlolra: Isn't the overhead massively dominated by the cloned trees in memory?  The tradeoff is the potential for a corrupted state, which seems worth it.

Unfortunately, documenting that sort of expectation is unreliable and I would caution to err on the side of programming defensively.

c633: > Unfortunately, documenting that sort of expectation is unreliable and I would caution to err on the side of programming defensively.

Yup, I will take it as a lesson. Thanks!

masomel:
> Btw, what do you think about ceb0a58? 

LGTM.

merkletree/pad.go
```diff
@@ -21,12 +21,12 @@ type PAD struct {
 	snapshots    map[uint64]*SignedTreeRoot
 	loadedEpochs []uint64 // slice of epochs in snapshots
 	latestSTR    *SignedTreeRoot
-	policies     Policies // the current policies in place
+	policies     *Policies // the current policies in place
```
arlolra: `pad.policies` is in use, so the same mutation comment applies here.

I think you should revert everything in this patch that isn't strictly about removing the interface, sorry.

c633: > pad.policies is in use, so the same mutation comment applies here.

Yup, that's why I have this bug bc36c45e36c9eb51bc7510034ac43f80c44f3b30 (fixed in this commit). Thanks!

> I think you should revert everything in this patch that isn't strictly about removing the interface, sorry.

Oops, my bad. I split it into a separate commit (af994216261da5c00278929e1638e3ca682d4247)

Btw, I'm so sorry because of these late commits and response.

protocol/directory.go
```diff
@@ -39,12 +39,15 @@ func (d *ConiksDirectory) Update() {
...
 
-func (d *ConiksDirectory) SetPolicies(epDeadline merkletree.TimeStamp, vrfKey vrf.PrivateKey) {
-	d.policies = merkletree.NewPolicies(epDeadline, vrfKey)
+func (d *ConiksDirectory) UpdatePolicies(epDeadline merkletree.TimeStamp, vrfKey vrf.PrivateKey) {
+	d.policies.Update(epDeadline, vrfKey)
 }
 
+// EpochDeadline returns the EpochDeadline of the current policies
+// If you want to query the EpochDeadline before the new policies
+// is updated, use LatestSTR().Policies.EpochDeadline
```
arlolra: But that's exactly what you want to use in `EpochUpdate`.

https://github.com/coniks-sys/coniks-go/blob/master/keyserver/server.go#L192

The timer needs to fire based on `LatestSTR`'s deadline.

Again, I really think you should revert everything here that isn't about removing the interface.  The new test is good to add, but orthogonal.

c633: > Again, I really think you should revert everything here that isn't about removing the interface. The new test is good to add, but orthogonal.

I didn't get your idea yet :( What should be removed?

arlolra: Sorry for not being clear. Just the `UpdatePolicies` stuff.  The shallow copy in `pad.policies = *policies` is all we were after.

Let me update this, one sec.

arlolra:
@c633 See d0aff9a.  If you're happy with that, please merge.  Thanks!

c633:
Thanks!

PR #79, opened 2016-09-01
Convert Response interface to struct

c633:
- Convert `Response` interface to a struct like `Request`
- Refactor server_test.go code

### DIFF
--- a/protocol/message.go
+++ b/protocol/message.go
@@ -60,16 +60,14 @@ type Response struct {
 type DirectoryResponse interface{}
 
 type DirectoryProof struct {
-	Type int
-	AP   *m.AuthenticationPath
-	STR  *m.SignedTreeRoot
-	TB   *m.TemporaryBinding `json:",omitempty"`
+	AP  *m.AuthenticationPath
+	STR *m.SignedTreeRoot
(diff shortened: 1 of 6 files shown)


protocol/message.go
```diff
@@ -52,48 +52,77 @@ type MonitoringRequest struct {
 }
 
 // Response messages
-type Response interface{}
-
-type ErrorResponse struct {
-	Error ErrorCode
+type Response struct {
```
arlolra: Shouldn't this be like?

```
type Response struct {
  Error        ErrorCode
  Type         int
  Response     interface{}
}
```

c633: I thought about it but in case the user requested an invalid message type, which's type the server should return in the response?

arlolra: In that case, we should drop `Type` from the response altogether.  `ExpectingDirProofResponse` is hinting at the fact that we already know what type to expect for responses anyways.

c633: > ExpectingDirProofResponse is hinting at the fact that we already know what type to expect for responses anyways.

You're right. I think we should drop `Type` from the response.

arlolra:
LGTM

masomel:
LGTM, too

PR #80, opened 2016-09-01
Only update policies immediately after epoch

arlolra:
TBs are handed out with the current vrf key.  If it's swapped afterwards, the indices won't matchup at the next epoch.

Not sure how crazy I am about the `UpdatePolicies` struct.  This is more about identifying the issue.

### DIFF
--- a/protocol/directory.go
+++ b/protocol/directory.go
@@ -15,10 +15,15 @@ type ConiksDirectory struct {
 	policies merkletree.Policies
 }
 
+type UpdatePolicies struct {
+	EpDeadline merkletree.TimeStamp
+	VrfKey     vrf.PrivateKey
+}
+
 func InitDirectory(epDeadline merkletree.TimeStamp, vrfKey vrf.PrivateKey,
 	signKey sign.PrivateKey, dirSize uint64, useTBs bool) *ConiksDirectory {
 	d := new(ConiksDirectory)
-	d.SetPolicies(epDeadline, vrfKey)
+	d.setPolicies(epDeadline, vrfKey)
 	pad, err := merkletree.NewPAD(d.policies, signKey, dirSize)
 	if err != nil {
 		panic(err)
@@ -31,15 +36,18 @@ func InitDirectory(epDeadline merkletree.TimeStamp, vrfKey vrf.PrivateKey,
 	return d
 }
 
-func (d *ConiksDirectory) Update() {
+func (d *ConiksDirectory) Update(up *UpdatePolicies) {
 	d.pad.Update(d.policies)
 	// clear issued temporary bindings
 	for key := range d.tbs {
 		delete(d.tbs, key)
 	}
+	if up != nil {
+		d.setPolicies(up.EpDeadline, up.VrfKey)
+	}
 }
 
-func (d *ConiksDirectory) SetPolicies(epDeadline merkletree.TimeStamp, vrfKey vrf.PrivateKey) {
+func (d *ConiksDirectory) setPolicies(epDeadline merkletree.TimeStamp, vrfKey vrf.PrivateKey) {
 	d.policies = merkletree.NewPolicies(epDeadline, vrfKey)
 }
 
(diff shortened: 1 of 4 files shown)


protocol/directory.go
```diff
@@ -15,10 +15,15 @@ type ConiksDirectory struct {
 	policies merkletree.Policies
 }
 
+type UpdatePolicies struct {
+	EpDeadline merkletree.TimeStamp
```
masomel: I'm a bit confused as to what `EpDeadline` represents now. I thought this is an actual time stamp for when clients should expect the next epoch, but I don't see it being updated every epoch (and I'm embarrassed I didn't notice before).

protocol/directory.go
```diff
@@ -31,15 +36,18 @@ func InitDirectory(epDeadline merkletree.TimeStamp, vrfKey vrf.PrivateKey,
 	return d
 }
 
-func (d *ConiksDirectory) Update() {
+func (d *ConiksDirectory) Update(up *UpdatePolicies) {
```
masomel: I think the `epochDeadline` should change every epoch (see https://github.com/coniks-sys/coniks-go/pull/80/commits/6b273ae7ac50dcda47091f950fa968bfd34793e7#r77198064). I'm wondering if the epoch deadline shouldn't be passed in from the server.

arlolra: See https://github.com/coniks-sys/coniks-go/blob/master/keyserver/server.go#L192

protocol/directory.go
arlolra: I think it represents how long until the next epoch (used to set a timeout).  See the link below.

protocol/directory.go
masomel: Ok yeah, I saw that, which is why I made my comment. The point of the epoch deadline was to give the clients a firm time for when to expect the next STR (hence the name deadline and the `TimeStamp` type). If it's just a time interval, clients have to compute for themselves if the server met the deadline, while if it's an actual time stamp, barring clock synchronization issues, the server is forced to publish the next STR by that time and clients can easily verify that the server has done this.

arlolra: Ok, sounds like we got this wrong then.  Let's open another issue though.

arlolra: At some point we discussed all this (to be able to store STRs in an array and ease the lookup).  I can't find a pointer to it though.  Maybe it was only on hangout.  We should get better at documenting these things that diverge from the paper.

c633: > If it's just a time interval, clients have to compute for themselves if the server met the deadline

Is that what we do to avoid the clock synchronization issues? I remember that we agreed to let client compute the timestamp for themselves (on hangout). If `EpochDeadline` is an actual time stamp, I think there has not much advantages when using `Epoch` as counter ?

protocol/directory.go
```diff
@@ -31,15 +36,18 @@ func InitDirectory(epDeadline merkletree.TimeStamp, vrfKey vrf.PrivateKey,
...
 	d.pad.Update(d.policies)
 	// clear issued temporary bindings
 	for key := range d.tbs {
 		delete(d.tbs, key)
 	}
+	if up != nil {
+		d.setPolicies(up.EpDeadline, up.VrfKey)
```
c633: Can you elaborate on that? If we change the `vrfKey` at epoch `t=2`,  then the `pad` is updated at epoch `t=3` with the old `vrfKey` and next, we update the directory's policies with new key. Thus, the new `vrfKey` will be used at epoch `t=4`, right?

protocol/directory.go
masomel: > I can't find a pointer to it though. Maybe it was only on hangout. We should get better at documenting these things that diverge from the paper.

Same here, I looked, but couldn't find any.

>  We should get better at documenting these things that diverge from the paper.

Absolutely!

> I remember that we agreed to let client compute the timestamp for themselves (on hangout).

Ah, I couldn't remember, sorry :(

>  If EpochDeadline is an actual time stamp, I think there has not much advantages when using Epoch as counter ?

We should discuss this in a separate issue as Arlo said.

arlolra:
Whooops

This is already handled with,
https://github.com/coniks-sys/coniks-go/blob/6b273ae7ac50dcda47091f950fa968bfd34793e7/protocol/directory.go#L40

PR #84, opened 2016-09-08
ConiksDirectory.EpochDeadline() should return EpochDeadline from the PAD's Policies

c633:
If the developers want to query `EpochDeadline` before the new policies is updated, `ConiksDirectory.EpochDeadline()` should return the `EpochDeadline` from the pad's policies instead of the current policies of the directory.


arlolra:
Yup, LGTM.

Issue #52, opened 2016-08-02
Which PKI should we use for the server's keys?


masomel:
This discussion began in https://github.com/coniks-sys/coniks-go/pull/39#discussion_r73075691

masomel:
This relates to #47.

arlolra:
> Which PKI should we use for the server's keys?

CONIKS all the way down.

masomel:
I'm thinking maybe something more like Certificate Transparency would work for the server since the server's name is part of a global namespace. Related to this is also the fact that there aren't as many privacy concerns regarding servers' keys; we want anyone to be able to lookup and use the a server's keys.

arlolra:
> CONIKS all the way down.

That was tongue in cheek, https://en.wikipedia.org/wiki/Turtles_all_the_way_down

masomel:
I got the "turtles all the way down" reference :) I still wanted to put that idea out there.

Issue #47, opened 2016-08-02
What's the protocol for rotating the signing key?


arlolra:
See https://github.com/coniks-sys/coniks-go/pull/39#discussion_r73082470

PR #86, opened 2016-09-19
[WIP] Write client-side protocol doc

### LINKED ISSUE #82: Add doc.go to client
This doc should detail all client-side operations, including possible attacks and countermeasures.

Protocols covered:
- [x] Registration
- [ ] Lookup in current epoch
- [ ] Lookup in previous epoch
- [ ] Monitoring for current epoch
- [ ] Monitoring with missed epochs
- [ ] Auditing in current epoch

masomel:
Part of #82


masomel:
TODO: Make sure to include assumptions about attacks, Tofu etc.

client/doc.go
```diff
@@ -0,0 +1,42 @@
...
+Registration
+
+- The user registers her username with the CONIKS client.
+
+- The client generates a new public-private key pair for this username,
+and stores the keys on the device.
+
+- The client sends a registration request
+        reg_req = (username, key)
+to the server (may be validated by a registration bot; see
```
Liamsi: Doesn't the client actually send its request to the `bot` instead of the coniks-/key-server? At least the code looks like it isn't possible that the client contacts the server directly for registration.

Liamsi: https://github.com/coniks-sys/coniks-go/blob/master/keyserver/server.go#L158-L170

masomel: It does in our implementation, though the bot is more of an implementation detail than part of the protocol, so I want to make that distinction clear.

arlolra: We should make those lines in server.go configurable.

Liamsi: Thanks for the clarification @masomel!
I think, if this file describes the actual (client-)protocol implementation it should at least mention the fact that the client (currently) sends a registration request to the bot (or, indirectly to the server but *through* the bot).
I opened an issue as @arlolra suggested (https://github.com/coniks-sys/coniks-go/issues/91)

masomel:
I've included all of the assumptions and possible attacks on the registration protocol, but I think (a) this section is actually not specific to the client but rather about the registration protocol as a whole, and (b) really long and possibly more suitable in a separate document (maybe a wiki page?). This way, we can keep this doc specific to the client-side protocol specs. If we separate the spec from this discussion, I think this will make this doc easier to read, and we can refer the reader to the discussion about assumptions and possible attacks. What do you think?

masomel:
> I agree, I think this section should belong to the `protocol` package. 

Hmm, the core registration protocol (as well as the other protocols) is actually implemented in the `protocol` package. I'm starting to think about how the documentation of the two packages (`client` vs `protocol`) is going to differ.

> Yes, since we also want to use this document for coniks-java, a wiki page on coniks.org makes more sense to me

That's a good idea! I hadn't considered using this documentation for `coniks-java`, too. I think dedicating a page to this on coniks.org makes a lot of sense. I've been meaning to move the coniks.org code to a separate repo under coniks-sys, so I'll do that and create a separate pull for this page in that repo.

Liamsi:
My take on this is the question where the documentation should go is the following:  Documentation relevant for developers, and documentation describing what the code is doing and maybe briefly describing why (assumptions etc), directly should go in the documentation of that code (`doc.go` or even a Readme if it's not only about the go code). On the other hand, I agree that documentation that in detail explains assumptions, threat models, or any details beyond what is necessary for a dev to understand, could go somewhere else. And coniks.org seems like a very good place.

client/doc.go
```diff
@@ -0,0 +1,156 @@
...
+Introduction
+
+One crucial component of the CONIKS key management system are the clients
+verifying the cryptographic proofs returned by the CONIKS server as part of
+the registration, lookup and monitoring protocols. client implements the
+operations performed by CONIKS clients during these protocols. This document
+outlines each of these protocols from a client perspective.
+
+For the server-side specification of the CONIKS protocols, see
+https://godoc.org/github.com/coniks-sys/coniks-go/server
```
arlolra: https://godoc.org/github.com/coniks-sys/coniks-go/keyserver

client/doc.go
```diff
@@ -0,0 +1,156 @@
...
+
+Registration
+
+Prerequisites: This protocol requires that the client pin both the
+server's public signing key for verifying signed tree roots (STRs) and
+the server's initial private index verification key (i.e. the public
+VRF key). We assume a separate PKI exists to manage the server's keys.
```
arlolra: Why pin the VRF key?  Isn't that in the policy?

> We assume a separate PKI exists to manage the server's keys.

Is this sentence about TLS?

client/doc.go
```diff
@@ -0,0 +1,156 @@
...
+auditing protocol specification. *Note: Auditing is not implemented yet.*
+
+- Assuming the audit passes, the client proceeds to verify the TB, if
+included: it verifies the server's signature on th TB and checks that
+they public key in the TB matches the public key it sent to the proxy
+in reg_req.
+
+- Next, the client verifies that auth_path is a proof of absence by
+checking that its private index is invalid for the username, and its
+commitment is invalid for the public key it registered in reg_req.
```
arlolra: > its commitment is invalid for the public key it registered in reg_req.

Why that?

client/doc.go
```diff
@@ -0,0 +1,156 @@
...
+The client then recomputes the server's directory root using auth_path,
+and verifies that str includes the recomputed root in its signature.
+
+- If the registration proof is invalid, the client notifies the user. The
+Developing the reporting mechanism is planned for a future release.
```
arlolra: typo here

> The Developing

client/doc.go
```diff
@@ -0,0 +1,156 @@
...
+
+* The registration request may be denied for one of the following reasons:
+(1) The client attempts to register a name that already exists in the
+CONIKS key directory (ErrorNameExisted); the server returns a
+privacy-preserving proof of inclusion in its response. (2) The server
+encounters an internal error when attempting to register the name
+(ErrorDirectory).
```
masomel: Hmm yeah. If there's an implementation error, though, how can the registration request succeed?

client/doc.go
masomel: No, this sentence isn't about TLS, though, now that you mention it, we do need to pin the server's TLS certificate as well.

I thought we had discussed needing to pin the initial VRF key, since the VRF key in the policy indicates the VRF key that will be used in the next epoch, not the current epoch. Or maybe I'm misremembering what we had decided about which VRF key is included in the policy?

arlolra: > No, this sentence isn't about TLS

So then what other PKI is necessary besides the pinning?  I guess this is the unresolved questions in #47 and #52.

> Or maybe I'm misremembering what we had decided about which VRF key is included in the policy?

I think you're confusing the two keys.  It's the next signing key that we're considering also including in the policy.  The VRF corresponds to a particular epoch so doesn't require any more management.

client/doc.go
masomel: So, because this documentation is meant to describe the client-side protocols at a lower level for developers, I think it's perfectly reasonable to include relevant implementation details, such as the fact that a registration request won't succeed if the server encounters some internal error. 

Maybe "denied" isn't the right term to use in this case, but rather use "success/failure" terminology?

masomel: Yeah, I'm writing this with the expectation that the audience of our godocs will most be developers (i.e. this should be more of an API doc than a spec). This is probably something I should have made clear from the start, sorry about that!

masomel:
This [discussion](https://github.com/coniks-sys/coniks-go/pull/86#discussion_r83597218) got me thinking: Since this whole document I've written so far is really a spec, shouldn't it all belong in a specification document on coniks.org (much like Tor maintains spec docs on its page)? This doc.go should still include a brief description of each protocol, but only enough to explain how the API is used. The details of the protocol should probably be laid out in a specification document. Thoughts?

client/doc.go
```diff
@@ -0,0 +1,156 @@
...
+allow a malicious server to place the client
+on a forked branch of its history. However, by auditing the initial STR,
+the client can detect this attack. Alternatively, the client could verify
+the server's entire prior history to ensure that the server has not attempted
+to equivocate about its initial STR. This approach would be more secure,
+but has performance penalties. For now, we believe auditing will mitigate
+this attack, but the need for a full prior history check upon registration
+remains under discussion.
+
+As mentinoed above, for performance reasons, the client does not check
```
arlolra: typo

mentioned

client/doc.go
```diff
@@ -0,0 +1,156 @@
...
+the server's prior hisotry upon registering a new name. This enables
+a malicious server to remove a legitimate username from the directory
+and allow another user to re-register the name as the registering
+client would not discover the prior existence of the name.
```
arlolra: I'm not sure how comfortable I am with this paragraph.  It might be better framed as the registering client not being able to tell if its name had previously been registered with the server.  At present, it sounds like anyone can register any name they want, which is misleading.

client/doc.go
masomel: I forgot to run the spell checker on this document :(

client/doc.go
```diff
@@ -0,0 +1,156 @@
...
+registering client can detect this
+attack during the auditing step of the registration protocol.
+
+What these attacks have in common is that they take advantage
+of the fact that the registration protocol does not require the client
+to verify the server's prior history allowing an illegitimate
+re-registration of a username.  As a countermeasure, CONIKS does not
+currently allow re-registration of retired usernames,
+i.e. the user has deactivated and forfeit her account thereby making her
+username available for use by another user. The design decision is
```
masomel: Right, not allowing re-registrations is by policy, but the append-only property of the tree is by design, though. You're right, there's no proof that a name was registered and then removed, so verifying the entire prior history upon registering a new name would mitigate this issue.

client/doc.go
masomel: This is my mistake. I thought our verification of a proof of absence also does a commitment verification.

masomel:
I've removed the registration protocol spec I started here and will continue working on it at https://github.com/coniks-sys/coniks-spec/tree/protocol-spec. I'll address reviews for https://github.com/coniks-sys/coniks-go/pull/86/commits/4c48f9500ab5f849880990a2ae6a0d85f891eb61 there.

Since we're holding off on the cgo hooks for our first release, the only thing in the `client` pkg will be the test client. I've edited the doc.go to reflect this, but I'm not sure what other information belongs in there at this point. @Liamsi @c633 @arlolra

c633:
I think we can close this and write the documentation for `client` in #93.

Liamsi:
> I think we can close this and write the documentation for client in #93.

Yes, for the API documentation that pull is good candidate for the documentation. For the RFC/spec-like, more general documentation, I agree that https://github.com/coniks-sys/coniks-spec/tree/protocol-spec, is the right choice!

PR #88, opened 2016-09-21
Key-value DB using `gob` for encoding/decoding

vqhuy:
I have made an improvement in commit 84fc324bc27fee50c0619a9538c0d91421b804d1. This implementation's performance is better than #37 now. 

**Update** (description below)

This pull uses `encoding/gob` to encode/decode the merkletree structs to/from a buffer, and it implements the db backend hook for the merkletree package.

The package `merkletree` is now exporting some functions for encoding/decoding the PAD, STR, Merkletree and Policies from a buffer (any types that implement the `io.Writer` and `io.Reader` interface). This way, I think we can hide the detail implementations of the encoding/decoding from the developers.

Move to the hook implementation, there are two cases when we want to use a persistent storage:
- Store the current directory (pad). 
- Store the old STRs.

This hook exports APIs for store/load the PAD, STR to/from the db.

### DIFF
--- a/merkletree/merkletree.go
+++ b/merkletree/merkletree.go
@@ -2,7 +2,9 @@ package merkletree
 
 import (
 	"bytes"
+	"encoding/gob"
 	"errors"
+	"io"
 
 	"github.com/coniks-sys/coniks-go/crypto"
 	"github.com/coniks-sys/coniks-go/utils"
@@ -74,10 +76,10 @@ func (m *MerkleTree) Get(lookupIndex []byte) *AuthenticationPath {
 		direction := lookupIndexBits[depth]
 		var hashArr [crypto.HashSizeByte]byte
 		if direction {
-			copy(hashArr[:], nodePointer.(*interiorNode).leftHash)
+			copy(hashArr[:], nodePointer.(*interiorNode).LeftHash)
 			nodePointer = nodePointer.(*interiorNode).rightChild
 		} else {
-			copy(hashArr[:], nodePointer.(*interiorNode).rightHash)
+			copy(hashArr[:], nodePointer.(*interiorNode).RightHash)
 			nodePointer = nodePointer.(*interiorNode).leftChild
 		}
 		authPath.PrunedTree = append(authPath.PrunedTree, hashArr)
@@ -91,13 +93,13 @@ func (m *MerkleTree) Get(lookupIndex []byte) *AuthenticationPath {
 	case *userLeafNode:
 		pNode := nodePointer.(*userLeafNode)
 		authPath.Leaf = &ProofNode{
-			Level:      pNode.level,
-			Index:      pNode.index,
-			Value:      pNode.value,
+			Level:      pNode.Level,
+			Index:      pNode.Index,
+			Value:      pNode.Value,
 			IsEmpty:    false,
-			Commitment: &crypto.Commit{pNode.commitment.Salt, pNode.commitment.Value},
+			Commitment: &crypto.Commit{pNode.Commitment.Salt, pNode.Commitment.Value},
 		}
-		if bytes.Equal(nodePointer.(*userLeafNode).index, lookupIndex) {
+		if bytes.Equal(nodePointer.(*userLeafNode).Index, lookupIndex) {
 			return authPath
 		}
 		// reached a different leaf with a matching prefix
@@ -108,8 +110,8 @@ func (m *MerkleTree) Get(lookupIndex []byte) *AuthenticationPath {
 	case *emptyNode:
 		pNode := nodePointer.(*emptyNode)
 		authPath.Leaf = &ProofNode{
-			Level:      pNode.level,
-			Index:      pNode.index,
+			Level:      pNode.Level,
+			Index:      pNode.Index,
 			Value:      nil,
 			IsEmpty:    true,
 			Commitment: nil,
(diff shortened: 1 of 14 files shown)


storage/kv/merkletreekv/storage.go
```diff
@@ -0,0 +1,12 @@
...
+
+const (
+	EmptyNodeIdentifier    = 'E'
+	LeafNodeIdentifier     = 'L'
+	InteriorNodeIdentifier = 'I'
+	NodeKeyIdentifier      = 'N'
+	STRIdentifier          = 'S'
+	PoliciesIdentifier     = 'P'
+	EpochIdentifier        = 'O'
+	TreeNonceIdentifier    = 'T'
```
Liamsi: Arent' those constants the same as in https://github.com/coniks-sys/coniks-go/blob/84fc324bc27fee50c0619a9538c0d91421b804d1/merkletree/storage.go ?

merkletree/node.go
```diff
@@ -7,28 +7,30 @@ import (
...
 	node
-	key        string
-	value      []byte
-	index      []byte
-	commitment *crypto.Commit
+	Level      uint32
+	Key        string
+	Value      []byte
+	Index      []byte
+	Commitment *crypto.Commit
```
Liamsi: I guess all this values need to public to correctly encode now?

liamsi (commented):
Looking good. I added a few questions. Hope to do a more detailed review before our hangouts session later.

storage/kv/merkletreekv/storage.go
c633: Oh, this is a old commit that has been deleted.

merkletree/node.go
c633: Yes, since we didn't export the structs so I think it still ensures our preservation, i.e., there's no way to change or read the node's fields.
Btw, the reason is (mostly) because I didn't want to implement the GobEncode/GobDecode interface manually.

merkletree/merkletree.go
```diff
@@ -13,8 +15,8 @@ var (
 )
 
 const (
-	EmptyBranchIdentifier = 'E'
-	LeafIdentifier        = 'L'
+	EmptyNodeKey = 'E'
+	LeafNodeKey  = 'L'
```
Liamsi: Nitpicking: We would need this naming to be consistent. I liked the identifier (we already have so many different "keys" ;-) Here, we still stick with identifier: https://github.com/coniks-sys/coniks-go/pull/88/files#diff-7e689735c327b0bd737cd6e956e20b4aR4
Anyways, choose what you prefer. I just think it should be consistent.

storage/kv/merkletreekv/strkv.go
```diff
@@ -0,0 +1,35 @@
...
+func LoadSTR(db kv.DB, epoch uint64) (*merkletree.SignedTreeRoot, error) {
+	bbuff, err := db.Get(STRKey(epoch))
+	if err != nil {
+		return nil, err
+	}
+	buff := bytes.NewBuffer(bbuff)
+	return merkletree.DecodeSTR(buff)
+}
+
+func STRKey(epoch uint64) []byte {
```
Liamsi: This could be unexported or not? If this should stay exported it would need a more speaking name.

liamsi (changes requested):
Some suggestions/questions.

merkletree/merkletree.go
c633: Oops, it must be my mistake. I have no intention to change this! Thanks!

storage/kv/merkletreekv/strkv.go
c633: Do you have any suggestion for this? :D

Liamsi: If it is unexported the name is OK.

Liamsi: Make it unexported

storage/kv/merkletreekv/strkv.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"bytes"
+
+	"github.com/coniks-sys/coniks-go/merkletree"
+	"github.com/coniks-sys/coniks-go/storage/kv"
+	"github.com/coniks-sys/coniks-go/utils"
+)
+
+func StoreSTR(wb kv.Batch, str *merkletree.SignedTreeRoot) error {
```
Liamsi: This and `LoadSTR` currently aren't used. I guess this will change?

c633: They would be used to load the old STRs which were deleted from the `pad.snapshot`.

Liamsi: Cool, thanks. Can we have a test for that?

Liamsi:
From my point of view this PR looks very good (just a few questions and nitpicks).

storage/kv/merkletreekv/strkv.go
c633: Yup!

c633:
Thanks!

masomel (changes requested):
The encoding, decoding and reconstruction functions aren't part of the core data structure, they are only relevant to the DB, and they are ultimately only used in `EncodePAD`, which is really only used by `StorePAD` in padkv.go. As such, all gob encoding and decoding functions related to the DB should be part of the `storage/kv/merkletreekv` package, not the core `merkletree` package. 

I agree that the encoding format is an implementation detail we should hide from the developer, but I don't see the reason why these functions need to be part of the core data structure if they are used only for a separate feature (the DB) of coniks-go.

liamsi (dismissed):

c633:
PTAL. I need to use this in extensions implementation ;)

Issue #90, opened 2016-09-22
Client cli


Liamsi:
The (test-)client command line interface should have the the following functionality:
- register a name-to-key binding
- update a name-to-key-binding
- lookup a key 

And deal with all the details/edge-cases that come with that, e.g, dealing with temporary bindings etc

The cli should clarify what a real client would need to implement _and_ what a client would need to know in advance/before doing its first request (like initial public key of the key-server, key-server's epoch duration, or "initial" STR if necessary).

For a start, we omit the possibility to alter the key-change policy (which is currently possible in the java-test client: https://github.com/coniks-sys/coniks-java/blob/master/coniks_test_client/src/main/java/org/coniks/coniks_test_client/TestClient.java#L447)

c633:
@Liamsi Maybe we can reuse this testutil (https://github.com/coniks-sys/coniks-go/blob/master/keyserver/testutil/testutil.go#L124)?

Liamsi:
Looks useful indeed. Thanks!

Liamsi:
Is there an equivalent dealing with the registration (from a clients perspective/directly contacting the twitter bot)? I see there is also https://github.com/coniks-sys/coniks-go/blob/master/keyserver/testutil/testutil.go#L154 but this is only for internal communication, thouugh.

c633:
> Is there an equivalent dealing with the registration (from a clients perspective/directly contacting the twitter bot)?

Yes, it is.

Liamsi:
> Yes, it is.

Could you point me to the code? I didn't find code which actually send the particular DMs to the bot (only code that directly calls `TwitterBot.HandleRegistration`).
If the command line client should help implementing a real client it should also use the twitter API (like the TorMessenger would), I guess? 
Hope I can finish the client today/tomorrow...

c633:
Oops, I misinterpreted your question. Sorry :(
I think we don't need to run a Twitter Bot in the command line client test, it should just connect to the server directly via a unix socket connection.
> If the command line client should help implementing a real client it should also use the twitter API (like the TorMessenger would)

Yes, if we decide to go to with a real client, we will need to use the twitter API (I think we can re-use the `go-twitter` package).

Ahh, sorry again :(

Liamsi:
Thanks for your reply and no worries!

> I think we don't need to run a Twitter Bot in the command line client test, it could bypass the bot and just connect to the server directly via a unix socket connection.

[...]

PR #74, opened 2016-08-18
Add verification methods

c633:
To be paired w/ #86 
- Temporarily ignore the cgo client library until we have a complete `protocol/client` library.

PR #93, opened 2016-09-24
go testclient

Liamsi:
Minimalistic test-client. Part of #90. 
- currently registration & lookup is implemented

Differences to a real client implementation: 
- client doesn't use (twitter-)bot for registration (see discussion: https://github.com/coniks-sys/coniks-go/issues/90#issuecomment-249356102)
- thus, registration only works from the same machine as the server is running on
- client does not store and verify the hash chain
- client uses `testutils` package (which doesn't verify server's tls-cert. chain, for instance)

I would prefer to have the current code reviewed and create separate (smaller) issues/PRs for the above bullet-points.

TODOs:
- [x] ~~integrate with #74 as soon as it is merged~~ **Update**:Add TODOs for parts of the code where the actual client struct/methods from #74 would be used (actually tried to make use of the consistency checks as much as possible)
- [x] Fix: json: cannot unmarshal object into Go value of type merkletree.AssocData (ugly hack ...)
- [x] Add any API documentation needed (formerly #86).


Liamsi:
@c633 @masomel @arlolra Whoever has time: I think this is ready for a first review.

keyserver/listener.go
```diff
@@ -107,7 +107,7 @@ func (server *ConiksServer) makeHandler(acceptableTypes map[int]bool) func(msg [
 		response, e := server.dir.HandleOps(req)
 		switch req.Type {
 		case KeyLookupType, KeyLookupInEpochType, MonitoringType:
-			server.RLock()
+			server.RUnlock()
```
c633: Could you rebase this pull on top of master branch?

c633:
> # wait for epoch to finish and try
./testclient lookup -n alice again

Did you mean

```
# wait for epoch to finish and try again
./testclient lookup -n alice
```

client/testclient/cmd/lookup.go
```diff
@@ -0,0 +1,100 @@
...
+		if err != nil {
+			fmt.Println("Couldn'r create request!")
+			os.Exit(-1)
+		}
+		resp, err := testutil.NewTCPClient(req)
+		if err != nil {
+			fmt.Println("Error while retrieving repsonse: " + err.Error())
+			os.Exit(-1)
+		}
+		response, errCode := client.UnmarshalResponse(p.KeyLookupType,
```
c633: For L36 - L80, I think we will eventually have `client.Verify(operationType, response)` function that wraps all these verifications. This function would also need to handle this TODO (https://github.com/coniks-sys/coniks-go/blob/master/client/cgo/verification.go#L40-L45), much like what you did below. What do you think? 

The same for registration code.

client/testclient/cmd/lookup.go
```diff
@@ -0,0 +1,100 @@
...
+		if !ok {
+			fmt.Println("Got unexpected response from server!")
+			os.Exit(-1)
+		}
+		switch errCode {
+		case p.Success:
+			c := res.Verify(name, []byte(res.AP.Leaf.Value), 0, nil,
+				conf.SigningPubKey)
+			// verify auth. path:
+			if c != p.Passed || res.AP == nil ||
```
c633: Shouldn't this verification be included in `res.Verify`? (see https://github.com/coniks-sys/coniks-go/blob/953f8fd0e24fb07f993c571bf38a4b7aed4db015/protocol/verification.go)

c633:
@Liamsi : Thanks for this great work! I added a few questions. 
Anw, I think this pull still needs to wait for #74 and #86 being merged first (see my comments).

keyserver/listener.go
Liamsi: Done

Liamsi:
> Anw, I think this pull still needs to wait for #74 and #86 being merged first (see my comments).
Thanks for you feedback.
#74 makes a lot of sense but why wait for the documentation ( #86 )?

client/testclient/cmd/lookup.go
Liamsi: Yes, I guess this makes sense. Especially if we want a clear API someone can use as a reference to implement his own client.

client/testclient/cmd/lookup.go
Liamsi: Makes sense! Do you think this should be done in this PR, too? (or in a separate one)

c633:
I think @arlolra  wants to use #86 as a reference for reviewing our implementation.

client/testclient/cmd/lookup.go
c633: Maybe in a separate one. I can handle this.

masomel:
@Liamsi I'm trying to run the test client with the following command:

```
testclient register -config /disk/local/melara/coniks-testclient/config.toml -n alice -k totally-fake-key -s /tmp/coniks.sock
```

and get the following error:  ``Failed to load config: open onfig: no such file or directory``. The `-config` flag doesn't seem to be working (looks like it's interpreting `-config` as `-c onfig`).

Liamsi:
@masomel It seems like you have to call it with `--config` instead of `-config` if you choose the long flag name. I didn't know that either (I'm using the cobra framework for the first time, too) but if you run it with the `-h` flag it shows ` -c, --config string   Config file ...`.

client/testclient/cmd/common.go
```diff
@@ -0,0 +1,39 @@
...
+Then, create a toml file which references these public-keys.
+
+For example create a file config.toml with the following content:
+
+vrf_pubkey_path = "/path_to/vrf.pub"
+sign_pubkey_path = "/path_to_/sign.pub"
+
+The client looks for a file called 'config.toml' in its current working directory.
+If you prefer the config-file to be named or stored somewhere different you can
+specify where to look for the config with the -config flag. For example:
```
masomel: `-config` should be `--config` or `-c`

masomel:
@Liamsi Ah ok, thanks for looking into this! I've added a comment in that file to reflect this in the usage message.

client/testclient/cmd/common.go
Liamsi: Oops, fixed in: a1f8c6f

client/testclient/cmd/register.go
```diff
@@ -0,0 +1,116 @@
...
+var registerCmd = &cobra.Command{
+	Use:   "register",
+	Short: "Register a name-to-key binding.",
+	Long: `Register a new name-to-key binding on the CONIKS-server.
+Note that, for the registration to work, the test-client needs to be running on the same machine as the key-server.
+A real client would first register an app here: https://apps.twitter.com (the UI should make this possible without actually leaving the client application). Then, the client would use the oauth1 API (find the tokens on the [Keys and Access Tokens] tab of the aforementioned page) and contact the (twitter-)bot for a registration request. The twitter bot is also accessible from outside.
+For more information consult: https://godoc.org/github.com/coniks-sys/coniks-go/bots
+
+Example call:
+  coniksclient register -name Alice@twitter -key fake_test_key`,
```
masomel: s/-name/--name and s/-key/--key

Liamsi: I should have sticked with the short version of flags in the usage string (which I tried myself) ;-) Thanks for testing it! Fixed in 	bd8d071

masomel:
@Liamsi LGTM. The test client works great!

client/testclient/cmd/lookup.go
c633: If we do so, should we also need to verify the hash chain as well?

c633: I think we can store the latest STR to `client/config.toml` (with an explanation that that field is being used for testing only, the developers should actually store the latest STR in some persistent storages) and whenever the test client runs, it reads the latest STR from this config file. What do you think?

client/testclient/cmd/common.go
c633: Maybe a `mkConfig` is fair enough? (see my comment https://github.com/coniks-sys/coniks-go/pull/93#discussion_r80605436).

Liamsi: I do not fully understand your comment. I guess, you mean create a `mkConfig`just for the client (makes sense); but how does that relate to storing the STR in the config, too?

client/testclient/cmd/lookup.go
Liamsi: > If we do so, should we also need to verify the hash chain as well?

I think, yes we should.

Liamsi: > I think we can store the latest STR to client/config.toml 

Yeah, maybe for testing purposes that's a good idea!

client/testclient/cmd/common.go
Liamsi: Are you proposing that for boostraping the chain we should write the initial STR into the config file? If yes, `mkConfig` is called before the server is started and hence before the first STR is created.

c633: I just want to have a generated template file for the test client.

> Are you proposing that for boostraping the chain we should write the initial STR into the config file?

It's a good idea. But do we run the test client on the same machine as the key server?

Liamsi: > I just want to have a generated template file for the test client.

OK, makes sense.

> It's a good idea. But do we run the test client on the same machine as the key server?

Currently, the testclient only works if ran on the same machine as the key-server: https://github.com/coniks-sys/coniks-go/issues/90#issuecomment-249356102
Eventually, this will change.

c633: Yup, great!

Liamsi: Done (`mkConfig`) as discussed in https://github.com/coniks-sys/coniks-go/pull/93/commits/2638588c9163dd217d6640eca4e8143e13426b7d

client/testclient/cmd/lookup.go
Liamsi: > Maybe in a separate one. I can handle this.

I could also take care of this (if you haven't started yet)

client/testclient/cmd/common.go
c633: Great!

client/testclient/cmd/register.go
```diff
@@ -0,0 +1,116 @@
...
+			}
+			// verify signature on TB:
+			tbb := res.TB.Serialize(res.STR.Signature)
+			if !conf.SigningPubKey.Verify(tbb, res.TB.Signature) {
+				fmt.Println("Couldn't verify signature of temporary binding.")
+			}
+			fmt.Println("Succesfully registered name: " + name)
+			// TODO Where should the client save seen STRs and the TB?
+			// (or should it wait till the next epoch to see if the
+			// TB was actually inserted?)
```
c633: There maybe have more than one TB since the client could also receive the TB of her contacts, and we definitely have to check if the TB as been included as promised. Given that we will have a `ConiksClient` to store verified STRs and TBs (see https://github.com/coniks-sys/coniks-go/pull/74/commits/266d8cd3c55756d627e1e0581f8f3cc2ec3413a0#diff-75d470fe65b4ef49cd55654004d3edeaR29), I'm thinking that should we store both STRs and TBs in this struct?

Liamsi: >There maybe have more than one TB since the client could also receive the TB of her contacts, and we definitely have to check if the TB as been included as promised. 

The client struct looks definitely like a good idea and I agree this would be the good place to store STRs and TBs (at least in memory). The current `ConiksClient` also only stores one TB, or not?

c633: It should store an array of TB. Wait me a sec.

client/testclient/cmd/update.go
```diff
@@ -0,0 +1,23 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/spf13/cobra"
+)
+
+// TODO figure out if this needs to be a separate command or we can simply re-use
```
c633: Or we can temporarily remove this command?

client/testclient/cmd/register.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/coniks-sys/coniks-go/keyserver/testutil"
+	p "github.com/coniks-sys/coniks-go/protocol"
+	"github.com/spf13/cobra"
+)
+
+var registerCmd = &cobra.Command{
+	Use:   "register",
+	Short: "Register a name-to-key binding.",
+	Long: `Register a new name-to-key binding on the CONIKS-server.
+Note that, for the registration to work, the test-client needs to be running on the same machine as the key-server.
```
c633: Since #130 has landed, could we remove this note?

client/testclient/cmd/update.go
Liamsi: Removed ...

client/testclient/cmd/register.go
Liamsi: Also removed

Liamsi:
This is ready for another review. I tried to use the consistency-checks to validate the server responses (they aren't stored yet). There is some ugly hack to unmarshal the `AssocData` field in the `SignedTreeRoot` struct. There should definitely be a more elegant way to deal with it but maybe this needs some refactoring (orthogonal to the test-client scaffolding). /cc @c633 @arlolra @masomel

arlolra:
Sorry for maybe having made things more convoluted than necessary.  See 3c66033 where I tried to clean up the FIXMEs a bit, though perhaps still not great (and the commit msg for another proposed solution).

client/config.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/coniks-sys/coniks-go/utils"
+	"io/ioutil"
+)
+
+// Config contains the client's configuration needed to send request to a
+// CONIKS-server:
+//  - the path to the server's VRF and signing public-key file, respectively;
+//  - the actual public-keys parsed from these files.
+type Config struct {
+	configPath string
```
c633: This seems to be used nowhere?

c633:
I think `merkletree` should be abstracted away from `client`, so it's worth to try Arlo's alternative solution. 
I did some TODO clean up in `client_go_2`, feel free to merge in if you approve these changes.

Liamsi:
The solution in https://github.com/coniks-sys/coniks-go/commit/3c6603393499ebf56ce063ba6e4839e2e3753df9 looks good on first sight but unfortunately the registration doesn't work anymore (`UnmarshalResponse` returns a "Malformed directory message"). It could be easy to fix, though. I'll check @c633's commits in client_go_2, first. (@arlolra @c633)

Liamsi:
>  I'll check @c633's commits in client_go_2, first. 

OK, the above mentioned decoding problem (introduced with https://github.com/coniks-sys/coniks-go/commit/3c6603393499ebf56ce063ba6e4839e2e3753df9) still exists on `client_go_2`. @c633's changes still look very cool. Thanks! I'll merge them.

Liamsi:
If you want to reproduce the problem: Put a `fmt.Println(err)` after the `json.Unmarshal` in the `case p.RegistrationType, p.KeyLookupType` of `UnmarshalResponse` or use a debugger; then, run a registration request. Now, instead of previously `json: cannot unmarshal object into Go value of type merkletree.AssocData` you'll see a an err with `json: cannot unmarshal object into Go value of type merkletree.RawAd` ...

c633:
This is my latest changes compare to this PR: https://github.com/coniks-sys/coniks-go/compare/client_go...client_go_2?expand=1. If you need me to do something, please tell me (I know you are very busy right now) :)

Liamsi:
Thanks @c633! Again, the latest commits there look very reasonable! But I guess they doesn't solve the `UnmarshalReponse` problem (https://github.com/coniks-sys/coniks-go/pull/93#issuecomment-262898783)

c633:
> But I guess they doesn't solve the UnmarshalReponse problem (#93 (comment))?

Yes, I just tried to clean up some TODOs, didn't dig into that problem yet :(

Liamsi:
There is a minor glitch in https://github.com/coniks-sys/coniks-go/commit/bdf9ee0e8a6d7427eebc52b012509088ebda8693#commitcomment-19960679 but I think I'll just merge `client_go_2` here and we'll see how to fix that later?

Liamsi:
> Yes, I just tried to clean up some TODOs, didn't dig into that problem yet :(

I just committed a fix for this one. Again, it's far from being nice but I only need to change two lines.

Liamsi:
There is one issue left: after a successful registration, the client, on lookup reports: `Error: [coniks] The directory returned an invalid registration promise` if you don't wait until the next epoch.

c633:
> There is one issue left: after a successful registration, the client, on lookup reports: Error: [coniks] The directory returned an invalid registration promise if you don't wait until the next epoch.

Oops, it must be a bug :(. Investigating now! Thank you!

c633:
> Oops, it must be a bug :(.

Yes, it is, stupid me. I fixed it in 89a4a2a282de4bb137bd68b7f5f988e69c2a4279.

Liamsi:
Wow, that was quick! 👍 Thanks a lot. For after the release: it might be better if we had a test for that, too.  
Thanks again :-)

c633:
> it might be better if we had a test for that case, too.

Absolutely. I fixed it in #133 (c4284c4ff0804cdac3dce4f510e838c70f5c4f2d).

protocol/tb.go
```diff
@@ -40,5 +40,5 @@ func (tb *TemporaryBinding) Serialize(strSig []byte) []byte {
 // the existed binding (TOFU).
 func (tb *TemporaryBinding) Verify(index, value []byte) bool {
 	return bytes.Equal(tb.Index, index) &&
-		(value != nil && bytes.Equal(tb.Value, value))
+		(value == nil || (value != nil && bytes.Equal(tb.Value, value)))
```
arlolra: See https://github.com/coniks-sys/coniks-go/pull/133#discussion_r89676845

client/config.go
```diff
@@ -0,0 +1,61 @@
...
+// Config contains the client's configuration needed to send request to a
+// CONIKS-server:
+//  - the path to the server's VRF and signing public-key file, respectively;
+//  - the actual public-keys parsed from these files.
+type Config struct {
+	SignPubkeyPath string `toml:"sign_pubkey_path"`
+	VrfPubkeyPath  string `toml:"vrf_pubkey_path"`
+
+	SigningPubKey sign.PublicKey
+	VrfPubKey     vrf.PublicKey
```
arlolra: This `VrfPubKey` isn't used.

client/config.go
```diff
@@ -0,0 +1,61 @@
...
+
+	RegAddress string `toml:"registration_address,omitempty"`
+	Address    string `toml:"address"`
+}
+
+// LoadConfig returns a client's configuration read from the given filename.
+// It reads the (VRF and signing) public-key files and parses the actual keys.
+// If there is any parsing or IO-error it returns and error (and the return
```
arlolra: an error

client/coniksclient/README.md
```diff
@@ -0,0 +1,35 @@
...
+|     |_ |       || | |   ||   ||    _  | _____| |
+|_______||_______||_|  |__||___||___| |_||_______|
+
+Usage:
+  coniksclient [command]
+
+Available Commands:
+  init        Creates a config file for the client.
+  lookup      Lookup a name.
+  register    Register a name-to-key binding.
```
arlolra: What we have here is fine for now.

Might be fun to use,
https://godoc.org/golang.org/x/crypto/ssh/terminal

client/coniksclient/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+Use "coniksclient [command] --help" for more information about a command.
+```
+
+## TODO
+- Update me
+- Clean up some more TODOs
+- Write doc (summary, etc)
```
arlolra: Can we remove this and open issues if necessary?

client/coniksclient/internal/cmd/register.go
```diff
@@ -0,0 +1,99 @@
...
+			return
+		}
+		req, err := createRegistrationMsg(name, key)
+		if err != nil {
+			fmt.Println("Couldn't marshal registration request!")
+			os.Exit(-1)
+		}
+
+		var res []byte
+		u, _ := url.Parse(conf.RegAddress)
```
arlolra: `conf.RegAddress` is defined with,

```
RegAddress string `toml:"registration_address,omitempty"`
```

Presumably this should fallback to `conf.Address` if empty.

merkletree/str.go
```diff
@@ -30,6 +31,41 @@ type SignedTreeRoot struct {
 	Ad              AssocData
 }
```
arlolra: Maybe add here,

// TODO(arlolra): Look into using a `DirSTR` instead, to avoid this mess.

protocol/tb.go
c633: It's already fixed in 9981d150dfbce72ec13ec16b07a64f0e0f565c72.

client/coniksclient/README.md
c633: Oh, sorry, this is my TODOs. I will remove it.

client/config.go
Liamsi: It should be used or not? (IIRC it was used before I changed the code to using the consistency-checks). Hope I'll find some time to look into this issue later.

c633: Yes, I think it's safe to remove it for now.

Liamsi:
Thanks for the changes @c633!

merkletree/str.go
c633: Do you want to do it in this PR, @arlolra @Liamsi ?

Liamsi:
I think it would be better if we merge this PR and open a separate issue
for how-to do this. (I'm currently in the train and mostly offline for the
next 7h btw.)

merkletree/str.go
arlolra: No, let's open a separate issue.  But, I'd like to add the TODO comment here.

c633: I opened #135 for this.

client/coniksclient/internal/cmd/register.go
```diff
@@ -0,0 +1,103 @@
...
+		req, err := createRegistrationMsg(name, key)
+		if err != nil {
+			fmt.Println("Couldn't marshal registration request!")
+			os.Exit(-1)
+		}
+
+		var res []byte
+		if conf.RegAddress == "" {
+			// fallback to conf.Address if empty
+			conf.RegAddress = conf.Address
```
arlolra: Can we define a `regAddress := conf.RegAddress` above and then set that here?

client/coniksclient/internal/cmd/register.go
```diff
@@ -0,0 +1,103 @@
...
+		"Username you want to register with the CONIKS server.")
+
+	registerCmd.Flags().StringP("key", "k", "",
+		"Key-material you want to bind to the user name.")
+	registerCmd.Flags().StringP("config", "c", "config.toml",
+		"Config file for the client (contains the server's initial public key etc.)")
+}
+
+func createRegistrationMsg(name, key string) ([]byte, error) {
+	return json.Marshal(&p.Request{
```
arlolra: This probably belongs in encoding.go

client/coniksclient/internal/cmd/register.go
```diff
@@ -0,0 +1,103 @@
...
+		err = cc.HandleResponse(p.RegistrationType, response, name, []byte(key))
+		switch err {
+		case p.CheckPassed:
+			switch response.Error {
+			case p.ReqSuccess:
+				fmt.Println("Succesfully registered name: " + name)
+				// TODO: Save the cc to verify the TB and for later
+				// usage (TOFU checks)
+			case p.ReqNameExisted:
+				// Key-change isn't currently supported; see:
```
arlolra: We won't be here for a key change because we won't get a `.CheckPassed`

client/coniksclient/internal/cmd/lookup.go
```diff
@@ -0,0 +1,90 @@
...
+		// FIXME same comment as in register.go
+		cc := p.NewCC(nil, true, conf.SigningPubKey)
+		err = cc.HandleResponse(p.KeyLookupType, response, name, nil)
+		switch err {
+		case p.CheckPassed:
+			switch response.Error {
+			case p.ReqSuccess:
+				key, err := response.GetKey()
+				if err != nil {
+					fmt.Println("Cannot the key from the response, error: " + err.Error())
```
arlolra: Cannot get the key

client/coniksclient/internal/cmd/register.go
c633: Probably we should finish #133 first.

arlolra: Ok, can we remove this comment in favour of a pointer to that then?

arlolra:
I think this looks good.  You should squash and merge it ;)

Liamsi:
> I think this looks good. You should squash and merge it ;)

I've followed the latest changes and I agree (I might be biased, though ;-). Thanks again for taking care and finishing this PR @c633 @arlolra!

PR #95, opened 2016-09-24
Fix #94

### LINKED ISSUE #94: Server "deadlock"
By sending multiple lookup requests to the server it will block forever. Most probably this is because on lookup the `RLocks` are never released: 
https://github.com/coniks-sys/coniks-go/blob/f21ba86a4a3ca89a0e82e881adc22ecddc78ffd9/keyserver/listener.go#L103-L110
(last line should be `server.RUnlock()` instead)

c633:
It should fix #94.


Liamsi:
LGTM

c633:
I think so. Merged in a4cd4e115778a2634324a6ef6296796d85a43d4d.

PR #97, opened 2016-09-27
cgoVerify should return errors of the response and the verification

c633:
We're going to want to verify some returned data, even when the response wasn't a success.

### DIFF
--- a/client/cgo/cgotest.go
+++ b/client/cgo/cgotest.go
@@ -1,19 +1,27 @@
 package main
 
 /*
-int testVerify(int type,
-	char *uname, int unameSize,
-	unsigned char *key, int keySize,
-	unsigned long long currentEpoch,
-	unsigned char *savedSTR, int strSize,
-	unsigned char *pk, int pkSize,
-	char *response, int responseSize) {
-	return cgoVerify(type,
-		uname, unameSize,
-		key, keySize,
-		currentEpoch,
-		savedSTR, strSize,
-		pk, pkSize, response, responseSize);
+struct cgoVerify_return {
+    int r0;
+    int r1;
+};
+extern struct cgoVerify_return cgoVerify(int p0, char* p1, int p2, void* p3,
+    int p4, long long unsigned int p5, void* p6, int p7, void* p8, int p9,
+    char* p10, int p11);
+
+struct cgoVerify_return testVerify(int type,
+    char *uname, int unameSize,
+    unsigned char *key, int keySize,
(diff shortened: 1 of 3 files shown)


client/cgo/cgotest.go
```diff
@@ -1,7 +1,13 @@
 package main
 
 /*
-int testVerify(int type,
+struct cgoVerify_return {
+	int r0;
+	int r1;
+};
+extern struct cgoVerify_return cgoVerify(int p0, char* p1, int p2, void* p3, int p4, long long unsigned int p5, void* p6, int p7, void* p8, int p9, char* p10, int p11);
```
masomel: This is just a stylistic comment: this is the only line that isn't 80 wide. Can you please change this?

client/cgo/verification.go
```diff
@@ -14,20 +14,25 @@ import (
...
-	cResponse *C.char, cResponseSize C.int) C.int {
+	cResponse *C.char, cResponseSize C.int) (C.int, C.int) {
 
 	if int(cUnameSize) == 0 ||
 		int(cKeySize) == 0 ||
 		(int(cStrSize) != sign.SignatureSize && int(cStrSize) != 0) ||
 		int(cPkSize) != sign.PublicKeySize ||
 		int(cResponseSize) == 0 {
-		return C.int(protocol.ErrorMalformedDirectoryMessage)
+		return C.int(protocol.ErrorMalformedDirectoryMessage), C.int(-1)
```
masomel: Instead of returning `-1`, I'm thinking we should return an called something like `ErrorCouldNotVerify`, which is defined in `protocol.error`. Same goes for L47.

client/cgo/cgotest.go
masomel: Also, you seem to be using `long long unsigned int` for `p5` here, but `unsigned long long` for `currentEpoch` below. Is this intentional?

client/cgo/verification.go
c633: Sure!

client/cgo/cgotest.go
c633: They should be the same, I didn't noticed that `long long unsigned int` was generated by cgo. Thanks!

c633: I fixed in ec29e2ef1527a27d72c4d1cf32870eadad8e2d68.

masomel:
LGTM

Liamsi:
LGTM, too.

PR #99, opened 2016-10-11
Treat ErrorCode as errors

c633:


Liamsi:
I might have missed the discussion around the error-codes (is there an issue, comment?). But this pull LGTM!

c633:
You didn't miss any discussion. I just knew `ErrorCode` can be treated as `error` by implementing `func Error() string`. I think it maybe look nicer if we do this. This is actually a minor pull.

protocol/error.go
```diff
@@ -36,26 +34,26 @@ var ErrorResponses = map[ErrorCode]bool{
...
+		ErrorDirectory:              "[coniks] Directory error",
+
+		Passed: "[coniks] Passed",
+		ErrorMalformedDirectoryMessage: "[coniks] Malformed directory message",
+		ErrorBadSignature:              "[coniks] Directory's signature is invalid",
+		ErrorBadVRFProof:               "[coniks] Bad VRF proof",
+		ErrorBadIndex:                  "[coniks] Directory returned a bad index",
+		ErrorBadMapping:                "[coniks] Returned name-to-key mapping is inconsistent with the root hash",
+		ErrorBadSTR:                    "[coniks] The hash chain is inconsistent",
+		ErrorBadCommitment:             "[coniks] Bad commitment",
```
masomel: Much like in my previous comments, I would make the error message for `ErrorBadCommitment` a bit more informative.

protocol/error.go
```diff
@@ -36,26 +34,26 @@ var ErrorResponses = map[ErrorCode]bool{
...
+		ErrorBadBinding:                "[coniks] Bad name-to-key binding",
```
masomel: I think the difference between `Mapping` and `Binding` isn't clear. To me, at least, they represent the same thing (mapping being a more general term) - a mapping from a name to a key. I would rename `ErrorBadMapping` to `ErrorBadAuthPath` and change the error message for `ErrorBadBinding` to something like "Key in the binding is inconsistent".

c633:
@masomel: Thanks for your suggestions. Can you help me commit these changes? :)

Liamsi:
> Update: I made these changes in 8c759c4. What do you think?

Still LGTM :-)

masomel:
LGTM, this can be merged!

PR #100, opened 2016-10-12
TemporaryBinding should include the issued epoch for verifications

c633:
After receiving a TB from the key server, the client needs to verify this TB at the next epoch. It could do this by compare the current saved STR with the STR included in the TB's signature, ~~which means the client needs to verify the TB's signature again. This could cause some overhead for the client.~~ (this is wrong). Thus, we add an extra field `IssuedEpoch` in the TB, but the key server isn't bound to `IssuedEpoch` by its signature. However, this field would be verified by the client when it receives the TB, using its saved STR, to prevent the tampering from the server.

### DIFF
--- a/merkletree/tb.go
+++ b/merkletree/tb.go
@@ -1,17 +1,26 @@
 package merkletree
 
-import "github.com/coniks-sys/coniks-go/crypto/sign"
+import (
+	"bytes"
+
+	"github.com/coniks-sys/coniks-go/crypto/sign"
+)
 
 type TemporaryBinding struct {
-	Index     []byte
-	Value     []byte
-	Signature []byte
+	// IssuedEpoch would not be included in the serialization.
+	// To verify the validation of IssuedEpoch, we can compare it
+	// against the epoch of STR included in the serialization.
+	IssuedEpoch uint64
+	Index       []byte
+	Value       []byte
+	Signature   []byte
 }
 
-func NewTB(key sign.PrivateKey, strSig, index, value []byte) *TemporaryBinding {
+func NewTB(key sign.PrivateKey, ep uint64, strSig, index, value []byte) *TemporaryBinding {
 	tb := &TemporaryBinding{
-		Index: index,
-		Value: value,
+		IssuedEpoch: ep,
+		Index:       index,
+		Value:       value,
 	}
 	tbPreSig := tb.Serialize(strSig)
 	tb.Signature = key.Sign(tbPreSig)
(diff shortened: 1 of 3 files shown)


merkletree/tb.go
```diff
@@ -25,3 +34,17 @@ func (tb *TemporaryBinding) Serialize(strSig []byte) []byte {
 	tbBytes = append(tbBytes, tb.Value...)
 	return tbBytes
 }
+
+func (tb *TemporaryBinding) Verify(epoch uint64, ap *AuthenticationPath) bool {
```
masomel: I think we should make it explicit that the epoch are verifying needs to come from an STR. Maybe we should be passing in an STR here instead of just an epoch (much like we're passing the whole auth path for the index and value comparison).

merkletree/tb.go
```diff
@@ -1,17 +1,26 @@
...
+	"bytes"
+
+	"github.com/coniks-sys/coniks-go/crypto/sign"
+)
 
 type TemporaryBinding struct {
-	Index     []byte
-	Value     []byte
-	Signature []byte
+	// IssuedEpoch would not be included in the serialization.
```
masomel: Hmm, I think we do want to include the `IssuedEpoch` in the serialization. Otherwise, the server isn't bound to the `IssuedEpoch` by its signature, and the field can be tampered with.

merkletree/tb.go
```diff
@@ -1,17 +1,26 @@
...
+	// To verify the validation of IssuedEpoch, we can compare it
+	// against the epoch of STR included in the serialization.
```
masomel: I'm assuming the validation you're referring to here is the client verifying that `IssuedEpoch == STR.Epoch+1` when it receives the TB?

merkletree/tb.go
c633: This field would be verified by the client when it receives the TB, I thought that's enough to prevent tampering. The main reason I think this field should be included is to help the client know when it should verify the promise.

merkletree/tb.go
c633: Yes.

merkletree/tb.go
c633: Hmm, I think we don't need to verify `IssuedEpoch` here.

c633:
Can we merge this pull yet, or do you want to include `IssuedEpoch` in the serialization? @masomel  @Liamsi ?

masomel:
I thought we decided that we would include `IssuedEpoch` in the serialization to be safe.

c633:
Yes, we did.

> where STR_n is the STR for the next epoch when the server needs to fulfill its promise

However, if STR_n is not that next epoch, the client needs to compare the STR by re-verifying the TB signature, right? I think it may cause some overhead for the client.

c633:
Closed since we can verify the TB by comparing `STR_tb.Epoch + 1 == STR_n.Epoch` where `STR_n` is the STR for the next epoch when the server needs to fulfill its promise.

Issue #32, opened 2016-07-22
Add doc.go file to each subpackage


masomel:
- [x] bots
- [x] client
- [x] crypto
- [x] keyserver
- [x] merkletree
- [x] utils
- [x] protocol

Liamsi:
Not sure if this is the right place but as it about documentation: An alternative to adding READMEs in each package could be adding a doc.go file (see here for an example of such a file: https://golang.org/src/encoding/gob/doc.go  and how the generated documentation would look like: https://godoc.org/encoding/gob)

c633:
I'm working on this in #102. Cooperate with me if you have time or have interest in this.

PR #103, opened 2016-10-15
Merkletree pkg documentation

c633:
Part of #32 
- Unexport functions and types that are not being used outside the package
- Add comments for exporting functions

### DIFF
--- a/merkletree/doc.go
+++ b/merkletree/doc.go
@@ -6,7 +6,17 @@ to help other developers use it in their implementation easily.
 
 Persistent Authenticated Dictionary
 
-This module implements a persistent authenticated dictionary (PAD) data structure.
+This module implements a persistent authenticated dictionary (PAD) data structure,
+which supports dictionary operations with two additional features:
+(1) lookups return a cryptographic proof of correctness along with the result,
+and (2) it supports taking and storing snapshots of the current contents of
+the dictionary to allow lookups in historical versions of the dictionary.
+In CONIKS, the general PAD design is extended to return proofs for inserts.
+This design does not support deletions, and that individual snapshots are linked
+via a hash chain to commit the entire history. This PAD implementation also supports
+randomizing the order of directory entries by changing the VRF private key.
+This protects the user's privacy against other malicious parties who
+wish to obtain information about users by querying the key directory.
 
 Merkle Prefix Tree
 
--- a/merkletree/doc.go
+++ b/merkletree/doc.go
@@ -14,8 +14,8 @@ This module implements the Merkle prefix tree, which is the data structure
 underlying our PAD implementation. It is a binary tree, and that there are
 two types of leaf nodes: empty leaf node and user leaf node. Each node contains
 its prefix index and its level value. It provides methods to insert, update
-(to update an existing key-value pair) and lookup. The tree is append-only,
-and it does not support the deletion of values.
+(to update an existing key-value pair) and lookup.
+The tree has the property of not being able to remove user leafs.
 This Merkle prefix tree implementation also preserves the privacy feature:
 the prefix used to search in the tree is a cryptographic transformation (VRF)
 of the search key, and values are concealed using cryptographic commitments.
--- a/merkletree/doc.go
+++ b/merkletree/doc.go
@@ -1,0 +1,29 @@
+/*
+Package merkletree implements a Merkle prefix tree and related data structures.
+The Merkle prefix tree is one of the most important components of the CONIKS protocol.
+We implemented this data structure separately as a library
+to help other developers use it in their implementation easily.
+
+Persistent Authenticated Dictionary
+
+This module implements a persistent authenticated dictionary (PAD) where the client
+can query a name-to-key binding at a specific epoch and get authenticated answer.
+The PAD is represented by a hash chain committing to the entire history. This hash chain
+is used to prove to the client that the PAD is maintaining a linear hash chain
+and there is no equivocations.
+
+Merkle Tree
+
+This module implements the Merkle prefix tree data structure. It provides methods to
+insert, update and lookup. The tree is append-only which means there is no ways to delete
+any node from the tree or change a user leaf node to an empty node.
+All the tree hash operations use the hash algorithm provided by our crypto package
+(see https://godoc.org/github.com/coniks-sys/coniks-go/crypto).
+
+Temporary Binding
+
+This module implements a signed promise for registration operations. The main purpose of
+temporary binding is to allow immediate use of the key without having to wait for the next epoch.
+This will be split into a separate protocol extension in a future release.
+*/
+package merkletree
--- a/merkletree/doc.go
+++ b/merkletree/doc.go
@@ -21,11 +21,11 @@ wish to obtain information about users by querying the key directory.
 Merkle Prefix Tree
 
 This module implements the Merkle prefix tree, which is the data structure
-underlying our PAD implementation. It is a binary tree, and that there are
+underlying our PAD implementation. It is a binary tree with
 two types of leaf nodes: empty leaf node and user leaf node. Each node contains
 its prefix index and its level value. It provides methods to insert, update
 (to update an existing key-value pair) and lookup.
-The tree has the property of not being able to remove user leafs.
+The tree has the property of not being able to remove user leaves.
 This Merkle prefix tree implementation also preserves the privacy feature:
 the prefix used to search in the tree is a cryptographic transformation (VRF)
 of the search key, and values are concealed using cryptographic commitments.
--- a/merkletree/doc.go
+++ b/merkletree/doc.go
@@ -19,11 +19,5 @@ insert, update and lookup. The tree is append-only which means there is no ways 
 any node from the tree or change a user leaf node to an empty node.
 All the tree hash operations use the hash algorithm provided by our crypto package
 (see https://godoc.org/github.com/coniks-sys/coniks-go/crypto).
-
-Temporary Binding
-
-This module implements a signed promise for registration operations. The main purpose of
-temporary binding is to allow immediate use of the key without having to wait for the next epoch.
-This will be split into a separate protocol extension in a future release.
 */
 package merkletree
--- a/merkletree/doc.go
+++ b/merkletree/doc.go
@@ -6,18 +6,20 @@ to help other developers use it in their implementation easily.
 
 Persistent Authenticated Dictionary
 
-This module implements a persistent authenticated dictionary (PAD) where the client
-can query a name-to-key binding at a specific epoch and get authenticated answer.
-The PAD is represented by a hash chain committing to the entire history. This hash chain
-is used to prove to the client that the PAD is maintaining a linear hash chain
-and there is no equivocations.
+This module implements a persistent authenticated dictionary (PAD) data structure.
 
-Merkle Tree
+Merkle Prefix Tree
 
-This module implements the Merkle prefix tree data structure. It provides methods to
-insert, update and lookup. The tree is append-only which means there is no ways to delete
-any node from the tree or change a user leaf node to an empty node.
-All the tree hash operations use the hash algorithm provided by our crypto package
+This module implements the Merkle prefix tree, which is the data structure
+underlying our PAD implementation. It is a binary tree, and that there are
+two types of leaf nodes: empty leaf node and user leaf node. Each node contains
+its prefix index and its level value. It provides methods to insert, update
+(to update an existing key-value pair) and lookup. The tree is append-only,
+and it does not support the deletion of values.
+This Merkle prefix tree implementation also preserves the privacy feature:
+the prefix used to search in the tree is a cryptographic transformation (VRF)
+of the search key, and values are concealed using cryptographic commitments.
+The VRF, commitment scheme and hash operations are provided by our crypto package
 (see https://godoc.org/github.com/coniks-sys/coniks-go/crypto).
 */
 package merkletree
--- a/merkletree/pad.go
+++ b/merkletree/pad.go
@@ -11,7 +11,12 @@ import (
 )
 
 var (
+	// ErrorSTRNotFound indicates that the STR has been evicted from memory,
+	// because the maximum number of cached PAD snapshots has been exceeded.
 	ErrorSTRNotFound = errors.New("[merkletree] STR not found")
+
+	// ErrorNilPolicies is used to panic if the PAD was created
+	// with a nil Policies struct.
 	ErrorNilPolicies = errors.New("[merkletree] Nil policies")
 )
 
@@ -26,7 +31,7 @@ type PAD struct {
 }
 
 // NewPAD creates new PAD consisting of an array of hash chain
-// indexed by the epoch and its maximum length is len
+// indexed by the epoch and its maximum length is len.
 func NewPAD(policies *Policies, signKey sign.PrivateKey, len uint64) (*PAD, error) {
 	if policies == nil {
 		panic(ErrorNilPolicies)
@@ -79,6 +84,11 @@ func (pad *PAD) updateInternal(policies *Policies, epoch uint64) {
 	}
 }
 
+// Update generates a new snapshot of the directory.
+// Specifically, it extends the hash chain by issuing
+// new signed tree root. It may remove some older signed tree roots from
+// memory if the cached PAD snapshots exceeded its maximum capacity.
+// The Policies pointer can be nil if the policies does not change.
 func (pad *PAD) Update(policies *Policies) {
 	// delete older str(s) as needed
 	if len(pad.loadedEpochs) == cap(pad.loadedEpochs) {
@@ -91,14 +101,22 @@ func (pad *PAD) Update(policies *Policies) {
 	pad.updateInternal(policies, pad.latestSTR.Epoch+1)
 }
 
+// Set computes the private index of the given name using
+// the current VRF private key (which will be inserted in the next
+// signed tree root) to create a new index-to-key binding,
+// and then Set inserts it into the Merkle tree underlying the PAD.
 func (pad *PAD) Set(name string, value []byte) error {
 	return pad.tree.Set(pad.Index(name), name, value)
 }
 
+// Lookup searches the requested name in the latest snapshot of the PAD.
 func (pad *PAD) Lookup(name string) (*AuthenticationPath, error) {
 	return pad.LookupInEpoch(name, pad.latestSTR.Epoch)
 }
 
+// LookupInEpoch searches the requested name in the snapshot at requested epoch.
+// It returns ErrorSTRNotFound if the signed tree root of the requested epoch
+// has been removed from memory.
 func (pad *PAD) LookupInEpoch(name string, epoch uint64) (*AuthenticationPath, error) {
 	str := pad.GetSTR(epoch)
 	if str == nil {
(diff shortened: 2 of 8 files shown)


merkletree/pad.go
```diff
@@ -10,7 +10,13 @@ import (
 )
 
 var (
+	// ErrorSTRNotFound indicates that the STR of the searching epoch
```
masomel: I would rephrase this to "the STR of a requested epoch"

merkletree/pad.go
```diff
@@ -10,7 +10,13 @@ import (
...
+	// has been removed from the memory, because the snapshot array has exceeded
+	// its maximum length.
```
masomel: This explanation may not make that much sense to a developer since the snapshot array isn't directly exposed (and there's currently no specification for our PAD design). Maybe say something like "the STR has been evicted from memory, because the maximum number of cached PAD snapshots has been exceeded" is a bit clearer?

merkletree/pad.go
```diff
@@ -10,7 +10,13 @@ import (
...
 	ErrorSTRNotFound = errors.New("[merkletree] STR not found")
+
+	// ErrorNilPolicies is used to panic if the PAD was created
+	// with a nil Policies.
```
masomel: I would change this to "a nil Polices struct". Otherwise, this is grammatically awkward even though it's correct terminology.

merkletree/pad.go
```diff
@@ -78,6 +84,10 @@ func (pad *PAD) updateInternal(policies *Policies, epoch uint64) {
 	}
 }
 
+// Update updates the history hash chain by issuing
+// new signed tree root. It may remove some older STRs from
+// memory if the snapshot array exceeded its maximum capacity.
+// The Policies pointer can be nil if the policies does not change.
```
masomel: I would start with more PAD-related terminology here: E.g. "Update generates a new snapshot of the directory. Specifically, it extends the hash chain by issuing..."

merkletree/pad.go
```diff
@@ -90,15 +100,19 @@ func (pad *PAD) Update(policies *Policies) {
 	pad.updateInternal(policies, pad.latestSTR.Epoch+1)
 }
 
+// Set calculates the private index of the given name and
+// inserts the new binding into the tree.
```
masomel: To be a bit more specific about what's going on here, I would clarify that `Set` creates a new index-to-key binding, and that `Set` inserts it into the Merkle tree underlying the PAD.

merkletree/pad.go
```diff
@@ -90,15 +100,19 @@ func (pad *PAD) Update(policies *Policies) {
...
 func (pad *PAD) Set(name string, value []byte) error {
 	index, _ := pad.computePrivateIndex(name, pad.policies.vrfPrivateKey)
 	return pad.tree.Set(index, name, value)
 }
 
+// Lookup searchs the requested name in the latest version of the tree.
```
masomel: s/searchs/searches, and again, I would refer to the PAD as opposed to the tree.

merkletree/pad.go
```diff
@@ -90,15 +100,19 @@ func (pad *PAD) Update(policies *Policies) {
...
 func (pad *PAD) Lookup(name string) (*AuthenticationPath, error) {
 	return pad.LookupInEpoch(name, pad.latestSTR.Epoch)
 }
 
+// LookupInEpoch searchs the requested name in the tree at requested epoch.
```
masomel: s/searchs/searches

merkletree/pad.go
```diff
@@ -110,17 +124,23 @@ func (pad *PAD) LookupInEpoch(name string, epoch uint64) (*AuthenticationPath, e
...
 	}
 	return pad.snapshots[epoch]
 }
 
+// LatestSTR returns the most recent version of the signed tree root.
 func (pad *PAD) LatestSTR() *SignedTreeRoot {
 	return pad.latestSTR
 }
 
+// TB issues a new promise for the new binding and
```
masomel: It may not be clear to a developer reading this what "the new binding" is, so I would make clear that this binding consists of the given name and value params.

merkletree/policy.go
```diff
@@ -8,6 +8,12 @@ import (
 
 type TimeStamp uint64
 
+// Policies is a summary of the directory's
+// current security policies. This includes the public part
+// of the VRF key used to generate private indices,
+// the cryptographic algorithms in use, as well as
+// the protocol version number.
+// See section 3.3 per paper for an detail example.
```
masomel: s/per paper for an detail/of the CONIKS paper for a detailed

I don't want to assume that a developer reading this necessarily knows which paper we're referring to.

merkletree/proof.go
```diff
@@ -36,6 +42,17 @@ func (n *ProofNode) hash(treeNonce []byte) []byte {
...
+// the prefix path between the corresponding leaf node
+// and the root. This is a proof of inclusion
+// or absence of requested index.
+// The proof type can be determined by
+//  if !bytes.Equal(ap.Leaf.Index, ap.LookupIndex) {
+//	 // Proof of absence
+//	} else {
+//	 // Proof of inclusion
+//  }
+// See section 3.2 per paper for detail.
```
masomel: Same here, I wouldn't assume that the developer knows which paper to look for.

merkletree/proof.go
```diff
@@ -59,7 +76,9 @@ func (ap *AuthenticationPath) authPathHash() []byte {
 	return hash
 }
 
-// Verify should be called after the vrf index is verified successfully
+// Verify verifies the authentication path against
```
masomel: "Verify verifies" sounds awkward. I would be a bit more specific about what this function does: "Verify recomputes the tree's root node from the authentication path, and compares it to the tree hash in the corresponding signed tree root".

merkletree/str.go
```diff
@@ -8,12 +8,12 @@ import (
...
 )
 
-// SignedTreeRoot represents a signed tree root, which is generated
+// SignedTreeRoot represents a signed tree root (STR), which is generated
 // at the beginning of every epoch.
 // Signed tree roots contain the current root node,
 // the current and previous epochs, the hash of the
 // previous STR, and its signature.
-// STR should be final
+// STR should be read-only.
```
masomel: This seems to imply that the developer should ensure that the STR is read-only. It might be helpful to be more specific as to why/how the developer can ensure this.

merkletree/str.go
```diff
@@ -56,6 +57,8 @@ func (str *SignedTreeRoot) Serialize() []byte {
 	return strBytes
 }
 
+// VerifyHashChain verifies the consistency of the current hash chain,
+// according to section 3.3 per paper.
```
masomel: As with a few other cases, I would be a bit more specific here about what `VerifyHashChain` actually does. Right now you're just stating the purpose of the function.

merkletree/pad.go
```diff
@@ -78,6 +83,11 @@ func (pad *PAD) updateInternal(policies *Policies, epoch uint64) {
 	}
 }
 
+// Update generates a new snapshot of the directory.
+// Specifically, it extends the hash chain by issuing
+// new signed tree root. It may remove some older signed tree roots from
+// memory if the cached PAD snapshots exceeded its maximum capacity.
```
masomel: Nitpicking: s/its/the

merkletree/pad.go
```diff
@@ -90,15 +100,23 @@ func (pad *PAD) Update(policies *Policies) {
 	pad.updateInternal(policies, pad.latestSTR.Epoch+1)
 }
 
+// Set computes the private index of the given name using
```
masomel: Do we want to use the term "key" instead of "name" here?

merkletree/pad.go
```diff
@@ -90,15 +100,23 @@ func (pad *PAD) Update(policies *Policies) {
...
 	index, _ := pad.computePrivateIndex(name, pad.policies.vrfPrivateKey)
 	return pad.tree.Set(index, name, value)
 }
 
+// Lookup searches the requested name in the latest snapshot of the PAD.
 func (pad *PAD) Lookup(name string) (*AuthenticationPath, error) {
 	return pad.LookupInEpoch(name, pad.latestSTR.Epoch)
 }
 
+// LookupInEpoch searches the requested name in the snapshot at requested epoch.
```
masomel: Nitpicking: s/requested/the requested

merkletree/pad.go
```diff
@@ -110,17 +128,24 @@ func (pad *PAD) LookupInEpoch(name string, epoch uint64) (*AuthenticationPath, e
 	return ap, nil
 }
 
+// GetSTR returns the signed tree root of the requested epoch.
+// This signed tree root is read from the caching snapshots of the PAD.
```
masomel: s/caching/cached

merkletree/proof.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/coniks-sys/coniks-go/utils"
 )
 
+// ProofNode can be a leaf node or an empty node,
+// which is included in the returned AuthenticationPath
+// of a given index. The type of that node can be determined
+// by IsEmpty value. It also provides an opening of
```
masomel: Nitpicking: s/by/by the

merkletree/proof.go
```diff
@@ -59,7 +75,9 @@ func (ap *AuthenticationPath) authPathHash() []byte {
 	return hash
 }
 
-// Verify should be called after the vrf index is verified successfully
+// Verify recomputes the tree's root node from the authentication path,
+// and compares it to the tree hash in the corresponding signed tree root.
```
masomel: This sentence could be a bit confusing since this function doesn't use any STRs. Maybe mention the `treeHash` parameter explicitly and mention that it has to come from an STR?

merkletree/str.go
```diff
@@ -56,6 +57,10 @@ func (str *SignedTreeRoot) Serialize() []byte {
 	return strBytes
 }
 
+// VerifyHashChain computes the hash of saved STR,
```
masomel: Nitpicking: s/of/of a

merkletree/str.go
```diff
@@ -56,6 +57,10 @@ func (str *SignedTreeRoot) Serialize() []byte {
...
+// and compares it to the hash of previous STR included
```
masomel: s/the hash of previous STR/the PreviousSTRHash (I think it's ok to just use the field name)

merkletree/tb.go
```diff
@@ -2,12 +2,20 @@ package merkletree
 
 import "github.com/coniks-sys/coniks-go/crypto/sign"
 
+// TemporaryBinding is a protocol extension which
+// serves two main functions: (1) as a signed promise
+// by the key server to include a newly registered
+// (or updated key) in the next epoch, and (2) to allow immediate
+// use of the key without having to wait for the next epoch.
```
masomel: As I mentioned before, I wouldn't use protocol-related terminology here. I think it's fine to say "PAD" instead of "key server" and "inserted" instead of "registered". Also, the use of "key" is ambiguous here, too.

c633:
> The main changes I suggest are to doc.go, which I think could use more detailed explanations of the data structures. Pretty much everything else are nitpicky edits.

Thanks! Sorry about those mistakes (I always try to avoid them when writing but apparently I didn't do it good enough)  :(

> Also, could you please save the new changes in a separate commit?

Sorry, I will do it.

masomel:
> Sorry about those mistakes (I always try to avoid them when writing but apparently I didn't do it good enough) :(

Don't worry, they really are tiny issues!!

merkletree/doc.go
```diff
@@ -0,0 +1,29 @@
...
+Package merkletree implements a Merkle prefix tree and related data structures.
+The Merkle prefix tree is one of the most important components of the CONIKS protocol.
+We implemented this data structure separately as a library
+to help other developers use it in their implementation easily.
+
+Persistent Authenticated Dictionary
+
+This module implements a persistent authenticated dictionary (PAD) where the client
+can query a name-to-key binding at a specific epoch and get authenticated answer.
+The PAD is represented by a hash chain committing to the entire history. This hash chain
```
c633: Thanks for clarifying!

c633:
> The main changes I suggest are to doc.go, which I think could use more detailed explanations of the data structures.

I suppose these explanations should go to https://github.com/coniks-sys/coniks-spec. Instead, I think I could write about `Set`, `Index`, `Sign` are done using the current PAD's underlying data structures (the tree or private keys), whereas `Lookup` is performed on the latest version of the tree. Some more detail about how the tree is rotated using the new VRF key, etc

masomel:
> I suppose these explanations should go to https://github.com/coniks-sys/coniks-spec.

The extra explanations I was suggesting aren't supposed to be a data structure specification, but only enough background for a developer to get a better idea of what the code does and the motivation behind it. The specification would have a much greater level of detail.

> I could write about `Set`, `Index`, `Sign` are done using the current PAD's underlying data structures (the tree or private keys), whereas `Lookup` is performed on the latest version of the tree. 

Those are important details since they tell the developer what is happening under the hood, but I think these explanations would belong with the individual methods that they refer to (much like what @Liamsi has pointed out for the protocol API doc as well).

> Some more detail about how the tree is rotated using the new VRF key, etc

This is an interesting point. This detail is really about `pad.reshuffle()`, right? Since this method isn't exported, a developer just reading the doc won't really know where to look. I'm wondering if this detail belongs with `pad.Update()`, which is where the reshuffling could happen (internally). But I think this is a detail a developer should have since it's not an obvious outcome of `Update`.

merkletree/doc.go
```diff
@@ -0,0 +1,25 @@
...
+
+This module implements a persistent authenticated dictionary (PAD) data structure.
+
+Merkle Prefix Tree
+
+This module implements the Merkle prefix tree, which is the data structure
+underlying our PAD implementation. It is a binary tree, and that there are
+two types of leaf nodes: empty leaf node and user leaf node. Each node contains
+its prefix index and its level value. It provides methods to insert, update
+(to update an existing key-value pair) and lookup. The tree is append-only,
```
arlolra: Maybe you mean the PAD is append-only?

merkletree/proof.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/coniks-sys/coniks-go/utils"
 )
 
+// ProofNode can be a leaf node or an empty node,
```
arlolra: s/leaf/user/

merkletree/doc.go
c633: I think both PAD and the tree are append-only

merkletree/proof.go
```diff
@@ -36,6 +42,16 @@ func (n *ProofNode) hash(treeNonce []byte) []byte {
 	}
 }
 
+// AuthenticationPath is a pruned tree containing
+// the prefix path between the corresponding leaf node
+// (of type ProofNode) and the root. This is a proof
+// of inclusion or absence of requested index.
+// The proof type can be determined by
```
arlolra: Maybe just say, a proof of inclusion is when the leaf index equals the lookup index?

merkletree/doc.go
arlolra: You can update the value in the tree and there's no record of its previous content.  It alone isn't, as I understand it.  The PAD, however, is only updated once an epoch and contains a snapshot of previous values.

merkletree/doc.go
```diff
@@ -0,0 +1,25 @@
...
+and it does not support the deletion of values.
```
arlolra: >  it does not support the deletion of values

Isn't that what updating does?

The tree does have the property of not being able to remove user leafs.

merkletree/str.go
```diff
@@ -42,8 +43,8 @@ func NewSTR(key sign.PrivateKey, policies Policies, m *MerkleTree, epoch uint64,
 	return str
 }
 
-// Serialize encodes the STR to a byte array with the following format:
-// [epoch, previous epoch, tree hash, previous STR hash, policies serialization]
+// Serialize serializes the signed tree root into
+// the correct format for signing.
```
arlolra: I think it'd be preferable if we could say s/the correct/a specified/ but guess that's alright for now.

merkletree/version.go
```diff
@@ -1,5 +1,7 @@
 package merkletree
 
 const (
-	Version = "0.1"
+	// Version indicates the current version of the library.
+	// This is matched with our release version on Github.
+	Version = "0.1.0"
```
arlolra: Hmm, this is included in the STR.  Should that really be the library version or a protocol version?

merkletree/doc.go
c633: > The tree does have the property of not being able to remove user leafs. 

Yay, I think it exactly describes what I (and @masomel ) want to say.

merkletree/version.go
c633: Oh, yep, this should be a protocol version.

c633:
> The extra explanations I was suggesting aren't supposed to be a data structure specification, but only enough background for a developer to get a better idea of what the code does and the motivation behind it. The specification would have a much greater level of detail.

Yup, it makes sense to me. I added it back to doc.go in 1c09c97c6074b71bf4835b1d58e9f27d2d717fd1.

merkletree/doc.go
```diff
@@ -6,18 +6,30 @@ to help other developers use it in their implementation easily.
...
-This module implements the Merkle prefix tree data structure. It provides methods to
-insert, update and lookup. The tree is append-only which means there is no ways to delete
-any node from the tree or change a user leaf node to an empty node.
-All the tree hash operations use the hash algorithm provided by our crypto package
+This module implements the Merkle prefix tree, which is the data structure
+underlying our PAD implementation. It is a binary tree, and that there are
+two types of leaf nodes: empty leaf node and user leaf node. Each node contains
+its prefix index and its level value. It provides methods to insert, update
+(to update an existing key-value pair) and lookup.
+The tree has the property of not being able to remove user leafs.
```
masomel: s/leafs/leaves

merkletree/pad.go
```diff
@@ -84,11 +80,16 @@ func (pad *PAD) updateInternal(policies *Policies, epoch uint64) {
...
-// Update generates a new snapshot of the directory.
+// Update generates a new snapshot of the tree.
+// It should be called in the beginning of each epoch.
 // Specifically, it extends the hash chain by issuing
 // a new signed tree root. It may remove some older signed tree roots from
 // memory if the cached PAD snapshots exceeded the maximum capacity.
-// The Policies pointer can be nil if the policies does not change.
+// policies could be nil if the PAD's policies do not change.
+// If the VRF private key is changed (by passing a new Policies),
+// the underlying tree would be reshuffled. It also means
```
masomel: s/would/will

merkletree/version.go
c633: @masomel : What do you say?

masomel: Hmm, `0.1.0` is definitely the library version, I'm not sure how to number our protocol version. I'd say that the version in the USENIX paper is the second iteration, and after working with you this summer, we've reiterated over it, so I'd say we're on `v3`.

c633: Does `3.0` sound good to you? This value is definitely the protocol version, right?

arlolra: See #121 where I moved this to the protocol package.

arlolra:
I think you should just call in `0.3` and merge it.  Don't let that be the blocker.

c633:
Merged in 18b43656842c4a8db51f3d41774b7c9abe84f1e7.

PR #105, opened 2016-10-19
protocol pkg documentation

masomel:
Part of #32

### DIFF
--- a/protocol/directory.go
+++ b/protocol/directory.go
@@ -74,13 +74,13 @@ func (d *ConiksDirectory) Update() {
 
 // SetPolicies sets this ConiksDirectory's epoch deadline and VRF
 // private key, which will be used in the next epoch.
-func (d *ConiksDirectory) SetPolicies(epDeadline merkletree.TimeStamp, vrfKey vrf.PrivateKey) {
+func (d *ConiksDirectory) SetPolicies(epDeadline merkletree.Timestamp, vrfKey vrf.PrivateKey) {
 	d.policies = merkletree.NewPolicies(epDeadline, vrfKey)
 }
 
 // EpochDeadline returns this ConiksDirectory's latest epoch deadline
 // as a timestamp.
-func (d *ConiksDirectory) EpochDeadline() merkletree.TimeStamp {
+func (d *ConiksDirectory) EpochDeadline() merkletree.Timestamp {
 	return d.pad.LatestSTR().Policies.EpochDeadline
 }
 
@@ -107,22 +107,20 @@ func (d *ConiksDirectory) NewTB(name string, key []byte) *TemporaryBinding {
 // server for logging purposes.
 //
 // A request without a username or without a public key is considered
-// malformed, and causes Register() to return a (error response,
-// ErrorMalformedClientMessage) tuple.
+// malformed, and causes Register() to return a
+// message.NewErrorResponse(ErrorMalformedClientMessage) tuple.
 // Register() inserts the new mapping in req
 // into a pending version of the directory so it can be included in the
-// snapshot taken at the end of the latest epoch, and returns a (proof
-// of absence, TB, Success) tuple if this operation succeeds.
+// snapshot taken at the end of the latest epoch, and returns a
+// message.NewRegistrationProof(ap=proof of absence, str, tb, Success) tuple
+// if this operation succeeds.
 // Otherwise, if the username already exists, Register() returns a
-// (registration proof, ErrorNameExisted)
-// tuple. The proof will be a proof of absence with a TB, if the username
-// is still pending inclusion in the next directory snapshot,
-// and a proof of inclusion it exists in the latest directory snapshot.
+// message.NewRegistrationProof(ap=proof of inclusion, str, null, ErrorNameExisted)
+// tuple. ap will be a proof of absence with a non-null TB, if the username
+// is still pending inclusion in the next directory snapshot.
+// In any case, str is the signed tree root for the latest epoch.
 // If Register() encounters an internal error at any point, it returns
-// an (error response, ErrorDirectory) tuple.
-//
-// See message.NewRegistrationProof() for details on the contents of the response
-// messages.
+// a message.NewErrorRespose(ErrorDirectory) tuple.
 func (d *ConiksDirectory) Register(req *RegistrationRequest) (
 	*Response, ErrorCode) {
 	// make sure the request is well-formed
@@ -171,21 +169,19 @@ func (d *ConiksDirectory) Register(req *RegistrationRequest) (
 // server for logging purposes.
 //
 // A request without a username is considered
-// malformed, and causes KeyLookup() to return a (error response,
-// ErrorMalformedClientMessage) tuple.
+// malformed, and causes KeyLookup() to return a
+// message.NewErrorResponse(ErrorMalformedClientMessage) tuple.
 // If the username doesn't have an entry in the latest directory
 // snapshot and also isn't pending registration (i.e. has a corresponding TB),
-// KeyLookup() returns a (proof of absence, ErrorNameNotFound) tuple.
-// Otherwise, KeyLookup() returns a (key lookup proof, Success) tuple.
-// The proof will be a proof of absence including a TB, if there is a
-// corresponding TB for the username,
-// but there isn't an entry in the directory yet, and a proof of inclusion
-// if there is an entry in the directory.
+// KeyLookup() returns a message.NewKeyLookupProof(ap=proof of absence,
+// str, null ErrorNameNotFound) tuple.
+// Otherwise, KeyLookup() returns a message.NewKeyLookupProof(ap=proof of absence, str, tb, Success)
+// tuple if, if there is a corresponding TB for the username,
+// but there isn't an entry in the directory yet, and a
+// a message.NewKeyLookupProof(ap=proof of inclusion, str, null, Success).
+// In any case, str is the signed tree root for the latest epoch.
 // If KeyLookup() encounters an internal error at any point, it returns
-// an (error response, ErrorDirectory) tuple.
-//
-// See message.NewKeyLookupProof() for details on the contents of the response
-// messages.
+// a message.NewErrorResponse(ErrorDirectory) tuple.
 func (d *ConiksDirectory) KeyLookup(req *KeyLookupRequest) (
 	*Response, ErrorCode) {
 
@@ -222,20 +218,22 @@ func (d *ConiksDirectory) KeyLookup(req *KeyLookupRequest) (
 //
 // A request without a username or with an epoch greater than the latest
 // epoch of this directory is considered malformed, and causes
-// KeyLookupInEpoch() to return a (error response,
-// ErrorMalformedClientMessage) tuple.
+// KeyLookupInEpoch() to return a
+// message.NewErrorResponse(ErrorMalformedClientMessage) tuple.
 // If the username doesn't have an entry in the directory,
 // in the directory snapshot for the indicated epoch, KeyLookupInEpoch()
-// returns a (proof of absence, ErrorNameNotFound) tuple.
-// Otherwise, KeyLookupInEpoch() returns a (proof of inclusion, Success) tuple.
+// returns a message.NewKeyLookupInEpochProof(ap=proof of absence, str,
+// ErrorNameNotFound) tuple.
+// Otherwise, KeyLookupInEpoch() returns a
+// message.NewKeyLookupInEpochProof(ap=proof of inclusion, str, Success) tuple.
+// In either case, str is a list of STRs for the epoch range [ep,
+// d.LatestSTR().Epoch], where ep is the past epoch for which
+// the client has requested the user's key.
 // KeyLookupInEpoch() proofs do not include temporary bindings since
 // the TB corresponding to a registered binding is discarded at the time
 // the binding is included in a directory snapshot.
 // If KeyLookupInEpoch() encounters an internal error at any point,
-// it returns an (error response, ErrorDirectory) tuple.
-//
-// See message.NewKeyLookupInEpochProof() for details on the contents of the response
-// messages.
+// it returns a message.NewErrorResponse(ErrorDirectory) tuple.
 func (d *ConiksDirectory) KeyLookupInEpoch(req *KeyLookupInEpochRequest) (
 	*Response, ErrorCode) {
 
@@ -275,14 +273,16 @@ func (d *ConiksDirectory) KeyLookupInEpoch(req *KeyLookupInEpochRequest) (
 //
 // A request without a username, with a start epoch greater than the
 // latest epoch of this directory, or a start epoch greater than the end epoch
-// is considered malformed, and causes Monitor() to return a (error response,
-// ErrorMalformedClientMessage) tuple.
-// Monitor() returns a (monitoring proof, Success) tuple.
+// is considered malformed, and causes Monitor() to return a
+// message.NewErrorResponse(ErrorMalformedClientMessage) tuple.
+// Monitor() returns a message.NewMonitoringProof(ap, str) tuple.
+// ap is a list of proofs of inclusion, and str is a list of STRs for the epoch
+// range [startEpoch, endEpoch], where startEpoch
+// and endEpoch are the epoch range endpoints indicated in the client's
+// request. If req.endEpoch is greater than d.LatestSTR().Epoch,
+// the end of the range will be set to d.LatestSTR().Epoch.
 // If Monitor() encounters an internal error at any point,
-// it returns an (error response, ErrorDirectory) tuple.
-//
-// See message.NewMonitoringProof() for details on the contents of the response
-// messages.
+// it returns a message.NewErrorResponse(ErrorDirectory) tuple.
 func (d *ConiksDirectory) Monitor(req *MonitoringRequest) (
 	*Response, ErrorCode) {
 
@@ -311,5 +311,5 @@ func (d *ConiksDirectory) Monitor(req *MonitoringRequest) (
 		strs = append(strs, str)
 	}
 
-	return NewMonitoringProof(aps, strs, Success)
+	return NewMonitoringProof(aps, strs)
 }
--- a/protocol/directory.go
+++ b/protocol/directory.go
@@ -89,6 +89,10 @@ func (d *ConiksDirectory) LatestSTR() *merkletree.SignedTreeRoot {
 	return d.pad.LatestSTR()
 }
 
+// NewTB creates a new temporary binding for the given name-to-key mapping.
+// NewTB computes the private index corresponding to the name, and
+// generates a digital signature of the index, the given key, and the latest STR
+// signature.
 func (d *ConiksDirectory) NewTB(name string, key []byte) *TemporaryBinding {
 	index := d.pad.Index(name)
 	return &TemporaryBinding{
--- a/protocol/directory.go
+++ b/protocol/directory.go
@@ -103,6 +103,9 @@ func (d *ConiksDirectory) NewTB(name string, key []byte) *TemporaryBinding {
 // be sent back to the client. The returned error is used by the key
 // server for logging purposes.
 //
+// A request without a username or without a public key is considered
+// malformed, and causes Register() to return a (error response,
+// ErrorMalformedClientMessage) tuple.
 // If the given username already exists in the latest snapshot of the
 // directory, Register() returns an (error response, ErrorNameExisted)
 // tuple.
@@ -159,6 +162,9 @@ func (d *ConiksDirectory) Register(req *RegistrationRequest) (
 // be sent back to the client. The returned error is used by the key
 // server for logging purposes.
 //
+// A request without a username is considered
+// malformed, and causes KeyLookup() to return a (error response,
+// ErrorMalformedClientMessage) tuple.
 // If the username doesn't have an entry in the directory and doesn't have a
 // corresponding TB, KeyLookup() returns
 // a (proof of absence, ErrorNameNotFound) tuple.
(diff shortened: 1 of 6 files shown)


masomel:
TODO:
- [ ] Add ErrorMalformedClientMessage case to all directory operations doc
- [ ] Finish message.go doc
- [ ] Minor formatting changes in error.go doc
- [ ] Add message module to doc.go

protocol/message.go
```diff
@@ -125,13 +127,15 @@ func NewErrorResponse(e ErrorCode) *Response {
...
 
-// Creates the response a CONIKS server sends to a client upon a RegistrationRequest.
+// Creates the response message a CONIKS server sends to a client upon a RegistrationRequest.
 // The contained DirectoryProof is either a proof of absence with a temporary
 // binding in the case of a successful registration (e = Success), or a proof of inclusion
 // in the case that the requested name already exists in the directory (e = ErrorNameExisted).
 // If the requested name has already been registered but not yet committed to the directory,
 // the DirectoryProof returns a proof of absence with the corresponding TB, but sets
 // the error code to ErrorNameExisted.
+// In all cases, the sserver must also include the signed tree root (STR) for the
```
c633: s/sserver/server

protocol/directory.go
```diff
@@ -31,6 +53,10 @@ func NewDirectory(epDeadline merkletree.TimeStamp, vrfKey vrf.PrivateKey,
 	return d
 }
 
+// Updates this ConiksDirectory, creating a new PAD snapshot. Update() is
+//called at the end of a CONIKS epoch. This implementation also deletes all
```
c633: nit-picky: s/called/ called.

c633:
I'm not sure how you feel about this but `golint` reports most warnings in the form `comment on exported function NewDirectory should be of the form "NewDirectory ..."`.
It is described aslo in [CoreReviewComments](https://github.com/golang/go/wiki/CodeReviewComments) 
> Comments should begin with the name of the thing being described and end in a period.

masomel:
> I'm not sure how you feel about this but golint reports all warnings in the form comment on exported function NewDirectory should be of the form "NewDirectory ...".

Thanks for pointing this out! This helps with consistent formatting, which I'm definitely in favor of :) So I'll run golint on my documentation.

protocol/error.go
```diff
@@ -26,8 +34,11 @@ const (
 	ErrorCouldNotVerify
 )
 
-// ErrorResponses contains error codes that
-// a response can omit the DirectoryResponse.
+// ErrorResponses contains the server responses indicating that
+// a client request could not be processed due to a
+// malformed client request or an internal server error.
+// Server response messages containing one of these errors therefore
+// omit the DirectoryResponse.
```
c633: I would like to change this to "ErrorResponses contains error codes indicating the client should omit the consistency checks. These errors indicate that either a client request could not be processed due to a malformed client request, an internal server error or due to a malformed server response." to make thing a bit clearer. 
Actually `ErrorResponses` is currently being used in the client side to omit the consistency check (it should include `ErrorMalformedDirectoryMessage` as well), and I think we should have a better name for this. What do you think?

masomel: > Actually ErrorResponses is currently being used in the client side to omit the consistency check

Ok. I didn't realize that this was an explicit purpose of these error responses. I'll definitely make this clear in the doc, thanks!

> it should include `ErrorMalformedDirectoryMessage` as well

Yes, and similarly, `ErrorCouldNotVerify` is another good candidate, I think.
> we should have a better name for this.

Agreed. I was thinking about this earlier and *all* possible responses (except passed and success) start with the "Error-" prefix, but only a handful of them are actually treated as errors. Maybe it makes sense to treat the non-errors more as different results, e.g. change the prefix to something like "Req-" for all request results (including Success?) and "Check-" for all consistency check results? So we'd be dealing with error codes such as "ReqNameNotFound" or "CheckBadCommitment". This also makes it clearer on which end of the protocol each "error" is returned. What do you think?

c633: > ErrorCouldNotVerify is another good candidate

It would be removed in https://github.com/coniks-sys/coniks-go/pull/113/files#diff-78429462c41fd4a6aab3cc13baa05919L28. 

> Maybe it makes sense to treat the non-errors more as different results, e.g. change the prefix to something like "Req-" for all request results (including Success?) and "Check-" for all consistency check results? So we'd be dealing with error codes such as "ReqNameNotFound" or "CheckBadCommitment". 

Yes, it makes sense to me, maybe we can keep `Error-` prefix for error codes in `ErrorResponses`.

masomel: > we can keep Error- prefix for error codes in ErrorResponses.

Yes, definitely, sorry if this wasn't clear. My idea was to distinguish between actual errors and results.

c633: I will submit a pull after this is merged

masomel: I just created issue #114 for this, and I think this is something we should push back to our next release. This pull is just for the documentation, so I think this pull should be merged after all other protocol-related pulls are merged. What do you think of the following merge order: #74, #113, #105, #114?

c633: SGTM. I'm waiting for others' review on #74. :)

protocol/directory.go
```diff
@@ -8,13 +15,28 @@ import (
...
+// of temporary bindings (TBs). This feature will be split into a separate
+// protocol extension in a future release.
 type ConiksDirectory struct {
 	pad      *merkletree.PAD
 	useTBs   bool
 	tbs      map[string]*merkletree.TemporaryBinding
 	policies *merkletree.Policies
 }
 
+// NewDirectory Constructs a new ConiksDirectory given the key server's PAD
```
arlolra: lowercase "constructs"

protocol/directory.go
```diff
@@ -39,18 +65,43 @@ func (d *ConiksDirectory) Update() {
...
 	d.policies = merkletree.NewPolicies(epDeadline, vrfKey)
 }
 
+// EpochDeadline returns this ConiksDirectory's current epoch deadline
+// as a timestamp.
 func (d *ConiksDirectory) EpochDeadline() merkletree.TimeStamp {
 	return d.pad.LatestSTR().Policies.EpochDeadline
 }
 
+// LatestSTR Returns this ConiksDirectory's latest STR.
```
arlolra: In the `EpochDeadline()` comment, it says "current" and here "latest".  We should probably just stick to one.

protocol/directory.go
```diff
@@ -39,18 +65,43 @@ func (d *ConiksDirectory) Update() {
...
+// (response, error).
+// The response (which also includes the error code) is supposed to
+// be sent back to the client. The returned error is used by the key
+// server for logging purposes.
+//
+// A request without a username or without a public key is considered
+// malformed, and causes Register() to return a (error response,
+// ErrorMalformedClientMessage) tuple.
+// If the given username already exists in the latest snapshot of the
+// directory, Register() returns an (error response, ErrorNameExisted)
```
arlolra: Should this be `(registration proof, ErrorNameExisted)`?

protocol/directory.go
```diff
@@ -83,14 +134,35 @@ func (d *ConiksDirectory) Register(req *RegistrationRequest) (
...
 	}
+
+	if err = d.pad.Set(req.Username, req.Key); err != nil {
+		return NewErrorResponse(ErrorDirectory), ErrorDirectory
+	}
+	return NewRegistrationProof(ap, d.LatestSTR(), nil, Success)
 }
 
+// KeyLookup gets the public key for the username indicated in the
+// KeyLookupRequest req received from a CONIKS client from the latest
```
arlolra: Again, "latest snapshot" or "current epoch"

protocol/directory.go
```diff
@@ -117,6 +189,25 @@ func (d *ConiksDirectory) KeyLookup(req *KeyLookupRequest) (
...
 
+// KeyLookupInEpoch gets the public key for the username for a prior
+// epoch in the directory history indicated in the
+// KeyLookupInEpochRequest req received from a CONIKS client,
+// and returns a tuple of the form (response, error).
+// The response (which also includes the error code) is supposed to
+// be sent back to the client. The returned error is used by the key
+// server for logging purposes.
+//
+// A request without a username or with an epoch greater than the latest
```
arlolra: Maybe we use "latest" more often.

protocol/directory.go
```diff
@@ -146,6 +237,21 @@ func (d *ConiksDirectory) KeyLookupInEpoch(req *KeyLookupInEpochRequest) (
...
+// Monitor gets the directory proofs for the username for the range of
+// epochs indicated in the MonitoringRequest req received from a
+// CONIKS client,
+// and returns a tuple of the form (response, error).
+// The response (which also includes the error code) is supposed to
+// be sent back to the client. The returned error is used by the key
+// server for logging purposes.
+//
+// A request without a username, with a start epoch greater than the
+// latest epoch of this directory, or a start epoch greater than the end epoch
```
arlolra: Yeah, "latest"

arlolra:
Overall, looks pretty good.  Two general comments is the consistency of "latest (snapshot)?" vs "current epoch".  And, I'm not sure if godoc recognizes it, but wrapping argument names in backticks, `likeThis`.

protocol/directory.go
masomel: Good point. I've noticed that we're using the two interchangeably and that could be confusing. I think "current" makes more sense if we consider that when the server updates the directory and increments the epoch counter to `t`, the system has "entered" epoch `t` and remains in epoch `t` until the next directory update. Might even be worth changing the code to say `CurrentSTR()` etc. where "latest" is used right now. What do you think? cc @c633 @Liamsi

protocol/directory.go
masomel: It should, thanks.

protocol/directory.go
```diff
@@ -117,6 +189,25 @@ func (d *ConiksDirectory) KeyLookup(req *KeyLookupRequest) (
...
+// The response (which also includes the error code) is supposed to
+// be sent back to the client. The returned error is used by the key
+// server for logging purposes.
+//
+// A request without a username or with an epoch greater than the latest
+// epoch of this directory is considered malformed, and causes
+// KeyLookupInEpoch() to return a (error response,
+// ErrorMalformedClientMessage) tuple.
+// If the username doesn't have an entry in the directory,
+// at the indicated snapshot, KeyLookupInEpoch() returns a (KeyLookupInEpoch
```
masomel: "Epoch" works, though, I was trying to use PAD-terminology here, since `KeyLookupInEpoch()` is really looking in the snapshot for the indicated epoch. I'll clarify this.

protocol/directory.go
c633: >  the system has "entered" epoch t and remains in epoch t until the next directory update

I think it should be `t+1`. Since it is *signed* tree root and in "current" epoch `t+1`, the tree root haven't been signed yet, "latest" makes more sense to me.

c633:
>  And, I'm not sure if godoc recognizes it, but wrapping argument names in backticks, likeThis.

Unfortunately, godoc doesn't recognize backticks format :(

masomel:
>  Two general comments is the consistency of "latest (snapshot)?" vs "current epoch".

Yeah, I actually like "current" better; it makes more sense to me to think of an epoch as a time interval instead of a discreet point in time. 

> Unfortunately, godoc doesn't recognize backticks format :(

Right, godoc is unfortunately very limited in the text formatting it allows. It does recognize code text, but only by writing it in its own line, indented, which isn't what we want in most cases.

protocol/directory.go
masomel: But lookups in the "current" epoch if "current" is `t+1` search the tree for `t` *because* the tree root hasn't been signed yet. So why doesn't it make sense to say that `t` is the current epoch if `t+1` doesn't exist and hasn't been signed yet?

c633: > But lookups in the "current" epoch if "current" is t+1 search the tree for t

Then maybe we should use lookups in the "latest" epoch instead. I think "latest" clearly indicates that we actually perform all operations on the tree which has been signed, whereas "current" makes me a little confuse.

masomel: > I think "latest" clearly indicates that we actually perform all operations on the tree which has been signed

This makes sense to me. My argument for "current" is that I think captures the user's intent better, since we don't expect user's to really have a notion of the underlying PAD. But if the three of you feel stronger about using "latest", let's go with that.

protocol/directory.go
```diff
@@ -39,18 +65,43 @@ func (d *ConiksDirectory) Update() {
...
+// ErrorMalformedClientMessage) tuple.
+// If the given username already exists in the latest snapshot of the
+// directory, Register() returns an (error response, ErrorNameExisted)
+// tuple.
+// Otherwise, Register() inserts the new mapping in req
+// into the PAD so it can be included in the snapshot taken at the end
+// of the current  epoch), and returns a (registration proof, Success)
+// if this operation succeeds.
+// If Register() encounters an internal error at any point, it returns
+// an (error response, ErrorDirectory) tuple.
```
masomel: No, it does assume that `useTBs` is true. I've added in the TB cases to make this explanation more complete.

masomel:
Any more thoughts on this before this gets merged? @arlolra @Liamsi?

protocol/doc.go
```diff
@@ -0,0 +1,40 @@
...
+
+protocol implements the server- and client-side components of the CONIKS key
+management and verification protocols. More specifically, protocol provides
+an API for maintaining an auditable, privacy-preserving key directory on a
+server, as well as an API for checking the consistency of the directory
+at the client.
+
+Consistency Checks
+
+This module implements all consistency checks perfomed by a CONIKS client
```
Liamsi: perfomed -> performed

protocol/doc.go
```diff
@@ -0,0 +1,40 @@
...
+privacy-preserving data structure that contains mappings from usernames
+to public keys. It currently supports registration of new mappings,
+latest-version key lookups, historical key lookups, and monitoring of mappings.
+
+Error
+
+This module defines the constants representing the types
+of errors that a CONIKS server may return to a client,
+and the results of a consistency check or cryptographic verification
+that a CONIKS client performs.
```
Liamsi: "[...] and the results of a consistency check or a cryptographic verification that a CONIKS client performs."

protocol/error.go
```diff
@@ -1,10 +1,15 @@
 // Defines constants representing the types
-// of errors that the server may return to a client.
+// of errors that the server may return to a client,
+// and that the client may throw after a consistency
+// check or cryptographic verification
```
Liamsi: "[...] or *a* cryptographic verification [...]" (?)

protocol/error.go
```diff
@@ -1,10 +1,15 @@
...
 
 package protocol
 
+// An ErrorCode is a special int.
```
Liamsi: The most interesting fact about `ErrorCode` isn't that it is internally an `int`, but rather that it is a integer representation of (specific) error-messages (those that travel through the wire??) 
Also, it's interesting (for an overview) that they can be mapped to human readable string representations.

protocol/doc.go
Liamsi: You are the native speaker :-) Feel free to ignore my comments...

protocol/error.go
masomel: Makes sense. I wasn't sure exactly what to write here. Thanks!

protocol/doc.go
```diff
@@ -0,0 +1,40 @@
...
+
+Message
+
+This module defines the message format of the CONIKS client requests
+and corresponding CONIKS server responses for each CONIKS protocol.
+It also provides constructors for the response messages for each
+protocol.
```
Liamsi: I just remembered that the paper also speaks of these "things", what we started calling operations, as separate protocols. So it makes sense to stay consistent, here, too.

protocol/error.go
c633: > they can be mapped to human readable string representations.

I would like to add something like "... it implements the built-in error interfaces".

protocol/message.go
```diff
@@ -1,87 +1,152 @@
+// Defines the message formats of the CONIKS protocols
```
Liamsi: Nitpicking: should be consistent with the description in doc.go "message format" (singular or plural)

protocol/message.go
```diff
@@ -118,6 +201,12 @@ func NewKeyLookupInEpochProof(ap *m.AuthenticationPath,
 	}, e
 }
 
+// NewMonitoringProof creates the response message a CONIKS server sends
+// to a client upon a MonitoringRequest. The contained DirectoryProofs
+// includes a list of proofs of inclusion and a list
+// of STRs for the epoch range [startEpoch, endEpoch], where startEpoch
+// and endEpoch are the epoch range endpoints indicated in the client's
+// request. The error code is set to Success.
```
Liamsi: "The error code is set to Success." <- That's confusing. But also in combination of what the constructor is doing: One can pass it an `ErrorCode` as a parameter (can be `Success` or anything else). This `ErrorCode` `e` is incorporated into the freshly created Response (OK) *and* returned (why)?

protocol/directory.go
Liamsi: I OK with either "current" or "latest" as long as it used consistently and we somewhere clearly state what it means (the lastest STR is the one that is ...)

protocol/directory.go
```diff
@@ -39,18 +66,48 @@ func (d *ConiksDirectory) Update() {
...
 func (d *ConiksDirectory) EpochDeadline() merkletree.TimeStamp {
 	return d.pad.LatestSTR().Policies.EpochDeadline
 }
 
+// LatestSTR Returns this ConiksDirectory's latest STR.
 func (d *ConiksDirectory) LatestSTR() *merkletree.SignedTreeRoot {
 	return d.pad.LatestSTR()
 }
 
+// Register inserts the username-to-key mapping contained in a
```
Liamsi: Super nitpicky: We use "key" for many different concepts. Maybe, public-key is better suited here?

protocol/message.go
```diff
@@ -1,87 +1,152 @@
...
+// The contained DirectoryProof is either a proof of absence with a temporary
+// binding in the case of a successful registration (e = Success), or a
+// proof of inclusion in the case that the requested name already exists
+// in the directory (e = ErrorNameExisted).
+// If the requested name has already been registered but not yet
+// committed to the directory,
+// the DirectoryProof returns a proof of absence with the corresponding
+// TB, but sets the error code to ErrorNameExisted.
+// In all cases, the sserver must also include the signed tree root
+// (STR) for the lastest epoch.
```
Liamsi: lastest -> latest

protocol/message.go
Liamsi: And just in case this isn't intentional: this comment won't show up in the rendered godoc documentation.

protocol/message.go
```diff
@@ -106,6 +181,14 @@ func NewKeyLookupProof(ap *m.AuthenticationPath, str *m.SignedTreeRoot,
...
 }
 
+// NewKeyLookupInEpochProof creates the response message a CONIKS server
+// sends to a client upon a KeyLookupInEpochRequest.
+// The contained DirectoryProofs is either a single proof of inclusion
+// (e = Success), or a proof of absence if the requested name does not
+// exist in the directory (e = ErrorNameNotFound).
+// In either case, the server returns a list of STRs for the epoch range [ep,
+// directory.LatestEpoch().Epoch], where ep is the past epoch for which
+// the client has requested the user's key.
```
Liamsi: Thinking aloud here: I wonder if it wouldn't make more sense to move the actual behaviour of the server to methods where it actually happens (for instance `KeyLookupInEpoch` creates the list of STRs which are, then, passed to this constructor)

Liamsi:
Some minor comments, questions, and nitpicks/typos. In overall: looks good to me!

protocol/doc.go
masomel: Yeah, that's the terminology that I've been trying to use consistently :)

protocol/message.go
masomel: > "The error code is set to Success." <- That's confusing. But also in combination of what the constructor is doing: One can pass it an ErrorCode as a parameter (can be Success or anything else)

If you look at `directory.Monitor()`, which is the only place where `NewMonitoringProof` is called, the error code is set to `Success`, so that's what I was referring to. But you're definitely right that `ErrorCode` *could* be set to anything. I think it would make sense to changing the signature of the constructor and to set the `ErrorCode` internally to `Success` if that's the only "error" the constructor can ever return.

> This ErrorCode e is incorporated into the freshly created Response (OK) and returned (why)?

I had the same question a while back. The error is returned so it can be propagated up to the server for logging: https://github.com/coniks-sys/coniks-go/pull/71#r75061038

protocol/message.go
masomel: I'm a bit confused, what do you mean by "behavior of the server"? Are you suggesting `KeyLookupInEpoch` shouldn't generate the list of STRs but that this should be done in the server? The list of STRs *is* part of the `KeyLookupInEpoch` protocol, though. Maybe my wording and use of the term "server" here is misleading?

Liamsi: Oh sorry, reading my comment, I was confused myself... Sorry!

What I meant is that *descriptions* of behaviour of the server should be moved where this behaviour happens; for instance, comments like the above: "In either case, the server returns a list of STRs for the epoch range" should appear above methods where the STRs are actually generated (the comment of `KeyLookupInEpoch` should talk about the STRs and not the necessarily this constructor where they are just passed as an argument and aren't generated/returned).

Does this make more sense?

protocol/message.go
Liamsi: > I think it would make sense to change the signature of the constructor and to set the ErrorCode internally to Success if that's the only "error" the constructor can ever return.

Yes, I agree.

masomel: Cool, I'll go ahead and make those tweaks, too.

c633:
Since we have moved `tb.go` to `protocol/`, this doc should cover the temporary binding documentation as well. Do you want me to do a copy, @masomel ?

masomel:
> Since we have moved tb.go to protocol/, this doc should cover the temporary binding documentation as well. Do you want me to do a copy, @masomel ?

I'll do a rebase. I need to the go through some of the descriptions again anyway, so I'll make sure to check that they still make sense with the new changes.

PR #109, opened 2016-10-24
Fix bug: Registration collision in the same epoch when the keyserver doesn't use TBs

c633:
- Add `LookupInCurrentEpoch` to lookup in the current tree 
- Rename `Lookup` to `LookupInLatestEpoch`
- See directory_test.go (TestRegisterExistedUserWithoutTB) for demostration scenario.

### DIFF
--- a/merkletree/pad.go
+++ b/merkletree/pad.go
@@ -95,10 +95,24 @@ func (pad *PAD) Set(name string, value []byte) error {
 	return pad.tree.Set(index, name, value)
 }
 
-func (pad *PAD) Lookup(name string) (*AuthenticationPath, error) {
+// LookupInCurrentEpoch searchs the requested name in the current tree
+// (which is not committed to the PAD).
+// The returned AuthenticationPath should be used only to verify the inclusion
+// of the requested name in the server-side
+// and should **not** be used to return to the client.
+func (pad *PAD) LookupInCurrentEpoch(name string) (*AuthenticationPath, error) {
+	lookupIndex, proof := pad.computePrivateIndex(name, pad.policies.vrfPrivateKey)
+	ap := pad.tree.Get(lookupIndex)
+	ap.VrfProof = proof
+	return ap, nil
+}
+
+// LookupInLatestEpoch searchs the requested name in the latest version of the tree.
+func (pad *PAD) LookupInLatestEpoch(name string) (*AuthenticationPath, error) {
 	return pad.LookupInEpoch(name, pad.latestSTR.Epoch)
 }
 
+// LookupInEpoch searchs the requested name in the tree at requested epoch.
 func (pad *PAD) LookupInEpoch(name string, epoch uint64) (*AuthenticationPath, error) {
 	str := pad.GetSTR(epoch)
 	if str == nil {
--- a/protocol/directory.go
+++ b/protocol/directory.go
@@ -88,7 +88,7 @@ func (d *ConiksDirectory) Register(req *RegistrationRequest) (
 	*Response, ErrorCode) {
 	// check whether the name already exists
 	// in the directory before we register
-	ap, err := d.pad.Lookup(req.Username)
+	ap, err := d.pad.LookupInLatestEpoch(req.Username)
 	if err != nil {
 		return NewErrorResponse(ErrorDirectory), ErrorDirectory
 	}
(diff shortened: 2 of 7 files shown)


protocol/directory.go
```diff
@@ -110,6 +110,12 @@ func (d *ConiksDirectory) Register(req *RegistrationRequest) (
 		d.tbs[req.Username] = tb
 		return NewRegistrationProof(ap, d.LatestSTR(), tb, Success)
 	} else {
+		currentAP, _ := d.pad.LookupInCurrentEpoch(req.Username)
+		if bytes.Equal(currentAP.LookupIndex, currentAP.Leaf.Index) {
+			// use the ap from the latest STR to return to the client
+			return NewRegistrationProof(ap, d.LatestSTR(), nil, ErrorNameExisted)
```
masomel: My only issue with this is that the client doesn't get any proof that the name already exists, and until the current tree is committed, I don't know if we should be relying on it to give us information since a malicious server can now easily cheat and give us false information about which names exist, don't exist etc and we wouldn't have any proof. I'm starting to think that TBs shouldn't be just an extension :( At least while we don't have a better solution for signed promises.

merkletree/pad.go
```diff
@@ -95,10 +95,24 @@ func (pad *PAD) Set(name string, value []byte) error {
 	return pad.tree.Set(index, name, value)
 }
 
-func (pad *PAD) Lookup(name string) (*AuthenticationPath, error) {
+// LookupInCurrentEpoch searchs the requested name in the current tree
+// (which is not committed to the PAD).
+// The returned AuthenticationPath should be used only to verify the inclusion
+// the of requested name in the server-side
+// and should **not** be used to return to the client.
+func (pad *PAD) LookupInCurrentEpoch(name string) (*AuthenticationPath, error) {
```
c633: > and I'm afraid that a developer will misuse this function (accidentally or intentionally) despite your description in the documentation.

I feel that too.

> Is there an alternative way we would keep track of the names that have been registered in the current epoch but aren't committed yet (e.g. in an array)?

Let's see what @arlolra  and @Liamsi  say. I'm not so sure about this :(

> Or could merkletree.Set() return an error if we attempt to add a private index that already exists?

Actually I think I want to keep it for key change operation, and we should just do this in PAD. What do you think?

protocol/directory.go
c633: > I don't know if we should be relying on it to give us information since a malicious server can now easily cheat and give us false information about which names exist, don't exist etc and we wouldn't have any proof.

That makes a lot of sense. It seems like we just mitigated a bug and open other security issues :(

merkletree/pad.go
masomel: >> Or could merkletree.Set() return an error if we attempt to add a private index that already exists?
>
> Actually I think I want to keep it for key change operation, and we should just do this in PAD. What do you think? 

Oh yes, you're right. `pad.Set()` is a better place for putting in this check.

protocol/directory.go
masomel: > That makes a lot of sense. It seems like we just mitigated a bug and open other security issues :(

Hmm yeah. That's why I'm thinking the solution may actually be to force the use of TBs, no matter what. What do you think?

c633: It seems like we have no choices so I agree.

masomel: @Liamsi @arlolra What are your thoughts on this?

merkletree/pad.go
```diff
@@ -95,10 +95,24 @@ func (pad *PAD) Set(name string, value []byte) error {
 	return pad.tree.Set(index, name, value)
 }
 
-func (pad *PAD) Lookup(name string) (*AuthenticationPath, error) {
+// LookupInCurrentEpoch searchs the requested name in the current tree
+// (which is not committed to the PAD).
+// The returned AuthenticationPath should be used only to verify the inclusion
+// the of requested name in the server-side
```
arlolra: s/the of/of the/

protocol/directory.go
arlolra: It's probably fine to continue to implement TBs as an extension (since we can still imagine other solutions) but, for the moment, we should panic if `useTBs` is `false`.

arlolra: We should probably create an issue w/ the recommended changes, referencing the pull, first.

c633:
Opened #110 for this.

masomel:
This pull can be closed since we have #111, right?

Issue #110, opened 2016-10-25
Re-register in the same epoch attack


c633:
Currently, if the key server **doesn't** use TBs, a malicious client can attempt to equivocate about the name  by re-registering the same username with a different key.
This attack can be mitigated by lookup in the current tree of the pad. The server could return a `ErrorNameExisted` along with a proof of absence (from the latest committed STR). 

However, the client doesn't get any proof that the name already exists, and until the current tree is committed. A malicious server can now easily cheat and give clients false information about which names exist, don't exist etc and they wouldn't have any proof.

It means that while we don't have a better solution for registration promises, the key server would be forced to use TBs.

See #109.

PR #108, opened 2016-10-23
Quickfix: Ensure start epoch of LookupInEpochRequest and MonitoringRequest is in PAD

masomel:
Otherwise directory.LookupInEpoch() fails with ErrorDirectory, though a start epoch that is > the latest epoch indicates a malformed client message.

Issue #89, opened 2016-09-22
Create a dedicated protocol/extensions directory


c633:
From our email discussion:
> Right now, the TB extension is very much embedded into the main data structure, but I think it would be more general (and maybe clearer?) to be able to just run `coniksserver run --with-tb --with-cosi` and for the server to plug in the indicated extensions at run-time.

Liamsi:
This might be related: https://tip.golang.org/pkg/plugin/ (investigate if protocol extensions could be build as optional plugins)

c633:
Thanks for this pointer. I took another approach in https://github.com/coniks-sys/coniks-go/commit/ab7952c8b7d7d578c4f42b16f6644d08ad883e1d, but I don't know which one is better.

c633:
Trillian starts rolling out extensions in https://github.com/google/trillian/pull/271, and their approach is the same as ours. 👍

PR #111, opened 2016-10-25
Panic if the server does not use TBs

c633:
Workaround for #110

### DIFF
--- a/protocol/directory_test.go
+++ b/protocol/directory_test.go
@@ -79,7 +79,7 @@ func TestRegisterExistedUserWithTB(t *testing.T) {
 	}
 }
 
-func TestRegisterWithoutTB(t *testing.T) {
+func TestNewDirectoryPanicWithoutTB(t *testing.T) {
 	// workaround for #110
 	defer func() {
 		if r := recover(); r == nil {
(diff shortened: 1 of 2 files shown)


masomel:
Merge this after #108 is merged?

c633:
> LGTM. Just to make sure, the keyserver has useTBs set to true everywhere?

Yes (https://github.com/coniks-sys/coniks-go/blob/master/keyserver/server.go#L112).

> Merge this after #108 is merged?

I think we can merge this and keep #108 opening until we address #110? Is that what @arlolra said?

masomel:
Should we merge this? @arlolra @Liamsi any thoughts before this is merged?

Liamsi:
> Should we merge this? @arlolra @Liamsi any thoughts before this is merged?

First, the merge conflicts need to be resolved.

protocol/directory_test.go
```diff
@@ -80,6 +80,13 @@ func TestRegisterExistedUserWithTB(t *testing.T) {
 }
 
 func TestRegisterWithoutTB(t *testing.T) {
+	// workaround for #110
```
arlolra: I'm wondering how quickly these `WithoutTB` tests are going to rot?

Liamsi:
This might be a little to late, but wouldn't it be more consequent to make TBs non-optional (and completely remove `useTBs` instead of panicking until they are really optional)?

arlolra:
Probably, but TBs were a good candidate to flesh out #89

protocol/directory_test.go
c633: Sorry, I didn't really understand your question. These tests should panic and rot right after directory initialization. Should we remove these tests?

arlolra: https://en.wikipedia.org/wiki/Software_rot

arlolra: I mean that since we're panicking early, the rest of the code won't be executed and will likely accumulate bugs or otherwise be unuseful over time.

arlolra: > Yeah, thanks for teaching me this jargon.

Sorry, I should be more careful with idioms.

> From my side, I think we would never run a key server without signed promises, so do other developers. Thus, removing these tests and keeping useTBs until we address #89 seem fine to me.

I'm ok with this reasoning.

c633: Sounds fine to me :)

arlolra:
LGTM

c633:
@Liamsi What do you think about this pull?

masomel:
LGTM with the truncated test.

PR #112, opened 2016-10-25
Make directory.HandleOps() unexported method of keyserver/listener.go

masomel:
While writing the API documentation for the protocol package, I thought that HandleOps is really a server-level function. This change makes the directory API strictly about operations on the data structure.


c633:
The main reason of having `HandleOps` is to avoid the developers mistakenly call the handlers with a bad request (think about what we did before #108). I'm still think that these validations should belong to the protocol.

masomel:
> The main reason of having `HandleOps` is to avoid the developers mistakenly call the handlers with a bad request (think about what we did before #108)

We definitely want to avoid developers misusing the handlers, but a `HandleOps` function that determines which handler to call based on the request type is really a server-level function since the directory is just a high-level data structure. If we want to validate the requests, why can't each handler fail early if the request is malformed? Then we can call `handleOps` from the server, which decides which handler to call.

c633:
Yeah, we can do that way. LGTM.

keyserver/listener.go
```diff
@@ -86,6 +86,31 @@ func malformedClientMsg(err error) ([]byte, error) {
 	return res, err
 }
 
+// andleOps validates the request message and then pass it to
```
c633: s/andleOps/handleOps

masomel:
> Can we make `handleOps` method of `ConiksServer`? I gave this a shot in 227a2ab.

Sure, this LGTM! We can cherry-pick 227a2ab into this branch, not sure if that would be the easiest way to incorporate your changes.

c633:
Yes, you can cherry pick.

PR #117, opened 2016-11-10
Add API documentation to bots

masomel:
- [x] bot.go
- [x] twitterbot.go
- [x] cmd subpkg
- [x] doc.go
- [x] Move account verification protocol spec to coniks-spec doc


masomel:
@c633 The `BotHandleFunc` type on https://github.com/coniks-sys/coniks-go/blob/master/bots/bot.go#L13 seems to be unused. Should I unexport it, or can I remove it?

c633:
Sorry, please remove it.

masomel:
@c633 Sorry another question about the implementation: I noticed that the twitterbot stores the `coniksAddress` and the `handle` in two different structs (`TwitterBot` and `TwitterConfig`). Is there any reason why `TwitterBot` can't just have a `TwitterConfig` field?

c633:
>  I wanted to include a description of these fields in the documentation, but it seems a bit odd to write that those two values are stored in two places.

I think we probably shouldn't write document for internal implementation, since it can be changed in the future, right?

> Is there any reason why TwitterBot can't just have a TwitterConfig field? 

Hmm, the reason is maybe I wanted to keep a copy of necessary fields of `TwitterConfig` only, so that the developer can't change them after initialization. Does this make sense?

masomel:
> I think we probably shouldn't write document for internal implementation, since it can be changed in the future, right?

A developer should have an idea of the information that a struct maintains. If the implementation is changed in the future, the documentation should change along with it. Documenting a struct with unexported fields isn't all that different than documenting a method that calls unexported methods, is it? But `TwitterBot` shouldn't be an unexported struct, right?

> the reason is maybe I wanted to keep a copy of necessary fields of TwitterConfig only, so that the developer can't change them after initialization. Does this make sense? 

Yeah, this makes sense. I'm mostly wary of keeping multiple copies of the same data around because consistency issues may arise, but I think it's definitely fine to leave this as is for now.

arlolra:
LGTM, but hopefully the documentation being removed here is preserved somewhere else?

c633:
@masomel moved it to https://github.com/coniks-sys/coniks-spec/blob/acct-verif-spec/account-verification.txt.

bots/twitterbot.go
```diff
@@ -23,19 +33,36 @@ type TwitterBot struct {
 
 var _ Bot = (*TwitterBot)(nil)
 
+// A TwitterConfig contains the address of the named UNIX socket
+// through which the bot and the CONIKS server communicate,
+// the OAuth information needed to authenticate the bot with Twitter,
+// and the bot's reserved Twitter handle. These values are specified
+// in a configuration file, which is read at initialization time.
```
c633: Maybe make this clearer by adding ", which is in the TOML (Tom's Obvious, Minimal Language) format."?

masomel:
> LGTM, but hopefully the documentation being removed here is preserved somewhere else?

Here's the commit: https://github.com/coniks-sys/coniks-spec/commit/2339272f2e991c927398f047d23786adbed7fe13

PR #118, opened 2016-11-11
Move TBs to protocol/

arlolra:
Guess this has something to do w/ #89


c633:
Yeah, I would do the same thing for #89. I like the `pad.Sign` and it LTGM!

merkletree/pad.go
```diff
@@ -121,11 +122,13 @@ func (pad *PAD) LatestSTR() *SignedTreeRoot {
 	return pad.latestSTR
 }
 
-func (pad *PAD) TB(name string, value []byte) (*TemporaryBinding, error) {
+func (pad *PAD) Sign(msg ...[]byte) []byte {
+	return pad.signKey.Sign(bytes.Join(msg, nil))
+}
+
+func (pad *PAD) NextIndex(name string) []byte {
```
c633: Can you add a comment for this method? Its name isn't really clear to me.

arlolra: Yeah, I wasn't sure exactly where we landed on the discussion in https://github.com/coniks-sys/coniks-go/pull/105#discussion_r86062465

There seems to have been an agreement on "latest" being the most recently signed epoch.  Do we like "current" as the one we're presently accepting registrations for?  Or, as I used, "next"?

I suppose this prefix needs to apply to the signature as well.

c633: Or we can remove the prefix, just like `pad.Set`? Then we may need to rename `Lookup` to something like `LookupInLatestEpoch`.

arlolra: Alright, I dropped the prefix.

arlolra:
@c633 If you're ok with this, feel bold about merging.

c633:
Merged in fa21b15cab82a86c6ddb7302f57bb1242f13dedd. Thanks @arlolra !

Issue #120, opened 2016-11-14
Move policies to protocol?


arlolra:
The policies are protocol specific, and don't really have anything to do with the underlying data structures, do they?

Further, it seems undesirable to be *storing* the VRF key in the policies, and relying on only extracting the public key when serializing.  In practice, that means many copies of secret material stored in memory / on disk.

What's worse is that usually rotating keys means that you can safely destroy the previous key material.  However, in order to support lookups in previous epochs, we need to keep those keys around indefinitely.

For now, I think maybe the PAD struct should gain an `vrfKey vrf.PrivateKey` field and policies should move to protocol, and become an opaque blob (`[]byte`) in the PAD.

Somewhat related to #47

c633:
Aligning VRF key and signing key makes a lot of sense to me, but how will `vrf.PrivateKey`s be handled in protocol?

arlolra:
> how will vrf.PrivateKeys be handled in protocol?

Can you clarify what you're asking?  For now, we won't have a mechanism to rotate the vrf key.

c633:
> For now, we won't have a mechanism to rotate the vrf key.

OK. My question was about how would we store vrf keys in protocol, but your statement answered it.

arlolra:
Yeah, see the third paragraph about what rotating would even mean, given the consequences.

c633:
Since we'll also support prior history verification in #106, it probably means we also have to support key lookup in epoch without time limitation?

masomel:
> probably means we also have to support key lookup in epoch without time limitation?

What do you mean by "without time limitation"?

c633:
I meant something like "you are not allowed to look up for key older than 1 month". I remember that is what we discussed in our previous meeting, right?

masomel:
> I meant something like "you are not allowed to look up for key older than 1 month". I remember that is what we discussed in our previous meeting, right?

Gotcha. Yeah, I wasn't sure if we had come to an agreement during our last meeting since we haven't figured out a scheme for rotating the VRF key. But I think you're right, that if we support prior lookup history, lookups at any time in the past should also be possible.

c633:
> I wasn't sure if we had come to an agreement during our last meeting

I think we haven't it yet.

c633:
A possible way is using an HKDF, if we support lookups at any time in the past:
- The server generates a `MasterVRFKey`
[...]

PR #121, opened 2016-11-15
Move policies to protocol

arlolra:
See #120 

This needs a sanity check.

### DIFF
--- a/merkletree/pad.go
+++ b/merkletree/pad.go
@@ -2,7 +2,6 @@ package merkletree
 
 import (
 	"bytes"
-	"crypto/subtle"
 	"errors"
 
 	"github.com/coniks-sys/coniks-go/crypto"
@@ -19,27 +18,29 @@ var (
 // PAD is an acronym for persistent authenticated dictionary
 type PAD struct {
 	signKey      sign.PrivateKey
+	vrfKey       vrf.PrivateKey
 	tree         *MerkleTree // will be used to create the next STR
 	snapshots    map[uint64]*SignedTreeRoot
 	loadedEpochs []uint64 // slice of epochs in snapshots
 	latestSTR    *SignedTreeRoot
-	policies     Policies // the current policies in place
+	policies     []byte // the current policies in place
 }
 
 // NewPAD creates new PAD consisting of an array of hash chain
 // indexed by the epoch and its maximum length is len.
-func NewPAD(policies *Policies, signKey sign.PrivateKey, len uint64) (*PAD, error) {
+func NewPAD(policies []byte, signKey sign.PrivateKey, vrfKey vrf.PrivateKey, len uint64) (*PAD, error) {
 	if policies == nil {
 		panic("[merkletree] PAD must be created with a non-NULL Policies struct")
 	}
 	var err error
 	pad := new(PAD)
 	pad.signKey = signKey
+	pad.vrfKey = vrfKey
 	pad.tree, err = NewMerkleTree()
 	if err != nil {
 		return nil, err
 	}
-	pad.policies = *policies
+	pad.policies = policies
 	pad.snapshots = make(map[uint64]*SignedTreeRoot, len)
 	pad.loadedEpochs = make([]uint64, 0, len)
 	pad.updateInternal(nil, 0)
@@ -63,20 +64,14 @@ func (pad *PAD) signTreeRoot(epoch uint64) {
 	pad.latestSTR = NewSTR(pad.signKey, pad.policies, m, epoch, prevHash)
 }
 
-func (pad *PAD) updateInternal(policies *Policies, epoch uint64) {
+func (pad *PAD) updateInternal(policies []byte, epoch uint64) {
 	// create STR with the policies that were actually used in the prev.
 	// Set() operation
 	pad.signTreeRoot(epoch)
 	pad.snapshots[epoch] = pad.latestSTR
 	pad.loadedEpochs = append(pad.loadedEpochs, epoch)
 	if policies != nil { // update the policies if necessary
-		vrfKeyChanged := 1 != subtle.ConstantTimeCompare(
-			pad.policies.vrfPrivateKey,
-			policies.vrfPrivateKey)
-		pad.policies = *policies
-		if vrfKeyChanged {
-			pad.reshuffle()
-		}
+		pad.policies = policies
 	}
 }
 
@@ -90,7 +85,7 @@ func (pad *PAD) updateInternal(policies *Policies, epoch uint64) {
 // the underlying tree would be reshuffled. It also means
 // the private index of all new key-to-value bindings
 // will be computed using the new VRF private key.
-func (pad *PAD) Update(policies *Policies) {
+func (pad *PAD) Update(policies []byte) {
 	// delete older str(s) as needed
 	if len(pad.loadedEpochs) == cap(pad.loadedEpochs) {
 		n := cap(pad.loadedEpochs) / 2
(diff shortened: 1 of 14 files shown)


arlolra:
This needs a sanity check.

masomel:
I'll have time to check later today or early tomorrow.

protocol/policy.go
```diff
@@ -9,34 +9,38 @@ import (
...
 	bs = append(bs, []byte(p.LibVersion)...)                       // lib Version
 	bs = append(bs, []byte(p.HashID)...)                           // cryptographic algorithms in use
-	bs = append(bs, util.ULongToBytes(uint64(p.EpochDeadline))...) // epoch deadline
-	bs = append(bs, p.VrfPublicKey...)                             // vrf public key
 	return bs
 }
+
+// FIXME: I'm just being lazy :(
+func ParsePubKey(policies []byte) vrf.PublicKey {
+	return vrf.PublicKey(policies[0:vrf.PublicKeySize])
```
c633: This has me concerned when Policies includes more unknown-length fields.

arlolra: Yup, that's "lazy" part in the FIXME there.

If we choose to go this route, we'd probably want to use a scheme of 1 byte for len, followed by len bytes, when serializing.

arlolra: But does my,

> policies ... become an opaque blob ([]byte) in the PAD

seem ok in principle?  Or is this going to be a lot of overhead to maintain this separation?

merkletree/pad.go
```diff
@@ -19,27 +18,29 @@ var (
...
-	policies     Policies // the current policies in place
+	policies     []byte // the current policies in place
 }
 
 // NewPAD creates new PAD consisting of an array of hash chain
 // indexed by the epoch and its maximum length is len.
-func NewPAD(policies *Policies, signKey sign.PrivateKey, len uint64) (*PAD, error) {
+func NewPAD(policies []byte, signKey sign.PrivateKey, vrfKey vrf.PrivateKey, len uint64) (*PAD, error) {
 	if policies == nil {
 		panic("[merkletree] PAD must be created with a non-NULL Policies struct")
```
masomel: Probably want to update this panic message to say "non-NULL policies", or something like that.

merkletree/pad.go
```diff
@@ -63,20 +64,14 @@ func (pad *PAD) signTreeRoot(epoch uint64) {
...
 	pad.signTreeRoot(epoch)
 	pad.snapshots[epoch] = pad.latestSTR
 	pad.loadedEpochs = append(pad.loadedEpochs, epoch)
 	if policies != nil { // update the policies if necessary
-		vrfKeyChanged := 1 != subtle.ConstantTimeCompare(
-			pad.policies.vrfPrivateKey,
-			policies.vrfPrivateKey)
-		pad.policies = *policies
-		if vrfKeyChanged {
-			pad.reshuffle()
```
masomel: At which point does the tree get reshuffled when we change the VRF key if not here? I don't think this is a feature that we should drop.

protocol/policy.go
masomel: Do we have any other reason to make the policies opaque to the PAD? I am concerned about the overhead to maintain this separation; even with 

> a scheme of 1 byte for len, followed by len bytes, when serializing

I still think I'd feel more comfortable if we wouldn't have to worry about converting back and forth.

merkletree/pad.go
masomel: Just saw this: https://github.com/coniks-sys/coniks-go/issues/120#issuecomment-260853410 Feel free to ignore.

protocol/policy.go
arlolra: Well, it makes the PAD more general, perhaps to be reused in other contexts.  We could even rename PAD{Policies} to PAD{AssociatedData}.

A thought is that `AssociatedData` could be an interface with serialize/parse methods.  What our code ignores right now is that the req/resp will have to be encoded/decoded for the wire anyways, and this conversation would only happen then.  As an interface, there'd just be the casting at the usage points.

c633: > But does my, ... seem ok in principle
> We could even rename PAD{Policies} to PAD{AssociatedData}.

It seems ok to me. 
> I am concerned about the overhead to maintain this separation

I"m thinking if we could use `unsafe` here to convert between the struct and byte array (like `memcpy` in C), so we would have something like `ParsePolicies(buf []byte) *Policies`. Making `AssociatedData` be an interface sounds good, too. Maybe it can help us somehow in #6.

protocol/version.go
```diff
@@ -1,6 +1,6 @@
-package merkletree
+package protocol
 
 const (
 	// Version indicates the current protocol version.
-	Version = "3.0"
+	Version = "0.3"
```
c633: Oops, sorry for not changing this.

merkletree/pad.go
```diff
@@ -149,10 +146,18 @@ func (pad *PAD) Sign(msg ...[]byte) []byte {
 	return pad.signKey.Sign(bytes.Join(msg, nil))
 }
 
+func (pad *PAD) VrfPublicKey() vrf.PublicKey {
```
c633: I'm not sure if this method would be used anywhere other than `directory.SetPolicies`. Why couldn't we just use `directory.policies.VrfPublicKey()` in `directory.SetPolicies` instead? This seems to be a workaround when we don't have VRF key rotation yet, right?

protocol/policy.go
masomel: > it makes the PAD more general, perhaps to be reused in other contexts. 

Sure, this is a good point.

> A thought is that `AssociatedData` could be an interface with serialize/parse methods. What our code ignores right now is that the req/resp will have to be encoded/decoded for the wire anyways, and this conversation would only happen then.
>
> Maybe it can help us somehow in #6.

Yeah, this sounds like a good solution. I think it could definitely help with #6.

> I'm thinking if we could use `unsafe` here to convert between the struct and byte array (like `memcpy` in C), so we would have something like `ParsePolicies(buf []byte) *Policies`. 

You mean because `unsafe.Pointer` is like `void *` in C? I suppose we could use `unsafe` for the conversions, but I'm not sure what the advantage would be over using the scheme (1 byte for len + value).

merkletree/pad.go
arlolra: Good point.  I'll remove it.

protocol/policy.go
c633: > You mean because unsafe.Pointer is like void * in C?

Yeah.

> I suppose we could use unsafe for the conversions, but I'm not sure what the advantage would be over using the scheme (1 byte for len + value).

Hmm, I prefer something "raw" in this case than a scheme, but it probably won't work, a scheme definitely helps us control over encoding/decoding. Feel free to ignore me.

c633:
I'm not sure how do you think about this, but according to #50, and this statement
> it makes the PAD more general, perhaps to be reused in other contexts

I imagine that `merkletree` will have a `Hasher` interface that exposes a HashID. Does this mean `merkletree` should control what is the value of `HashID` being included in `AssociatedData`? 

In the other hand, if we go this route, `Hasher` and `HashID` are probably parts of `protocol`. 
What I mean is `HashID` is definitely a part and being used in `PAD` but now it's still being used in `PAD` but a part of `protocol`. (Sorry for my bad expression). It feels like to have a conflict here.

Back to #120, I think your main concern is storing many copies of `vrfPrivateKey` in memory / on disk, so should we just remove `vrfPrivateKey` out of `Policies` and keep `Policies` in `merkletree` (We could reuse a part of your code here). Does this make sense?

arlolra:
I started working on this a bit in c355ab1b.

> It feels like to have a conflict here.

I see what you're saying, but it's not clear that everyone who uses the PAD would necessarily want to include the hash id as part of the chain.

We can definitely put this on the agenda for tomorrow :)

arlolra:
Ok, implemented `AssociatedData` as an interface.  Let me know what you think.  We can discuss tomorrow as well, if preferred.

arlolra:
Merged in 29020d03.

PR #124, opened 2016-11-16
Add documentation for utils package

c633:


utils/utils.go
```diff
@@ -43,13 +43,8 @@ func UInt32ToBytes(num uint32) []byte {
...
-		return 1
-	}
-	return 0
-}
-
+// WriteFile writes buf to a file whose path is indicated by filename.
+// The file permissions are set to 0644.
 func WriteFile(filename string, buf bytes.Buffer) {
 	if _, err := os.Stat(filename); err == nil {
 		log.Printf("%s already exists\n", filename)
```
arlolra: Any reason why `[]byte(buf.String())` below isn't just `buf.Bytes()`?

c633: I also noticed this when writing doc. Going to fix it now.

utils/doc.go
```diff
@@ -0,0 +1,4 @@
+/*
+Package utils contains utility functions which are being used in coniks-go library.
```
masomel: Nitpicky: I would just say "...utility functions *for the* coniks-go..."

arlolra:
> I'm also not sure what's causing Travis to fail (doesn't seem to be a failing test or go fmt, AFAICT).

The failure is because he needs to run `go fmt ./...`.

c633:
Merged in 6fc3929883d2c4e375da6b0658f36ba46beb0717.

PR #126, opened 2016-11-18
HandleResponse() should validate response first

c633:


c633:
> this will be done twice now (which isn't bad; only a little inelegant)

Actually we did more than 2 castings in HandleResponse ;)

c633:
Oops, sorry. Wrong click :( `master` now is bcfaa441014907788a30a27bbef5915a60f2c92c.

PR #130, opened 2016-11-18
Make the usage of the registration bot configurable

c633:
Which do you prefer:
- Set `LocalAddress` to an empty string to disable the registration bot
- Or, have a `IsUseBot` field in `ServerConfig`?

### DIFF
--- a/keyserver/server.go
+++ b/keyserver/server.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net"
+	"net/url"
 	"os"
 	"os/signal"
 	"sync"
@@ -26,14 +27,14 @@ type ServerConfig struct {
 	DatabasePath        string          `toml:"database"`
 	LoadedHistoryLength uint64          `toml:"loaded_history_length"`
 	Policies            *ServerPolicies `toml:"policies"`
-	TLS                 *TLSConnection  `toml:"tls"`
+	Addresses           []*Address      `toml:"addresses"`
 }
 
-type TLSConnection struct {
-	PublicAddress string `toml:"public_address"` // address:port
-	LocalAddress  string `toml:"local_address"`  // unix socket
-	TLSCertPath   string `toml:"cert"`
-	TLSKeyPath    string `toml:"key"`
+type Address struct {
+	Address           string `toml:"address"`
+	AllowRegistration bool   `toml:"allow_registration,omitempty"`
+	TLSCertPath       string `toml:"cert,omitempty"`
+	TLSKeyPath        string `toml:"key,omitempty"`
 }
 
 type ServerPolicies struct {
@@ -57,7 +58,6 @@ type ConiksServer struct {
 	configFilePath string
 	reloadChan     chan os.Signal
 	epochTimer     *time.Timer
-	tlsConfig      *tls.Config
 }
 
 func LoadServerConfig(file string) (*ServerConfig, error) {
@@ -91,9 +91,10 @@ func LoadServerConfig(file string) (*ServerConfig, error) {
 	conf.Policies.signKey = signKey
 	// also update path for db & TLS cert files
 	conf.DatabasePath = utils.ResolvePath(conf.DatabasePath, file)
-	conf.TLS.TLSCertPath = utils.ResolvePath(conf.TLS.TLSCertPath, file)
-	conf.TLS.TLSKeyPath = utils.ResolvePath(conf.TLS.TLSKeyPath, file)
-
+	for _, addr := range conf.Addresses {
+		addr.TLSCertPath = utils.ResolvePath(addr.TLSCertPath, file)
+		addr.TLSKeyPath = utils.ResolvePath(addr.TLSKeyPath, file)
+	}
 	return &conf, nil
 }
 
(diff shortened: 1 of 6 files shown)


masomel:
> Set `LocalAddress` to an empty string to disable the registration bot

This approach seems a little safer to me.

arlolra:
Proposal: How about we allow a list of addresses to bind to in the config.toml?  Something like,

```
[[addresses]]
address = "0.0.0.0:3000"
cert = "server.pem"
key = "server.key"

[[addresses]]
address = "0.0.0.0:6000"

[[addresses]]
address = "/tmp/coniks.sock"
allow_registration = true
```

masomel:
> How about we allow a list of addresses to bind to in the config.toml?

A few questions: How does the server know which address to choose for which type of connection (bot vs. client, testing vs deployment, etc.)? It also seems like the second address in the list is an unencrypted connection (as opposed to the first one with TLS settings) -- do we want to have a non-TLS option? What does the `allow_registration = true` entry in the third address mean, is this an alternative to a `useBot` entry in the configuration/what would `allow_registration = false` mean?

arlolra:
> How does the server know which address to choose for which type of connection (bot vs. client, testing vs deployment, etc.)?

Currently, the keyserver doesn't know anything about bots; it just listens on two addresses, with differing permission.  The only change here is really to make that configurable, instead of hardcoded.

> What does the allow_registration = true entry in the third address mean

I figured the safest default would be to disallow registration on all addresses, and for the unix socket you'd explicitly enable it.

> It also seems like the second address in the list is an unencrypted connection (as opposed to the first one with TLS settings) -- do we want to have a non-TLS option?

Maybe not.

c633:
Does "disallow registration on all addresses" mean "all addresses can accept any request except registration"? Otherwise this totally makes sense to me.

arlolra:
Yup.  All the other request are of the "read" type.  Registration is the only "write".  If you want to think about it like that.  So, by default, addresses are "read-only".

masomel:
> All the other request are of the "read" type. Registration is the only "write". If you want to think about it like that.

Fair point, I agree that this is a good way to think about it.

c633:
I gave it a shot in 88103d43a14cf9d19046df90da823025dae32b84.

arlolra:
Nice work.

keyserver/server_test.go
```diff
@@ -77,14 +87,53 @@ func startServer(t *testing.T, kvdb kv.DB, epDeadline Timestamp, policiesPath st
...
+	}
+	ln, _, perms := resolveAndListen(addr)
+	defer ln.Close()
+	if perms[RegistrationType] != false {
+		t.Error("Expect disallowing registration permission.")
+	}
+
+	// test Unix network
+	addr = &Address{
+		Address:           "unix:" + testutil.LocalConnection,
```
arlolra: unix:// ?

keyserver/server_test.go
```diff
@@ -77,14 +87,53 @@ func startServer(t *testing.T, kvdb kv.DB, epDeadline Timestamp, policiesPath st
...
 	})
 }
 
+func TestResolveAddresses(t *testing.T) {
+	dir, teardown := testutil.CreateTLSCertForTest(t)
+	defer teardown()
+
+	// test TCP network
+	addr := &Address{
+		Address:     "tcp:" + testutil.PublicConnection,
```
arlolra: Mayb just change `testutil.PublicConnection` to include the protocol?

c633: I think the latest is here: https://github.com/coniks-sys/coniks-go/pull/130/files#diff-d24d2023b15c72fdfbc6fec8561dcf0dR55

> Maybe just change testutil.PublicConnection to include the protocol?

It is also being used by `testutil.NewTCPClient`. Yes, it should be a constant instead.

keyserver/server.go
```diff
@@ -119,60 +120,23 @@ func NewConiksServer(conf *ServerConfig) *ConiksServer {
...
-	unixaddr, err := net.ResolveUnixAddr(scheme, tc.LocalAddress)
-	if err != nil {
-		panic(err)
-	}
-	localLn, err := net.ListenUnix(scheme, unixaddr)
-	if err != nil {
-		panic(err)
+	for i := 0; i < len(addrs); i++ {
+		addr := addrs[i]
+		ln, tlsConfig, perms := resolveAndListen(addr)
```
c633: Yes, it makes sense. Added in 0f78e49ac4b22a8051818f30f103e7e74f2abf28. Thanks!

arlolra:
Alright, good to merge.

c633:
Yes, wait me for one sec.

masomel:
@c633 Maybe this is a silly question, how does one configure the server to use/not use the registration bot? It seems like the server default (i.e. in `cmd/init.go`) is still to use the bot. Or does this have to be configured manually after `init`?

arlolra:
Maybe a better framing is to say that the default configuration is to only allow local registrations, and if you want to deviate, you need to manually edit.  We leave the bot out of the picture.

masomel:
> Maybe a better framing is to say that the default configuration is to only allow local registrations, and if you want to deviate, you need to manually edit.

That works I guess (a flag for init may be more convenient). This isn't stated in the README or anywhere yet. Will have to add that.

c633:
> Will have to add that.

Thanks!

PR #131, opened 2016-11-22
Extra documentation for bots

masomel:
- Add package summary for `bots/coniksbot`
- Add README with usage instructions

### DIFF
--- a/bots/README.md
+++ b/bots/README.md
@@ -1,0 +1,43 @@
+# CONIKS Registration Bot for Twitter account verification in Golang
+
+## Usage
+```
+⇒  go install github.com/coniks-sys/coniks-go/bots/coniksbot
+⇒  coniksbot -h
+
+Usage:
+  coniksbot [command]
+
+Available Commands:
+  init        Create a configuration file
+  run         Run a CONIKS bot instance
+
+Flags:
+  -h, --help          help for coniksbot
+
+Use "coniksbot [command] --help" for more information about a command.
+```
+
+### Configure the bot
+
+- Generate the configuration file:
+```
+⇒  mkdir coniksbot; cd coniksbot
+⇒  coniksbot init # creates the configuration file in the current directory
+```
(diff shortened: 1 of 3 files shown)


bots/coniksbot/coniksbot.go
```diff
@@ -1,3 +1,5 @@
+// Package coniksbot provides the main executable for
```
c633: s/coniksbot/main?

masomel: I wasn't sure which one to choose. godoc.org shows it as `bots/coniksbot` so that's why I went for that. Which will you use for the keyserver? As long as we're consistent, I'm fine with either one.

c633: Maybe we can remove that summary for the `main` package...

c633: Sure, I understand this. But then the package name makes me confused, so I think it would be better to remove it, any Gopher would know what this is. Let's see what @arlolra  and @Liamsi  say.

bots/README.md
```diff
@@ -0,0 +1,43 @@
...
+### Configure the bot
+
+- Generate the configuration file:
+```
+⇒  mkdir coniksbot; cd coniksbot
+⇒  coniksbot init # creates the configuration file in the current directory
+```
+- Create a Twitter account for your bot.
+- Obtain OAuth tokens to authorize the bot for the new Twitter account:
+    - visit https://apps.twitter.com
```
c633: Nit-picking: s/visit/Visit

bots/coniksbot/coniksbot.go
masomel: > But then the package name makes me confused, so I think it would be better to remove it, any Gopher would know what this is.

Fair enough.

masomel:
> Are you going to merge this branch with fix-doc first?

I just merged it in d82cc8f. We should work on any doc fixes on that branch.

Issue #127, opened 2016-11-18
Should the `doRequestAndVerify` ignore the result's returned key for the KeyLookupType case?


Liamsi:
Code in question:
```go
var (
	key = []byte("key")
)
// [...]

func doRequestAndVerify(d *ConiksDirectory, cc *ConsistencyChecks,
	requestType int, name string) error {
	switch requestType {
	// ...
	case KeyLookupType:
		request := &KeyLookupRequest{
			Username: name,
		}
		res, _ := d.KeyLookup(request)
		return cc.HandleResponse(requestType, res, name, key)
        // ...
	}
	// ...
}

```
https://github.com/coniks-sys/coniks-go/blob/master/protocol/consistencychecks_test.go#L20-L24

Shouldn't HandleResponse take the returned key contained in the result `res` instead; although it should be the same value (`key = []byte("key")`)  it feels wrong that this test-method basically ignores the result of the lookup. A client doing a key-lookup usually wouldn't know the key in advance. Sorry for noticing so late!

c633:
`HandleResponse` takes `key` to verify the binding and it TOFU in case of key lookup (see https://github.com/coniks-sys/coniks-go/blob/master/protocol/consistencychecks.go#L218-L222). 
I'm not sure if we need to have a `ConsistencyCheck`'s method to extract the lookup result from a response. Probably worth to discuss it tonight.

Liamsi:
> HandleResponse takes key to verify the binding and it TOFU in case of key lookup  (see https://github.com/coniks-sys/coniks-go/blob/master/protocol/consistencychecks.go#L218-L222).

Not sure I understand what you are saying. You mean the consistency check in the lookup case and calling `HandleResponse` is only there to check if the key has changed?

> I'm not sure if we need to have a ConsistencyCheck's method to extract the lookup result from a response. 

The tests suggest so: https://github.com/coniks-sys/coniks-go/blob/master/protocol/consistencychecks_test.go#L20-L24 (the case "key-lookup" is handled by a consistency check method) but maybe I'm missing sth. and this is only to verify that the lookup is still the same as in the first request as your TOFU remark suggests.

> Probably worth to discuss it tonight.

Yes 👍

c633:
How is this going, @Liamsi ? 

[...]

PR #133, opened 2016-11-23
CheckBadBinding when registration

c633:
- This somehow relates to #127

### DIFF
--- a/client/coniksclient/internal/cmd/register.go
+++ b/client/coniksclient/internal/cmd/register.go
@@ -75,14 +75,14 @@ Example call:
 		case p.CheckBindingsDiffer:
 			switch response.Error {
 			case p.ReqNameExisted:
-				fmt.Println(`Are you trying to update your binding? Unfortunately, KeyChange hasn't been supported yet.`)
+				fmt.Println(`Are you trying to update your binding? Unfortunately, KeyChange isn't supported yet.`)
 			case p.ReqSuccess:
-				fmt.Println("Oops! Probably the server is trying to sneak in some other key.")
+				fmt.Println("Oops! The server snuck in some other key.")
 				recvKey, err := response.GetKey()
 				if err != nil {
 					fmt.Println("Cannot get the key from the response, error: " + err.Error())
 				}
-				fmt.Println("You got [" + string(recvKey) + "] instead of [" + string(key) + "]")
+				fmt.Println("[" + string(recvKey) + "] was registered instead of [" + string(key) + "]")
 			}
 		default:
 			fmt.Println("Error: " + err.Error())
--- a/client/coniksclient/internal/cmd/register.go
+++ b/client/coniksclient/internal/cmd/register.go
@@ -70,8 +70,19 @@ Example call:
 				// TODO: Save the cc to verify the TB and for later
 				// usage (TOFU checks)
 			case p.ReqNameExisted:
-				// TODO: See #133
 				fmt.Println("Name is already registered.")
+			}
+		case p.CheckBindingsDiffer:
+			switch response.Error {
+			case p.ReqNameExisted:
+				fmt.Println(`Are you trying to update your binding? Unfortunately, KeyChange hasn't been supported yet.`)
+			case p.ReqSuccess:
+				fmt.Println("Oops! Probably the server is trying to sneak in some other key.")
+				recvKey, err := response.GetKey()
+				if err != nil {
+					fmt.Println("Cannot get the key from the response, error: " + err.Error())
+				}
+				fmt.Println("You got [" + string(recvKey) + "] instead of [" + string(key) + "]")
 			}
 		default:
 			fmt.Println("Error: " + err.Error())
--- a/merkletree/proof.go
+++ b/merkletree/proof.go
@@ -80,21 +80,25 @@ func (ap *AuthenticationPath) authPathHash() []byte {
 	return hash
 }
 
-// VerifyBinding verifies both the value and the commitment
-// of the proof node of the authentication path.
-//
-// This should be called only when ap is a proof of inclusion.
+// VerifyBinding verifies the value of the proof node
+// of the authentication path. It also verifies the commitment
+// of the proof node if ap is a proof of inclusion.
 func (ap *AuthenticationPath) VerifyBinding(key, value []byte) bool {
-	return bytes.Equal(ap.Leaf.Value, value) &&
-		ap.Leaf.Commitment.Verify(key, value)
+	return (ap.ProofType() == ProofOfAbsence && ap.Leaf.Value == nil) ||
+		(bytes.Equal(ap.Leaf.Value, value) && ap.Leaf.Commitment.Verify(key, value))
 }
 
-// VerifyIndex checks if the private index of the proof node and
-// the lookup index match in the first l bits with l is the Level
-// of the proof node.
+// VerifyIndex verifies the private index of the proof node.
 //
-// This should be called only when ap is a proof of absence.
+// If the proof is a proof of absence, it checks if the index
+// of the proof node and the lookup index match in the first
+// l bits with l is the Level of the proof node.
+//
+// If the proof is a proof of inclusion, it returns true immediately.
 func (ap *AuthenticationPath) VerifyIndex() bool {
+	if ap.ProofType() == ProofOfInclusion {
+		return true
+	}
 	// Check if i and j match in the first l bits
 	indexBits := utils.ToBits(ap.Leaf.Index)
 	lookupIndexBits := utils.ToBits(ap.LookupIndex)
@@ -112,19 +116,6 @@ func (ap *AuthenticationPath) VerifyIndex() bool {
 // the authentication path.
 func (ap *AuthenticationPath) VerifySTR(treeHash []byte) bool {
 	return bytes.Equal(treeHash, ap.authPathHash())
-}
-
-// Verify combines VerifyBinding(), VerifyIndex(), and VerifySTR(),
-// and abstracts away the detail of the failure.
-// This should be called after the VRF index is verified successfully.
-func (ap *AuthenticationPath) Verify(key, value, treeHash []byte) bool {
-	switch {
-	case ap.ProofType() == ProofOfAbsence && ap.VerifyIndex(),
-		ap.ProofType() == ProofOfInclusion && ap.VerifyBinding(key, value):
-		return ap.VerifySTR(treeHash)
-	default:
-		return false
-	}
 }
 
 func (ap *AuthenticationPath) ProofType() ProofType {
--- a/merkletree/proof.go
+++ b/merkletree/proof.go
@@ -12,6 +12,8 @@ var (
 	// ErrBindingsDiffer indicates that the value included in the proof
 	// is different from the expected value.
 	ErrBindingsDiffer = errors.New("[merkletree] The values included in the bindings are different")
+	// ErrUnverifiableCommitment indicates that the leaf node's commitment is unverifiable.
+	ErrUnverifiableCommitment = errors.New("[merkletree] Could not verify the commitment")
 	// ErrIndicesMismatch indicates that there is a mismatch
 	// between the lookup index and the leaf index.
 	ErrIndicesMismatch = errors.New("[merkletree] The lookup index is inconsistent with the index of the proof node")
@@ -120,9 +122,11 @@ func (ap *AuthenticationPath) Verify(key, value, treeHash []byte) error {
 		}
 	} else {
 		// Verify the key-value binding returned in the ProofNode
-		if !bytes.Equal(ap.Leaf.Value, value) ||
-			!ap.Leaf.Commitment.Verify(key, value) {
+		if !bytes.Equal(ap.Leaf.Value, value) {
 			return ErrBindingsDiffer
+		}
+		if !ap.Leaf.Commitment.Verify(key, value) {
+			return ErrUnverifiableCommitment
 		}
 	}
 
--- a/merkletree/proof.go
+++ b/merkletree/proof.go
@@ -113,6 +113,8 @@ func (ap *AuthenticationPath) Verify(key, value, treeHash []byte) error {
 				return ErrIndicesMismatch
 			}
 		}
+		// expect the value is nil since we suppressed
+		// the salt & value (see Get())
 		if ap.Leaf.Value != nil {
 			return ErrBindingsDiffer
 		}
--- a/merkletree/proof.go
+++ b/merkletree/proof.go
@@ -80,38 +80,51 @@ func (ap *AuthenticationPath) authPathHash() []byte {
 	return hash
 }
 
-func (ap *AuthenticationPath) verifyBinding(key, value []byte) bool {
+// VerifyBinding verifies both the value and the commitment
+// of the proof node of the authentication path.
+//
+// This should be called only when ap is a proof of inclusion.
+func (ap *AuthenticationPath) VerifyBinding(key, value []byte) bool {
 	return bytes.Equal(ap.Leaf.Value, value) &&
 		ap.Leaf.Commitment.Verify(key, value)
 }
 
-// Verify recomputes the tree's root node from the authentication path,
-// and compares it to treeHash, which is taken from a STR.
-// Specifically, treeHash has to come from the STR whose tree returns
-// the authentication path.
-// This should be called after the VRF index is verified successfully.
-func (ap *AuthenticationPath) Verify(key, value, treeHash []byte) bool {
-	if ap.ProofType() == ProofOfAbsence { // proof of absence
-		// Check if i and j match in the first l bits
-		indexBits := utils.ToBits(ap.Leaf.Index)
-		lookupIndexBits := utils.ToBits(ap.LookupIndex)
-		for i := 0; i < int(ap.Leaf.Level); i++ {
-			if indexBits[i] != lookupIndexBits[i] {
-				return false
-			}
-		}
-	} else { // proof of inclusion
-		// Verify the key-value binding returned in the ProofNode
-		if !ap.verifyBinding(key, value) {
+// VerifyIndex checks if the private index of the proof node and
+// the lookup index match in the first l bits with l is the Level
+// of the proof node.
+//
+// This should be called only when ap is a proof of absence.
+func (ap *AuthenticationPath) VerifyIndex() bool {
+	// Check if i and j match in the first l bits
+	indexBits := utils.ToBits(ap.Leaf.Index)
+	lookupIndexBits := utils.ToBits(ap.LookupIndex)
+	for i := 0; i < int(ap.Leaf.Level); i++ {
+		if indexBits[i] != lookupIndexBits[i] {
 			return false
 		}
 	}
+	return true
+}
 
-	// step 2. Verify the auth path of the returned leaf
-	if !bytes.Equal(treeHash, ap.authPathHash()) {
+// VerifySTR recomputes the tree's root node from the authentication path,
+// and compares it to treeHash, which is taken from a STR.
+// Specifically, treeHash has to come from the STR whose tree returns
+// the authentication path.
+func (ap *AuthenticationPath) VerifySTR(treeHash []byte) bool {
+	return bytes.Equal(treeHash, ap.authPathHash())
+}
+
+// Verify combines VerifyBinding(), VerifyIndex(), and VerifySTR(),
+// and abstracts away the detail of the failure.
+// This should be called after the VRF index is verified successfully.
+func (ap *AuthenticationPath) Verify(key, value, treeHash []byte) bool {
+	switch {
+	case ap.ProofType() == ProofOfAbsence && ap.VerifyIndex(),
+		ap.ProofType() == ProofOfInclusion && ap.VerifyBinding(key, value):
+		return ap.VerifySTR(treeHash)
+	default:
 		return false
 	}
-	return true
 }
 
 func (ap *AuthenticationPath) ProofType() ProofType {
--- a/merkletree/proof.go
+++ b/merkletree/proof.go
@@ -2,9 +2,22 @@ package merkletree
 
 import (
 	"bytes"
+	"errors"
 
 	"github.com/coniks-sys/coniks-go/crypto"
 	"github.com/coniks-sys/coniks-go/utils"
+)
+
+var (
+	// ErrBindingsDiffer indicates that the value included in the proof
+	// is different from the expected value.
+	ErrBindingsDiffer = errors.New("[merkletree] The values included in the bindings are different")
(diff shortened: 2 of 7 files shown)


protocol/consistencychecks.go
```diff
@@ -245,7 +245,11 @@ func (cc *ConsistencyChecks) updateTBs(requestType int, msg *Response,
 	case RegistrationType:
 		df := msg.DirectoryResponse.(*DirectoryProof)
 		if df.AP.ProofType() == m.ProofOfAbsence {
-			if err := cc.verifyReturnedPromise(df, key); err != nil {
+			tbKey := key
+			if msg.Error == ReqNameExisted {
+				tbKey = df.TB.Value
+			}
```
arlolra: I don't agree with this.

First, why only the TB case?  What about if you try to register a different key from one that's already made its way into a signed root?  That would return a `ReqNameExisted` and a differing key, and similarly fail to verify the auth path.

And, rightly so.  It might be nice if the returned error was `ErrBindingsDiffer` or something, but this is still a case we want to fail a check on, no?

protocol/consistencychecks_test.go
```diff
@@ -3,29 +3,28 @@ package protocol
...
-		return cc.HandleResponse(requestType, res, name, key)
-	case KeyLookupType:
-		request := &KeyLookupRequest{
-			Username: name,
-		}
-		res, _ := d.KeyLookup(request)
-		return cc.HandleResponse(requestType, res, name, key)
-	case MonitoringType:
-	case KeyLookupInEpochType:
+func registerAndVerify(d *ConiksDirectory, cc *ConsistencyChecks,
```
arlolra: These functions should really be returning the error from the direction operation we're ignoring.  See all the FIXMEs already in this file.

protocol/message.go
```diff
@@ -208,3 +208,23 @@ func NewMonitoringProof(ap []*m.AuthenticationPath,
...
+//
+// If the response contains a range of authentication paths,
+// the key is obtained from the authentication path corresponding
+// with the latest signed tree roots.
+func (msg *Response) GetKey() []byte {
+	switch df := msg.DirectoryResponse.(type) {
+	case DirectoryProof:
+		return df.AP.Leaf.Value
+	case DirectoryProofs:
+		return df.AP[len(df.AP)-1].Leaf.Value
```
arlolra: Is this going to work well with `KeyLookupInEpoch` ranges?

protocol/tb.go
```diff
@@ -40,5 +40,5 @@ func (tb *TemporaryBinding) Serialize(strSig []byte) []byte {
 // the existed binding (TOFU).
 func (tb *TemporaryBinding) Verify(index, value []byte) bool {
 	return bytes.Equal(tb.Index, index) &&
-		(value != nil && bytes.Equal(tb.Value, value))
+		(value == nil || (value != nil && bytes.Equal(tb.Value, value)))
```
arlolra: When is value nil?  Also, if we do need this, it should be `(value == nil || bytes.Equal(tb.Value, value))`.

arlolra: Ok, I see what's happening here now.

In `verifyAuthPath` we do,
https://github.com/coniks-sys/coniks-go/blob/master/protocol/consistencychecks.go#L225-L229

Maybe we should do the same in `verifyReturnedPromise`?

protocol/consistencychecks_test.go
c633: Did you mean errors included in `Errors` map? Are they already being returned in `HandleResponse`?

protocol/tb.go
c633: Sure. I thought that too.

protocol/consistencychecks.go
c633: > What about if you try to register a different key from one that's already made its way into a signed root? 

Ah, right. Sorry for being hasty on this.
> It might be nice if the returned error was ErrBindingsDiffer or something, but this is still a case we want to fail a check on, no?

We are all use TOFU here, so I'm not sure if returning `ReqNameExisted` along with `CheckPassed` makes more sense?

protocol/consistencychecks_test.go
arlolra: I mean below, where you're removing `// FIXME: Check that we got an ReqNameExisted`.  I don't think that's ok.

protocol/consistencychecks.go
arlolra: > We are all use TOFU here, so I'm not sure if returning ReqNameExisted along with CheckPassed makes more sense?

Can you elaborate?  Because, I think that's what we want.  The `CheckPassed` is telling us that the `ReqNameExisted` is valid, and that the server proved it to us.  The client might still need to act on `ReqNameExisted`.

c633: I think the question is when we want to fail this check? After all the checks are passed or right away in e.g., `verifyAuthPath`.

protocol/consistencychecks_test.go
c633: Since this assertion should be checked in `directory_test.go`, I think it can be removed.

protocol/message.go
c633: Oops, I will fix the doc. 

This would be used in #93, but I should add a test for it, too.

protocol/consistencychecks.go
arlolra: Alright, why do you want this to fail later?

protocol/consistencychecks_test.go
arlolra: You're saying that since it's covered there, it's redundant here.  That's true.

As a counterpoint though, the code currently looks like,

```
// test error name existed
if registerAndVerify(d, cc, alice, key) != CheckPassed {
  t.Fatal("Unexpected verification result")
}
```

As someone reading it for correctness, it's really not obvious that `CheckPassed` means that `ReqNameExisted`.  To me, `CheckPassed` just means that the response was consistent.  It doesn't tell me what it was though.

protocol/consistencychecks.go
arlolra: I think you're right that there's definitely some nuance here, but the reason I prefer what we currently have is because, although the response is consistent in and of itself, it isn't consistent with *our expectation* that the already registered key is the one we've supplied.

protocol/consistencychecks_test.go
c633: I guess my point is the purpose of `consistencychecks_test.go` is to test the consistency verification. 
> You're saying that since it's covered there, it's redundant here

What I'm trying to say is that since we already have a test for this there, we can quite confident about the checks here. Maybe my sense was wrong :(

protocol/consistencychecks.go
c633: > although the response is consistent in and of itself, it isn't consistent with our expectation that the already registered key is the one we've supplied.

I'm not sure if we are in the same page, but this is the reason why I want to this fail later. I think most of errors could be use for whistleblowing, but `ErrBindingsDiffer` shouldn't be treated that way. I'm thinking what the client should do when receiving `ErrBindingsDiffer`.

protocol/message.go
c633: Okay, I added tests and also fixed `GetKey` in 5fcdfb71a4dfbe46c42000c6d516c6948473ab36. Thanks for asking me to write tests.

protocol/consistencychecks.go
c633: We get `ErrBindingsDiffer` (either in `verifyReturnedPromise` or `verifyAuthPath`) iff `msg.Error == ReqNameExisted`, right? In other words, does `ReqNameExisted` always inconsistent with our expectation (assume that we never do registration twice for a user)? 
If so, do we want to fail this check?

arlolra: > We get `ErrBindingsDiffer` (either in verifyReturnedPromise or verifyAuthPath) iff msg.Error == ReqNameExisted, right?

No, not "iff".  If the already registered key is the one we expect, we'd get `ReqNameExisted` but `CheckPassed`.  Also, on request, if the server does the registration but sneaks in some other key, it would return `Success` but we'd get `ErrBindingsDiffer`.

> In other words, does ReqNameExisted always inconsistent with our expectation (assume that we never do registration twice for a user)?

Are you asking if finding a `ReqNameExisted` is always, in itself, unexpected?  Or just finding a differing key?  I think there're a lot of cases where this may come up.  For example, the client disconnects when registering.  The server may have received the requests but didn't have time to send a response.  At which point, the client would reconnect and try again.  It would then receive a `ReqNameExisted` but the bindings would match.  Another more common scenario might be, a user registers normally but then loses their device or deletes their local profile or sets up a new device, or in some other way legitimately generates a new key for their account.  They attempt to register and now they'd get `ReqNameExisted` but `ErrBindingsDiffer`.

protocol/consistencychecks_test.go
arlolra: I'm not saying you're wrong; just that the tests would be more understandable to me that way.  There's nothing in `registerAndVerify(d, cc, alice, key)` returning `CheckPassed` that indicates to me that the `error name existed`.  The registration could have equally well succeeded.  We know it couldn't, because of the tests elsewhere, as you said, but that just isn't clear when trying to read this linearly.

arlolra:
@c633 Can we pull the less contention part of this that #93 needs? (The change to `verifyReturnedPromise`, I think.)

protocol/consistencychecks_test.go
c633: Fair enough. Sorry for wasting your time on this :(.

arlolra: > Sorry for wasting your time on this :(

You certainly did not waste my time.  I'm happy that you're challenging my position; please continue to do so.   That's how we improve.

c633:
I pulled out the change to `verifyReturnedPromise` and `GetKey` along with its tests in #136. What do you think?

protocol/consistencychecks.go
c633: It totally makes sense. I just have few more questions, though.
>  Another more common scenario might be, a user registers normally but then loses their device or deletes their local profile or sets up a new device, or in some other way legitimately generates a new key for their account.

Should `KeyChange` be used here instead?
>  They attempt to register and now they'd get ReqNameExisted but ErrBindingsDiffer.

In this case, how does client distinguish its own registration with other client's registration? Also, if it gets `ReqNameExisted` and `ErrBindingsDiffer` because of other user's registration, will calling `GetKey()` and re-verifying with `HandleResponse(new_key)` make sense?

arlolra: > Should KeyChange be used here instead?

Ideally, but we don't have that yet.

> In this case, how does client distinguish its own registration with other client's registration?

I'm going to assume you actually do mean "client" here (ie. the "user" is trying to reregister for some legitimate reason.)  I imagine, in both cases, the client would notify the user that a different binding already exists for that account, and we'd let the user disambiguate the situation.  If they understood that they're trying to initiate a key change, they'd do that, or otherwise report that something is amiss.

In other words, I'm not sure the client itself can distinguish which situation it's in.  I think this is getting in the realm of UI, and that's something we've hardly tackled to this point.

The case where the registration succeeds but we get an `ErrBindingsDiffer` is a little more clearcut though, but I'm not sure why any deceitful server would bother doing this.

> Also, if it gets ReqNameExisted and ErrBindingsDiffer because of other user's registration, will calling GetKey() and re-verifying with HandleResponse(new_key) make sense?

Can you clarify what you mean by "other user's registration"?

c633: > Can you clarify what you mean by "other user's registration"?

I meant the user name is already registered by someone else. I was thinking of what the client should do when getting an `ErrBindingsDiffer`.

arlolra: How can someone else register your username?  Doesn't that imply they have access to your account?  Again, I think this is a case covered above where the user themself needs to be notified.

c633: Maybe the service doesn't use a bot though. Maybe I'm just overthinking about this.

> Again, I think this is a case covered above where the user themself needs to be notified.

Yep, I agree with this.

arlolra: > Maybe the service doesn't use a bot though.

Regardless of the particular solution (bots being one), I believe the server should use an authenticated way of registering bindings.

client/coniksclient/internal/cmd/register.go
```diff
@@ -70,9 +70,20 @@ Example call:
...
 			case p.ReqNameExisted:
-				// TODO: See #133
 				fmt.Println("Name is already registered.")
 			}
+		case p.CheckBindingsDiffer:
+			switch response.Error {
+			case p.ReqNameExisted:
+				fmt.Println(`Are you trying to update your binding? Unfortunately, KeyChange hasn't been supported yet.`)
+			case p.ReqSuccess:
+				fmt.Println("Oops! Probably the server is trying to sneak in some other key.")
```
arlolra: "Oops! The server snuck in some other key."

client/coniksclient/internal/cmd/register.go
```diff
@@ -70,9 +70,20 @@ Example call:
...
+				recvKey, err := response.GetKey()
+				if err != nil {
+					fmt.Println("Cannot get the key from the response, error: " + err.Error())
+				}
+				fmt.Println("You got [" + string(recvKey) + "] instead of [" + string(key) + "]")
```
arlolra: "[ ... ] was registered instead of [ ... ]"

protocol/consistencychecks.go
```diff
@@ -224,12 +224,19 @@ func verifyAuthPath(uname string, key []byte,
...
 
-	// verify auth path
-	if !ap.Verify([]byte(uname), key, str.TreeHash) {
-		return CheckBadAuthPath
+	if ap.ProofType() == m.ProofOfAbsence {
+		if !ap.VerifyIndex() {
+			return CheckBadLookupIndex
+		}
+	} else {
+		if !ap.VerifyBinding([]byte(uname), key) {
```
arlolra: Shouldn't you do the same as below and check `key != nil &&` and remove setting key above?

protocol/consistencychecks.go
```diff
@@ -224,12 +224,19 @@ func verifyAuthPath(uname string, key []byte,
...
+			return CheckBindingsDiffer
+		}
 	}
-
-	return nil
+	if ap.VerifySTR(str.TreeHash) {
+		return nil
```
arlolra: Can you return `CheckBadAuthPath` here and then `nil` as the final case?

merkletree/proof.go
```diff
@@ -80,38 +80,51 @@ func (ap *AuthenticationPath) authPathHash() []byte {
...
 // Specifically, treeHash has to come from the STR whose tree returns
 // the authentication path.
+func (ap *AuthenticationPath) VerifySTR(treeHash []byte) bool {
+	return bytes.Equal(treeHash, ap.authPathHash())
+}
+
+// Verify combines VerifyBinding(), VerifyIndex(), and VerifySTR(),
+// and abstracts away the detail of the failure.
 // This should be called after the VRF index is verified successfully.
 func (ap *AuthenticationPath) Verify(key, value, treeHash []byte) bool {
```
arlolra: This is dead code now, no?

client/coniksclient/internal/cmd/register.go
```diff
@@ -70,9 +70,20 @@ Example call:
 				// TODO: Save the cc to verify the TB and for later
 				// usage (TOFU checks)
 			case p.ReqNameExisted:
-				// TODO: See #133
 				fmt.Println("Name is already registered.")
 			}
+		case p.CheckBindingsDiffer:
+			switch response.Error {
+			case p.ReqNameExisted:
+				fmt.Println(`Are you trying to update your binding? Unfortunately, KeyChange hasn't been supported yet.`)
```
masomel: nitpicky: s/hasn't been/isn't

merkletree/proof.go
```diff
@@ -80,38 +80,51 @@ func (ap *AuthenticationPath) authPathHash() []byte {
 	return hash
 }
 
-func (ap *AuthenticationPath) verifyBinding(key, value []byte) bool {
+// VerifyBinding verifies both the value and the commitment
+// of the proof node of the authentication path.
+//
+// This should be called only when ap is a proof of inclusion.
+func (ap *AuthenticationPath) VerifyBinding(key, value []byte) bool {
```
masomel: I feel like this should also actually check that `ap.Leaf.Index == ap.LookupIndex`. Correct me if I'm wrong, but I think we got rid of this check entirely since we introduced `a.ProofType()`?

merkletree/proof.go
masomel: Seems so to me, too.

merkletree/proof.go
arlolra: Yes, rather than the comment,

```
// This should be called only when ap is a proof of inclusion.
```

why not just assert `ap.ProofType() == ProofOfInclusion` as well?

c633: Hmm, I think I was trying to eliminate some duplicate computations here. Now, I got the idea.

merkletree/proof.go
c633: Hmm, a separate API of the merkletree package, no?

merkletree/proof.go
masomel: > why not just assert ap.ProofType() == ProofOfInclusion as well?

This particular check is already done at the `cc.verifyRegistration` level, isn't it? I'm referring to the explicit check of the leaf's index and the `ap.LookupIndex`. I don't see this check on the client-side anywhere anymore, and I don't think it's enough to just ensure that the server set `ap.ProofType() = ProofOfInclusion`.

c633: Is what we have in 54202007e7d2f32ea211b27d61678e089225bb82 right? 

> the server set ap.ProofType() = ProofOfInclusion.

I'm not sure what it means but are you saying that the `ap.ProofType` is set by the server? It is computed by the client when it receives an auth path (see https://github.com/coniks-sys/coniks-go/blob/master/merkletree/proof.go#L118).

masomel: > you saying that the ap.ProofType is set by the server? It is computed by the client when it receives an auth path.

Ah, my bad. I missed that somehow. I assumed `ap.ProofType()` was just a getter for the proof type, and that the server was setting the proof type on its end. Ignore my comment :)

masomel: Oh, `ProofType()` needs documentation, though.

merkletree/proof_test.go
```diff
@@ -7,6 +7,16 @@ import (
 	"github.com/coniks-sys/coniks-go/utils"
 )
 
+func verifyAuthPath(ap *AuthenticationPath, key, value []byte, treehash []byte) bool {
```
arlolra: I'm wondering if, rather than inlining the `ap.Verify` and `tb.Verify` and ending up w/ a duplicated `verifyAuthPath`, it would be better to change the return type of those methods to `error`?

c633: Hmm, then the `merkletree` package has to import `protocol/error.go`, right?

merkletree/proof.go
masomel: Sorry, one more question: Should the server be setting `proofType = undeterminedProof` explicitly in `merkltree.Get`? It seems like a server could set `proofType` to whatever it wanted and `ap.ProofType()` wouldn't validate that setting except for in the case of `undeterminedProof`.

c633: Yes, since `ap.proofType` is unexported, it shouldn't be transmitted over the wire. So setting `proofType` from the server-side has no impact on the client.

merkletree/proof_test.go
arlolra: No, you'd do an error type conversion in protocol.

```
if error == merkletree.BindingsDiffer{
  return CheckBindingsDiffer
}
```

merkletree/proof.go
masomel: Got it, that makes sense. Thanks!

merkletree/proof_test.go
c633: Done in 75cf9d76221f45a33893b41f094f465eab347400. What do you think?

merkletree/proof.go
```diff
@@ -80,38 +93,41 @@ func (ap *AuthenticationPath) authPathHash() []byte {
...
 		lookupIndexBits := utils.ToBits(ap.LookupIndex)
 		for i := 0; i < int(ap.Leaf.Level); i++ {
 			if indexBits[i] != lookupIndexBits[i] {
-				return false
+				return ErrIndicesMismatch
 			}
 		}
-	} else { // proof of inclusion
+		if ap.Leaf.Value != nil {
+			return ErrBindingsDiffer
```
arlolra: Is this true?  The returned leaf doesn't have to be empty.

arlolra: Ok, thank you, I forgot we did that.  Is it worth putting a comment here for people like me who don't remember so well?

c633: Sure. 
> people like me who don't remember so well.

I doubt about this :P

merkletree/proof.go
```diff
@@ -80,38 +93,43 @@ func (ap *AuthenticationPath) authPathHash() []byte {
...
+		// the salt & value (see Get())
+		if ap.Leaf.Value != nil {
+			return ErrBindingsDiffer
+		}
+	} else {
 		// Verify the key-value binding returned in the ProofNode
-		if !ap.verifyBinding(key, value) {
-			return false
+		if !bytes.Equal(ap.Leaf.Value, value) ||
+			!ap.Leaf.Commitment.Verify(key, value) {
```
arlolra: Should there be a separate error type for the commitment verification failing?

protocol/consistencychecks.go
```diff
@@ -224,12 +224,16 @@ func verifyAuthPath(uname string, key []byte,
...
-	// verify auth path
-	if !ap.Verify([]byte(uname), key, str.TreeHash) {
+	switch ap.Verify([]byte(uname), key, str.TreeHash) {
+	case m.ErrBindingsDiffer:
+		return CheckBindingsDiffer
+	case m.ErrIndicesMismatch:
+		return CheckBadLookupIndex
+	case m.ErrUnequalTreeHashes:
 		return CheckBadAuthPath
+	default:
```
arlolra: Would it be safer if the default was to panic and add a case for nil?

c633: Definitely. Thanks!

arlolra:
Ok, LGTM.  A general comment is that we probably need greater test coverage around these error cases in the merkletree, since there wasn't much updating to do there.

c633:
Yes, the tests are on their way.

c633:
I add some more tests in 4a21bea46e98695c25f685f255506033400ad6a0.

arlolra:
Thanks!  Looks good.

c633:
Merged in 2c3561a93b40ab21c45927ff710334434a85c7d0.

Issue #116, opened 2016-11-03
Client should check all cached TBs have been included in next epoch


masomel:
The client already ensures that a cached TB for its user is inserted by the next epoch during monitoring, but the client should also ensure that all other cached TBs for an epoch are included to detect a broken promise early and avoid using a potentially compromised key any further.

The idea would be that during the monitoring phase, the client also performs a lookup for the user in each TB. These lookups aren't part of the monitoring protocol per se, so a failed TB check should not cause a monitoring check to fail.

PR #134, opened 2016-11-24
Implement monitoring verification

vqhuy:
Merge #155, #170, address #173 first.

Need to consider comments in https://github.com/coniks-sys/coniks-go/issues/144 when reviewing. 

TODO:

- [x] Refactor `ConsistencyChecks` constructors.
- [ ]  Support multiple users' verification.

### DIFF
--- a/protocol/consistencychecks.go
+++ b/protocol/consistencychecks.go
@@ -27,7 +27,15 @@ import (
 type ConsistencyChecks struct {
 	// SavedSTR stores the latest verified signed tree root.
 	SavedSTR *DirSTR
+	// Bindings stores all the verified name-to-key bindings.
 	Bindings map[string][]byte
+	// RegEpoch keeps the registration epoch of each user.
+	// Entries are kept even the binding was inserted into the directory.
+	RegEpoch map[string]uint64
+
+	// oldSTR stores the previous verified STR.
+	// This is used as a cache to verify the TB.
+	oldSTR *DirSTR
 
 	// extensions settings
 	useTBs bool
@@ -39,7 +47,8 @@ type ConsistencyChecks struct {
 // NewCC creates an instance of ConsistencyChecks using
 // a CONIKS directory's pinned STR at epoch 0, or
 // the consistency state read from persistent storage.
-func NewCC(savedSTR *DirSTR, useTBs bool, signKey sign.PublicKey) *ConsistencyChecks {
+func NewCC(savedSTR *DirSTR, signKey sign.PublicKey, regs map[string]uint64,
+	useTBs bool, tbs map[string]*TemporaryBinding) *ConsistencyChecks {
 	// TODO: see #110
 	if !useTBs {
 		panic("[coniks] Currently the server is forced to use TBs")
@@ -49,9 +58,18 @@ func NewCC(savedSTR *DirSTR, useTBs bool, signKey sign.PublicKey) *ConsistencyCh
 		Bindings: make(map[string][]byte),
 		useTBs:   useTBs,
 		signKey:  signKey,
+		RegEpoch: regs,
+		oldSTR:   savedSTR,
+	}
+	if len(regs) == 0 {
+		cc.RegEpoch = make(map[string]uint64)
 	}
 	if useTBs {
-		cc.TBs = make(map[string]*TemporaryBinding)
+		if tbs == nil {
+			cc.TBs = make(map[string]*TemporaryBinding)
+		} else {
+			cc.TBs = tbs
+		}
 	}
 	return cc
 }
@@ -79,8 +97,8 @@ func (cc *ConsistencyChecks) HandleResponse(requestType int, msg *Response,
 		if _, ok := msg.DirectoryResponse.(*DirectoryProof); !ok {
 			return ErrMalformedDirectoryMessage
 		}
-	case MonitoringType, KeyLookupInEpochType:
-		if _, ok := msg.DirectoryResponse.(*DirectoryProofs); !ok {
+	case MonitoringType:
+		if _, ok := msg.DirectoryResponse.(*DirectoryProofs); !ok || msg.Error != ReqSuccess {
 			return ErrMalformedDirectoryMessage
 		}
 	default:
@@ -118,9 +136,34 @@ func (cc *ConsistencyChecks) updateSTR(requestType int, msg *Response) error {
 		}
 		// Otherwise, expect that we've entered a new epoch
 		if err := cc.verifySTRConsistency(cc.SavedSTR, str); err != nil {
+			// FIXME: this check should be removed (see #173)
 			return err
 		}
 
+	case MonitoringType:
+		// FIXME: we are returning the error immediately
+		// without saving the inconsistent STR
+		// see: https://github.com/coniks-sys/coniks-go/pull/74#commitcomment-19804686
+		strs := msg.DirectoryResponse.(*DirectoryProofs).STR
+		switch {
+		case strs[0].Epoch == cc.SavedSTR.Epoch:
+			if err := cc.verifySTR(strs[0]); err != nil {
+				return err
+			}
+		case strs[0].Epoch == cc.SavedSTR.Epoch+1:
+			if err := cc.verifySTRConsistency(cc.SavedSTR, strs[0]); err != nil {
+				return err
+			}
+		default:
+			return CheckUnexpectedMonitoringEpoch
+		}
+
+		for i := 1; i < len(strs); i++ {
+			if err := cc.verifySTRConsistency(strs[i-1], strs[i]); err != nil {
+				return err
+			}
+		}
+		str = strs[len(strs)-1]
 	default:
 		panic("[coniks] Unknown request type")
 	}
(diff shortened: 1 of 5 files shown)


c633:
Whoever has time: I think this is ready for a first review.

protocol/consistencychecks.go
```diff
@@ -215,6 +256,45 @@ func (cc *ConsistencyChecks) verifyKeyLookup(msg *Response,
...
+	dfs := msg.DirectoryResponse.(*DirectoryProofs)
+
+	str0 := dfs.STR[0]
+	ap0 := dfs.AP[0]
+	regEp, ok := cc.RegEpoch[uname]
+
+	switch {
+	case !ok && str0.Epoch == 0 && ap0.ProofType() == m.ProofOfAbsence: /* prior history verification */
+	case ok && str0.Epoch == regEp && ap0.ProofType() == m.ProofOfAbsence: /* registration epoch */
+	case ok && str0.Epoch >= regEp+1 && ap0.ProofType() == m.ProofOfInclusion: /* after registration */
```
Liamsi: Maybe define `wasUnameAbsent := ap0.ProofType() == m.ProofOfAbsence` before the `switch` and use this var in this 3 cases (with a `!` for last case)? Makes it easier to read (at least I think so; might be wrong)

protocol/consistencychecks.go
```diff
@@ -215,6 +256,45 @@ func (cc *ConsistencyChecks) verifyKeyLookup(msg *Response,
...
+	default:
+		return CheckBadAuthPath
+	}
+	if err := verifyAuthPath(uname, key, ap0, str0); err != nil {
+		return err
+	}
+
+	for i := 1; i < len(dfs.STR); i++ {
+		str := dfs.STR[i]
+		ap := dfs.AP[i]
```
Liamsi: Do we verify that `len(dfs.STR) == len(dfs.AP)` before this for-loop/always before calling `verifyMonitoring`?

protocol/consistencychecks.go
```diff
@@ -113,14 +123,42 @@ func (cc *ConsistencyChecks) updateSTR(requestType int, msg *Response) error {
...
+		case strs[0].Epoch == cc.SavedSTR.Epoch:
+			if err := verifySTR(cc.SavedSTR, strs[0]); err != nil {
+				return err
+			}
+		case strs[0].Epoch == cc.SavedSTR.Epoch+1:
+			if err := cc.verifySTRConsistency(cc.SavedSTR, strs[0]); err != nil {
+				return err
+			}
+		default:
+			panic("[coniks] Passt.")
```
Liamsi: `Passt`? Sounds very [german](https://www.dict.cc/german-english/passt.html) ;-) I'm not sure what this is intended to say.

liamsi (commented):
First pass: A few remarks/suggestions/questions.

protocol/consistencychecks.go
c633: Yes, it also saves us from some condition checks. ;)

protocol/consistencychecks.go
c633: Sure. I changed a bit in `HandleResponse` to explicitly validate each response type instead of calling `msg.validate()`.

protocol/consistencychecks.go
```diff
@@ -118,9 +118,33 @@ func (cc *ConsistencyChecks) updateSTR(requestType int, msg *Response) error {
...
+		case strs[0].Epoch == cc.SavedSTR.Epoch:
+			if err := cc.verifySTR(strs[0]); err != nil {
+				return err
+			}
+		case strs[0].Epoch == cc.SavedSTR.Epoch+1:
+			if err := cc.verifySTRConsistency(cc.SavedSTR, strs[0]); err != nil {
+				return err
+			}
+		default:
+			panic("[coniks] Unexpected monitoring epoch")
```
masomel: Do we want to panic here? This case seems more like the result of a malformed server response.

protocol/consistencychecks_test.go
```diff
@@ -225,3 +238,118 @@ func TestVerifyKeyLookupResponseWithTB(t *testing.T) {
...
+
+func TestLookupBeforeMonitoring(t *testing.T) {
+	d, pk := NewTestDirectory(t, true)
+	cc := NewCC(d.LatestSTR(), pk, make(map[string]uint64), true)
+	registerAndVerify(d, cc, alice, key)
+
+	// monitor binding was inserted
+	d.Update()
+
+	// do a lookup before the clock tells us to do monitoring
```
masomel: It's not entirely clear to me what this test is doing. It's checking whether the binding can be looked up after it's been inserted? I think there should also be a test that checks if `the binding returned to a lookup == the binding returned for monitoring`

masomel (commented):
LGTM for the most part. My main comment is that the generic auditor code should be removed since this really belongs in #170. I also have a few minor questions.

protocol/auditor.go
```diff
@@ -0,0 +1,55 @@
+// This module implements a generic CONIKS auditor, i.e. the
+// functionality that clients and auditors need to verify
+// a server's STR history.
```
c633: Yes, I intended to have #170 merged first, and this will be rebase on top of that.

protocol/consistencychecks.go
c633: It's also possibly that the client sent an invalid epoch, but I agree that we should return an error here.

protocol/consistencychecks.go
```diff
@@ -118,9 +118,33 @@ func (cc *ConsistencyChecks) updateSTR(requestType int, msg *Response) error {
 		}
 		// Otherwise, expect that we've entered a new epoch
 		if err := cc.verifySTRConsistency(cc.SavedSTR, str); err != nil {
+			// FIXME: we are returning the error immediately
+			// without saving the inconsistent STR
+			// see: https://github.com/coniks-sys/coniks-go/pull/74#commitcomment-19804686
```
c633: You mean because the registration and key lookup don't fetch the new STR as we discussed in #173 ?

protocol/consistencychecks_test.go
c633: > `the binding returned to a lookup == the binding returned for monitoring`

This check is implicit in `lookupAndVerify` and `monitorAndVerify` with `alice, key` parameters. But this should be removed because of #173.

protocol/consistencychecks.go
```diff
@@ -278,27 +289,29 @@ func (cc *ConsistencyChecks) updateTBs(requestType int, msg *Response,
...
+		// 		return err
+		// 	}
+		// 	delete(cc.TBs, uname)
 
 		case msg.Error == ReqSuccess && proofType == m.ProofOfAbsence:
 			if err := cc.verifyReturnedPromise(df, key); err != nil {
 				return err
 			}
-			cc.savePromise(uname, df.TB)
+			cc.TBs[uname] = df.TB
```
masomel: Why isn't `RegEpoch[uname]` set here, too?

masomel: Nevermind, I understand now. I was a bit confused by the FIXME comment, since it seemed to me that we were caching the issued epoch via the `RegEpoch` map. But I think it's best to address #116 in a separate PR.

protocol/consistencychecks.go
```diff
@@ -314,21 +327,15 @@ func (cc *ConsistencyChecks) updateTBs(requestType int, msg *Response,
...
-func (cc *ConsistencyChecks) savePromise(uname string, tb *TemporaryBinding) {
-	cc.TBs[uname] = tb
-	cc.RegEpoch[uname] = cc.SavedSTR.Epoch
-}
-
 // verifyFulfilledPromise verifies issued TBs were inserted
 // in the directory as promised.
 func (cc *ConsistencyChecks) verifyFulfilledPromise(uname string, str *DirSTR,
 	ap *m.AuthenticationPath) error {
-	if regEp, ok := cc.RegEpoch[uname]; ok && str.Epoch == regEp+1 {
```
masomel: I'm assuming this check will be added back in when we address #116?

protocol/auditor.go
masomel: I see. I thought we were going to merge this PR first. It's certainly closer to being done. This still duplicates code from #170, though, so in either case, I think the generic auditor code should be removed from this PR since it will be added as part of #170.

protocol/consistencychecks_test.go
masomel: Makes sense!

protocol/consistencychecks.go
masomel: Right. At least for now, we can implement this assuming that the latest STR will have been fetched as part of monitoring.

protocol/consistencychecks.go
c633: Keeping a `RegEpoch` map for other contacts seems not very nice for me, but yes, this should be fully addressed in a separate PR for #116.

protocol/auditor.go
c633: OK.

masomel (approved):

PR #136, opened 2016-11-27
Fix tb.Verify()

c633:
- Also refactor tests

- And add `response.GetKey()` to extract key from the response & add tests for it.

(Extracted from #133).

### DIFF
--- a/protocol/consistencychecks_test.go
+++ b/protocol/consistencychecks_test.go
@@ -171,7 +171,7 @@ func TestVerifyKeyLookupResponseWithTB(t *testing.T) {
 		t.Error("Expect", ReqSuccess, "got", err)
 	}
 	if err := cc.HandleResponse(KeyLookupType, res, alice, nil); err != CheckPassed {
-		t.Error("Expect", ReqSuccess, "got", err)
+		t.Error("Expect", CheckPassed, "got", err)
 	}
 	recvKey, e := res.GetKey()
 	if e != nil && !bytes.Equal(recvKey, key) {
@@ -190,7 +190,7 @@ func TestVerifyKeyLookupResponseWithTB(t *testing.T) {
 		t.Error("Expect", ReqSuccess, "got", err)
 	}
 	if err := cc.HandleResponse(KeyLookupType, res, alice, nil); err != CheckPassed {
-		t.Error("Expect", ReqSuccess, "got", err)
+		t.Error("Expect", CheckPassed, "got", err)
 	}
 	recvKey, e = res.GetKey()
 	if e != nil && !bytes.Equal(recvKey, key) {
--- a/protocol/consistencychecks_test.go
+++ b/protocol/consistencychecks_test.go
@@ -173,7 +173,8 @@ func TestVerifyKeyLookupResponseWithTB(t *testing.T) {
 	if err := cc.HandleResponse(KeyLookupType, res, alice, nil); err != CheckPassed {
 		t.Error("Expect", ReqSuccess, "got", err)
 	}
-	if !bytes.Equal(res.GetKey(), key) {
+	recvKey, e := res.GetKey()
+	if e != nil && !bytes.Equal(recvKey, key) {
 		t.Error("The directory has returned a wrong key.")
 	}
 
@@ -191,7 +192,8 @@ func TestVerifyKeyLookupResponseWithTB(t *testing.T) {
 	if err := cc.HandleResponse(KeyLookupType, res, alice, nil); err != CheckPassed {
 		t.Error("Expect", ReqSuccess, "got", err)
 	}
-	if !bytes.Equal(res.GetKey(), key) {
+	recvKey, e = res.GetKey()
+	if e != nil && !bytes.Equal(recvKey, key) {
 		t.Error("The directory has returned a wrong key.")
 	}
 
--- a/protocol/consistencychecks_test.go
+++ b/protocol/consistencychecks_test.go
@@ -1,31 +1,33 @@
 package protocol
 
-import "testing"
-
-var (
-	key = []byte("key")
+import (
+	"bytes"
+	"testing"
 )
 
-func doRequestAndVerify(d *ConiksDirectory, cc *ConsistencyChecks,
-	requestType int, name string) error {
-	switch requestType {
-	case RegistrationType:
-		request := &RegistrationRequest{
-			Username: name,
-			Key:      key,
-		}
-		res, _ := d.Register(request)
-		return cc.HandleResponse(requestType, res, name, key)
-	case KeyLookupType:
-		request := &KeyLookupRequest{
-			Username: name,
-		}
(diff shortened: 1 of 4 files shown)


protocol/consistencychecks_test.go
```diff
@@ -38,8 +40,8 @@ func TestVerifyWithError(t *testing.T) {
 
 	cc := NewCC(&str, true, pk)
 
-	if err := doRequestAndVerify(d, cc, RegistrationType, "alice"); err != CheckBadSTR {
-		t.Fatal("Expect", CheckBadSTR, "got", err)
+	if e1, e2 := registerAndVerify(d, cc, alice, key); e1 != ReqSuccess || e2 != CheckBadSTR {
+		t.Fatal("Expect", CheckBadSTR, "got", e2)
```
arlolra: Something with `e1` here as well.

protocol/consistencychecks_test.go
```diff
@@ -64,31 +66,19 @@ func TestVerifyRegistrationResponseWithTB(t *testing.T) {
...
 		t.Fatal("Expect the directory to return a signed promise")
 	}
 
 	// test error name existed
-	// FIXME: Check that we got an ReqNameExisted
-	if doRequestAndVerify(d, cc, RegistrationType, "alice") != CheckPassed {
-		t.Fatal("Unexpected verification result")
-	}
-
-	// test error name existed with different key
```
arlolra: Did you mean to drop this test?  It would return `ReqNameExisted` and some sort of `CheckBad*`, which we probably want to improve to `CheckBadBinding` or so.  We don't actually have an `ErrBindingsDiffer` yet, do we?

We can also leave a pointer to #133

protocol/consistencychecks_test.go
```diff
@@ -149,28 +143,51 @@ func TestVerifyKeyLookupResponseWithTB(t *testing.T) {
...
-	// do lookup in the same epoch
-	if doRequestAndVerify(d, cc, KeyLookupType, "alice") != CheckPassed {
-		t.Fatal("Unexpected verification result")
+	// do lookup in the same epoch - TB TOFU
+	if e1, e2 := lookupAndVerify(d, cc, alice, nil); e1 != ReqSuccess || e2 != CheckPassed {
+		t.Error(e1)
+		t.Error(e2)
+	}
+	// re-do the request and get the key from the response
+	// the key would be extracted from the TB
```
arlolra: Maybe don't use the `lookupAndVerify` helper so that you don't to do this twice.

protocol/consistencychecks_test.go
```diff
@@ -149,28 +143,51 @@ func TestVerifyKeyLookupResponseWithTB(t *testing.T) {
...
 	d.Update()
 
-	if doRequestAndVerify(d, cc, KeyLookupType, "alice") != CheckPassed {
-		t.Fatal("Unexpected verification result")
+	if e1, e2 := lookupAndVerify(d, cc, alice, key); e1 != ReqSuccess || e2 != CheckPassed {
+		t.Error(e1)
+		t.Error(e2)
+	}
+	// re-do the request and get the key from the response
+	// this time, the key would be extracted from the AP.
```
arlolra: Same

protocol/message.go
```diff
@@ -208,3 +208,30 @@ func NewMonitoringProof(ap []*m.AuthenticationPath,
...
+// a _verified_ CONIKS server's response.
+//
+// If the response contains a single authentication path,
+// the key is obtained from that authentication path or the
+// temporary binding (which depends on the returned proof type).
+//
+// If the response contains a range of authentication paths,
+// the key is obtained from the authentication path corresponding
+// with the most recent signed tree root.
+func (msg *Response) GetKey() []byte {
```
arlolra: Should we somehow assert that this is only called on responses that have been through,

```
	if Errors[msg.Error] {
		return msg.Error
	}
	if err := validateResponse(requestType, msg); err != nil {
		return err.(ErrorCode)
	}
```

protocol/tb.go
```diff
@@ -36,9 +36,8 @@ func (tb *TemporaryBinding) Serialize(strSig []byte) []byte {
 
 // Verify validates the received tb by comparing
 // index, value against the Index and Value
-// of tb. value could be nil if we have no information about
-// the existed binding (TOFU).
+// of tb.
 func (tb *TemporaryBinding) Verify(index, value []byte) bool {
 	return bytes.Equal(tb.Index, index) &&
-		(value != nil && bytes.Equal(tb.Value, value))
+		bytes.Equal(tb.Value, value)
```
arlolra: Should we keep the `value != nil` assertion?

protocol/consistencychecks_test.go
```diff
@@ -153,34 +160,37 @@ func TestVerifyKeyLookupResponseWithTB(t *testing.T) {
...
 	request := &KeyLookupRequest{
 		Username: alice,
 	}
-	res, _ := d.KeyLookup(request)
+	res, err := d.KeyLookup(request)
+	if err != ReqSuccess {
+		t.Error("Expect", ReqSuccess, "got", err)
+	}
+	if err := cc.HandleResponse(KeyLookupType, res, alice, nil); err != CheckPassed {
+		t.Error("Expect", ReqSuccess, "got", err)
```
arlolra: s/ReqSuccess/CheckPassed/

protocol/consistencychecks_test.go
```diff
@@ -153,34 +160,37 @@ func TestVerifyKeyLookupResponseWithTB(t *testing.T) {
...
 	request = &KeyLookupRequest{
 		Username: alice,
 	}
-	res, _ = d.KeyLookup(request)
+	res, err = d.KeyLookup(request)
+	if err != ReqSuccess {
+		t.Error("Expect", ReqSuccess, "got", err)
+	}
+	if err := cc.HandleResponse(KeyLookupType, res, alice, nil); err != CheckPassed {
+		t.Error("Expect", ReqSuccess, "got", err)
```
arlolra: Same

protocol/message.go
```diff
@@ -219,19 +219,28 @@ func NewMonitoringProof(ap []*m.AuthenticationPath,
 // If the response contains a range of authentication paths,
 // the key is obtained from the authentication path corresponding
 // with the most recent signed tree root.
-func (msg *Response) GetKey() []byte {
+func (msg *Response) GetKey() ([]byte, error) {
+	// repeat some assertions from consistencychecks.go
```
arlolra: Maybe we could have turned that into a helper?

c633: Yes, I was thinking it too. Please see e3c4cf418fa2dffad29c264f9bbb3dc0e020ff5c.

arlolra:
Should we merge?  And then rebase #93.

c633:
Merged in 21b146f27e740f9c435cabb49e2b63bc26890300.

Issue #101, opened 2016-10-13
cgo client library for Tor Messenger


c633:
We already agreed on the idea of `cgoDoClientBusiness` in which the cgo library would receive the response message and do all the verifications then return the verification results to the OTR addon.

Given that we would have a client module in `protocol` package, should we write the CONIKS client completely in Go and expose some high-level APIs such as
- `cgoInitClient(username, key)`
- `cgoDoRegistration(username)`
- `cgoDoLookup(username)`
- `cgoDoMonitoring(username)`
- `cgoGetLastError` - this function returns the last error of the verification, then the OTR addon could use this error message to show up to the GUI.

c633:
I think this will be landed in go 1.10: https://github.com/golang/go/issues/11058

PR #137, opened 2016-11-28
[WIP] Client-side storage

### LINKED ISSUE #127: Should the `doRequestAndVerify` ignore the result's returned key for the KeyLookupType case?

c633:
- Store bindings to file for later lookups

TODO:

- see https://github.com/coniks-sys/coniks-go/pull/137#issuecomment-264236460

### DIFF
--- a/client/coniksclient/internal/cmd/lookup.go
+++ b/client/coniksclient/internal/cmd/lookup.go
@@ -48,10 +48,12 @@ var lookupCmd = &cobra.Command{
 			return
 		}
 		response := client.UnmarshalResponse(p.KeyLookupType, res)
-		// FIXME reuse/load the cc from the registration instead
-		// FIXME same comment as in register.go
-		cc := p.NewCC(nil, true, conf.SigningPubKey)
-		err = cc.HandleResponse(p.KeyLookupType, response, name, nil)
+		cc := loadState(conf.KeyStoragePath, conf.SigningPubKey)
+		if key, ok := cc.Bindings[name]; ok {
+			err = cc.HandleResponse(p.KeyLookupType, response, name, []byte(key))
+		} else {
+			err = cc.HandleResponse(p.KeyLookupType, response, name, nil)
+		}
 		switch err {
 		case p.CheckPassed:
 			switch response.Error {
@@ -61,6 +63,9 @@ var lookupCmd = &cobra.Command{
 					fmt.Println("Cannot get the key from the response, error: " + err.Error())
 				} else {
 					fmt.Println("Success! Key bound to name is: [" + string(key) + "]")
+					if err := storeState(cc, conf.KeyStoragePath); err != nil {
+						fmt.Println("Cannot save the key to the file. Error: " + err.Error())
+					}
 				}
 			case p.ReqNameNotFound:
 				fmt.Println("Name isn't registered.")
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ server.key
 server.pem
 coniks.db/
 coniks.pid
+# CONIKS client
+keys.coniks
--- a/client/config.go
+++ b/client/config.go
@@ -11,15 +11,17 @@ import (
 
 // Config contains the client's configuration needed to send request to a
 // CONIKS-server: the path to the server's signing public-key file
-// and the actual public-key parsed from that file; the server's addresses
+// and the actual public-key parsed from that file;
+// a file path for storing the known bindings; and the server's addresses
 // for sending registration requests and other types of request, respectively.
 //
 // Note that if RegAddress is empty, it will be fallback to use Address for
 // all request types.
 type Config struct {
 	SignPubkeyPath string `toml:"sign_pubkey_path"`
+	SigningPubKey  sign.PublicKey
 
-	SigningPubKey sign.PublicKey
+	KeyStoragePath string `toml:"keys_storage_path"`
 
 	RegAddress string `toml:"registration_address,omitempty"`
 	Address    string `toml:"address"`
@@ -34,6 +36,8 @@ func LoadConfig(file string) (*Config, error) {
 	if _, err := toml.DecodeFile(file, &conf); err != nil {
 		return nil, fmt.Errorf("Failed to load config: %v", err)
 	}
+
+	conf.KeyStoragePath = utils.ResolvePath(conf.KeyStoragePath, file)
 
 	// load signing key
 	signPath := utils.ResolvePath(conf.SignPubkeyPath, file)
--- a/client/coniksclient/internal/cmd/common.go
+++ b/client/coniksclient/internal/cmd/common.go
@@ -1,10 +1,14 @@
 package cmd
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 
 	"github.com/coniks-sys/coniks-go/client"
+	"github.com/coniks-sys/coniks-go/crypto/sign"
+	"github.com/coniks-sys/coniks-go/protocol"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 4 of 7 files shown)


client/coniksclient/internal/cmd/lookup.go
```diff
@@ -61,6 +67,10 @@ var lookupCmd = &cobra.Command{
 					fmt.Println("Cannot get the key from the response, error: " + err.Error())
 				} else {
 					fmt.Println("Success! Key bound to name is: [" + string(key) + "]")
+					bindings[name] = string(key)
+					if err := writeKeys(bindings, conf.KeyStoragePath); err != nil {
+						fmt.Println("Cannot save the key to the file. Error: " + err.Error())
+					}
```
masomel: Should we also have a `CheckBindingsDiffer` case? It seems like this would be a result that the client might get if the server has inserted a fake key.

client/coniksclient/internal/cmd/lookup.go
```diff
@@ -22,6 +22,8 @@ var lookupCmd = &cobra.Command{
 			cmd.Usage()
 			return
 		}
+		bindings := loadKeys(conf.KeyStoragePath)
```
masomel: I'm wondering if `bindings` should be part of `cc`? I think it makes total sense to (re-)load the keys whenever the client does a lookup as it's implemented right now. But I'm also wondering if there would be some advantage to keeping one global `bindings` map as part of the `cc` state instead.

client/coniksclient/internal/cmd/lookup.go
c633: Hmm, I think it is handled in `default`.

masomel: Hmm, I see. I guess for the test client that's fine. For an actual client, I think the error should probably be treated separately since it reflects a consistency violation error, which is a likely indication of misbehavior, as opposed to, say, a signature verification error, which might be a misconfiguration error. Either way, we probably want to start thinking about how to report these different errors to the user.

c633: > Either way, we probably want to start thinking about how to report these different errors to the user.

I agree. Put on the agenda for the next meeting.

client/coniksclient/internal/cmd/lookup.go
c633: I gave a shot in 398dc3e124711cd4aceb2e1b3bbe528e8a18c9e5. Not sure how you feel about https://github.com/coniks-sys/coniks-go/pull/137/files#diff-8eead7b3ac1a29e7ad9e067fb3d9aaf8R59?

arlolra:
Is this really for 0.1.0?  We don't have on disk storage for the directory though.

It might be more in line with our current code to flesh out the in memory structure here, and then integrate https://godoc.org/golang.org/x/crypto/ssh/terminal in the client, to have an actual session.

Just a thought.

masomel:
> Is this really for 0.1.0? We don't have on disk storage for the directory though.

I think it should be for 0.1.0, at least the portion of this that keeps track of bindings over time (even if only in-memory), since lookups are incomplete otherwise. Plus, the persistent storage for the client shouldn't depend on/be blocked by our solution for persistent storage on the server side.

> then integrate https://godoc.org/golang.org/x/crypto/ssh/terminal in the client, to have an actual session.

Hmm, that's definitely an interesting thought. Integration with `terminal` would really only make sense for the test client, though, right?

arlolra:
> the persistent storage for the client shouldn't depend on/be blocked by our solution for persistent storage on the server side.

That's true.  I was just trying to draw the parallel.

> Integration with terminal would really only make sense for the test client, though, right?

Right, this would allow for a session, rather than two disjoint command calls.

arlolra:
> I think it should be for 0.1.0

Ok, sounds good.  Here, as in the other pull, I'm just trying to keep the momentum going :)

masomel:
> Right, this would allow for a session, rather than two disjoint command calls.

Yup, makes total sense! That's a better version of what I've done with the Java test client where we just have a loop until the user quits.

> Here, as in the other pull, I'm just trying to keep the momentum going :)

Yup :) We're so close!!

arlolra:
@c633 What do you think? And is this something you want to tackle? If not, I can hopefully find some time later in the week.

masomel:
Isn't only the documentation update left for this pull?

arlolra:
I was proposing removing the on-disk storage.

masomel:
> I was proposing removing the on-disk storage.

Ah right, sorry, that wasn't immediately clear. SGTM, though.

c633:
Oh, SGTM.  One sec. Actually it takes me sometime to figure out what you mean exactly (the session part).

arlolra:
Terminal gives you an interpreter, where memory persists until you exit.  Like when you invoke `python` from the command line.

arlolra:
I mean REPL, not interpreter, sorry.

arlolra:
Although, maybe interpreter was, since we're sorta creating a DSL.

https://en.wikipedia.org/wiki/Read%E2%80%93eval%E2%80%93print_loop
https://en.wikipedia.org/wiki/Interpreter_(computing)
https://en.wikipedia.org/wiki/Domain-specific_language

I'll stop pasting links now and be quiet.

c633:
Those links are really good! Thanks for that!
I pushed some code in `key2`, please take a look. There are still some problems I have to deal with the raw mode of the terminal.

c633:
The question just came up when I was working on this. Since we already have a running key server, the `terminal` session doesn't really help in this 
> at least the portion of this that keeps track of bindings over time (even if only in-memory), since lookups are incomplete otherwise

does it?
I also think that when we address #101, we would like to use file as a persistent storage, same as what what we are doing with Tor Messenger. What do you think?

c633:
> Terminal gives you an interpreter, where memory persists until you exit. Like when you invoke python from the command line.

I was actually looking for something like persistent files like `.coniks/coniks.keys` in `~/`. It takes me nearly a half hour to realize what you meant.

arlolra:
> Since we already have a running key server, the terminal session doesn't really help in this ... does it?

If I understand what you're asking, it does, since it satisfies, "even if only in-memory".

> I also think that when we address #101, we would like to use file as a persistent storage, same as what what we are doing with Tor Messenger. What do you think?

Yes, we'll eventually need some sort of persistent storage, but we'd want to do some sort of abstract API for that since clients in general will want to pick their storage format, much like we're planning for the keyserver.

c633:
> If I understand what you're asking, it does, since it satisfies, "even if only in-memory".

We will lost all verified bindings when the terminal is closed. I mean it is not completely 100% what we expected for keeping bindings over time, right? Is the idea we accept that minor point?

arlolra:
> Is the idea we accept that minor point?

Yes, I think so.  We have a similar issue with terminating the server process.

c633:
Yup, it makes sense. Thanks!

c633:
Close this now since it's stale.

PR #138, opened 2016-11-30
Integrate terminal to have sessions

arlolra:
In contrast with #137


c633:
Okay, I think you can take a look at this now. @arlolra @masomel @Liamsi :)

arlolra:
> I still have a question regarding to this pull

See https://github.com/coniks-sys/coniks-go/pull/137#issuecomment-264236460

> I will finish #137 tonight.

It's unclear to me what that would look like though.  I think we should hold off on that until 0.2.0

c633:
Fair enough. How about this pull?

arlolra:
> How about this pull?

This is pretty awesome; I'm looking at it now.

Noticed this,

```
>> Enter the lookup name:
>> 
[+] Error: [coniks] Malformed directory message
```

client/coniksclient/internal/cmd/run.go
```diff
@@ -0,0 +1,186 @@
...
+	p "github.com/coniks-sys/coniks-go/protocol"
+	"github.com/spf13/cobra"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+var runCmd = &cobra.Command{
+	Use:   "run",
+	Short: "Run the test client.",
+	Long: `run gives you a REPL, so that you can invoke commands to perform CONIKS operations including registration and key lookup. Currently, run supports:
+	- register:
```
arlolra: How do you feel about this?

```
register [name] [key]
```

because I keep typing,

```
lookup [name]
```

by accident.

Also, there should be some feedback for unrecognized commands.

client/coniksclient/internal/cmd/common.go
```diff
@@ -35,3 +36,9 @@ func loadConfigOrExit(cmd *cobra.Command) *client.Config {
 	}
 	return conf
 }
+
+// append "\n\r" to msg and then write to terminal in raw mode.
+func writeLineInRawMode(term *terminal.Terminal, msg string) {
+	msg += "\n\r"
```
arlolra: \r\n, no?

client/coniksclient/internal/cmd/run.go
c633: Oops, I forgot that style. Thanks!

client/coniksclient/internal/cmd/common.go
arlolra: Did this in 05d1fde

client/coniksclient/internal/cmd/run.go
arlolra: > `register [name] [key]`

Did this in 4b70fed2

arlolra:
@masomel @Liamsi Can you give this a try?  I really think it should be in 0.1.0

masomel:
> This is pretty awesome; I'm looking at it now.
> Noticed this,
>```
>>Enter the lookup name:
>>[+] Error: [coniks] Malformed directory message
```

I don't think that's the error we should be getting. The client shouldn't allow a blank name to be sent to the server in the first place.

**Comment by arlolra** [comment]
> I don't think that's the error we should be getting. The client shouldn't allow a blank name to be sent to the server in the first place.

Fixed in 89a76b728

The server should be able to handle receiving an empty name.  The problem was in the client, not handling the response properly.

**Comment by masomel** [comment]
> @masomel @Liamsi Can you give this a try? I really think it should be in 0.1.0

Will do. We *really* need to establish a cutoff between releases, though.

**Comment by masomel** [comment]
> The server should be able to handle receiving an empty name. The problem was in the client, not handling the response properly.

Oh I see. Thanks for taking care of this.

**Comment by arlolra** [comment]
> The problem was in the client, not handling the response properly.

To elaborate, that check used to be there, but we refactored it the `msg.validate()`.  Moving the cast checks ahead of that was the problem.  Maybe those belong in validate?  I'll think about that.

File: protocol/consistencychecks.go [hunk 3]
```diff
@@ -26,6 +27,7 @@ import (
 type ConsistencyChecks struct {
 	// SavedSTR stores the latest verified signed tree root.
 	SavedSTR *m.SignedTreeRoot
+	Bindings map[string][]byte
```
arlolra: @c633 points out that we need to think about where the right place for this is.

masomel:
> To elaborate, that check used to be there, but we refactored it the msg.validate(). Moving the cast checks ahead of that was the problem. Maybe those belong in validate? I'll think about that.

Got it, thanks for the clarification.

Also, I just got `[+] Error: [coniks] The hash chain is inconsistent` during testing, though the server gives me a `[coniks] Successful client request`.

arlolra:
> I just got [+] Error: [coniks] The hash chain is inconsistent during testing

Do you have the logs in your terminal?  Can you paste them?  That should be reproducible if you do, then we can debug.

arlolra:
> Got it, thanks for the clarification.

We should add a test for that (client sending an invalid message).

masomel:
Here are the client logs:

```
coniksclient run
> register alice a_key
[+] Name is already registered.
> register   key
[!] Incorrect number of args to register.
> register "" key 
[+] Succesfully registered name: ""
> lookup  
[!] Incorrect number of args to lookup.
> lookup 
[+] Error: [coniks] Malformed client message
> lookup ""
[+] Found! Key bound to name is: [key]
> lookup alice
[+] Found! Key bound to name is: [a_key]
> lookup bob
[+] Name isn't registered.
> register bob
[!] Incorrect number of args to register.
> register bob 6758w834rh
[+] Error: [coniks] The hash chain is inconsistent
> lookup bob
[+] Error: [coniks] The hash chain is inconsistent
> 
```

Should I restart the server so you can try to reproduce it?

masomel:
> We should add a test for that (client sending an invalid message).

I'm surprised we don't have such a test yet.

c633:
> Should I restart the server so you can try to reproduce it?

Hmm, I think we don't need to do it. I can debug locally.

masomel:
> I  can debug locally.

Fair point. So, after restarting the client, I no longer get this error. Here's the log:

```
> register alice
[!] Incorrect number of args to register.
> register alice a_key
[+] Name is already registered.
> lookup alice
[+] Found! Key bound to name is: [a_key]
> register   hey
[!] Incorrect number of args to register.
> register hey heyisforhorses
[+] Succesfully registered name: hey
> lookup bob
[+] Found! Key bound to name is: [6758w834rh]
>
```

masomel:
I'm out for the night, I hope my logs are somewhat helpful.

c633:
Here is my local test
```
> register alice a_key
[+] Succesfully registered name: alice
> register   key
[!] Incorrect number of args to register.
> register "" key
[+] Succesfully registered name: ""
> lookup
[!] Incorrect number of args to lookup.
> lookup
[!] Incorrect number of args to lookup.
> lookup ""
[+] Found! Key bound to name is: [key]
> lookup alice
[+] Found! Key bound to name is: [a_key]
> lookup bob
[+] Name isn't registered.
> register bob
[!] Incorrect number of args to register.
> register bob 6758w834rh
[+] Succesfully registered name: bob
> lookup bob
[+] Found! Key bound to name is: [6758w834rh]
>
```
I'm not sure if this is a bug. Btw, I made some changes according to your comments.

protocol/consistencychecks.go
c633: I think we eventually need to have this map in `ConsistencyChecks`, and it makes sense as @masomel has pointed out. For now, there is no APIs to deal with this (just being used for storing verified bindings), but when we address the persistent storage from the client side, we may need to do something with this then. Maybe just unexport this for now?

client/coniksclient/internal/cmd/run.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+var help = `
+	- register [name] [key]:
+		Register a new name-to-key binding on the CONIKS-server.
+	- lookup [name]:
+		Lookup the key of some known contact or your own bindings.
+	- exit:
+		Close the REPL and exit the client.
+`
```
arlolra: - help
   Display this message.

arlolra:
> [+] Error: [coniks] The hash chain is inconsistent

We need to figure out what went wrong.  Also, maybe this is an indication we need better debugging tooling.

Liamsi:
The help command's output seems wrongly formatted:
```
./coniksclient run
> help

	- register [name] [key]:
                                		Register a new name-to-key binding on the CONIKS-server.
                                                                                                        	- lookup [name]:
                                                                                                                                	Lookup the key of some known contact or your own bindings.
                                                         	- exit:
                                                                       		Close the REPL and exit the client.

> 
```

Liamsi:
And: This pull is very cool!

Liamsi:
Fixed the help command's output in ec7ed978b90413ed3d040ade11f006dbf7fd7ed0 (branch `key2_fix_help_command`). Feel free to merge it.

client/coniksclient/internal/cmd/run.go
Liamsi: Somehow the carriage-returns are missing when printing this message. (See https://github.com/coniks-sys/coniks-go/pull/138#issuecomment-264400700) Hence, the help message looks broken.

client/coniksclient/internal/cmd/run.go
```diff
@@ -0,0 +1,193 @@
...
+	cc := p.NewCC(nil, true, conf.SigningPubKey)
+
+	state, err := terminal.MakeRaw(int(os.Stdin.Fd()))
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer terminal.Restore(int(os.Stdin.Fd()), state)
+	term := terminal.NewTerminal(os.Stdin, "> ")
+	for {
+		term.SetPrompt("> ")
```
Liamsi: Why not set the prompt to something more specific? For example "coniks>" or "coniks-client>" (maybe that's worth debating for 0.2.0, though ;-)

c633:
> We need to figure out what went wrong. Also, maybe this is an indication we need better debugging tooling.

Yes, I agree. But it seems unreproducible :(

protocol/consistencychecks_test.go
```diff
@@ -46,6 +46,20 @@ func TestVerifyWithError(t *testing.T) {
...
+
+	request := &RegistrationRequest{
+		Username: "", // invalid username
+		Key:      key,
+	}
+	res, _ := d.Register(request)
+	if err := cc.HandleResponse(RegistrationType, res, "", key); err != ErrMalformedClientMessage {
+		t.Error("Unexpected verification result")
+	}
+}
```
Liamsi: We could have more tests like this. What is happens for instance if either the name or the key exceed a certain length (and what is expected to happen?)

c633: I think currently we have no limitations for the key size or name length? And this test is mostly for `HandleResponse` with `msg.validate()`. Other tests could be (and already) covered in `directory_test.go`.

client/coniksclient/internal/cmd/run.go
c633: Ok, it looks fun. Let me use "coniks-client>" !

Liamsi:
> Yes, I agree. But it seems unreproducible :(

Did you try to reproduce it locally? And this error happen on the (remote) test-server? This could indicate that that this only occurs if multiple people are accessing (writing to) the server at the same time (I hope not).

Liamsi:
I had a similar error just right now
<details>
<summary>The hash chain is inconsistent</summary>
<pre>
<code>
coniks-client> help

- register [name] [key]:
        Register a new name-to-key binding on the CONIKS-server.
- lookup [name]:
        Lookup the key of some known contact or your own bindings.
- help:
        Display this message.
- exit:
        Close the REPL and exit the client.

coniks-client> lookup Ismail$
[+] Name isn't registered.
coniks-client> lookup Ismail 
[+] Found! Key bound to name is: [someNotRandomLookingKey]
coniks-client> help

- register [name] [key]:
        Register a new name-to-key binding on the CONIKS-server.
- lookup [name]:
        Lookup the key of some known contact or your own bindings.
- help:
        Display this message.
- exit:
        Close the REPL and exit the client.

coniks-client> register someone
[!] Incorrect number of args to register.
coniks-client> register someone with_a_key
[+] Error: [coniks] The hash chain is inconsistent
coniks-client> 

</code>
</pre>
</details>

c633:
I couldn't reproduce it because of `epoch_deadline`. What is your `epoch_deadline` value? Investigating it...

Liamsi:
I just used the test-server in Princeton. IIRC the epoch deadline is 1 min there.

c633:
I think the bug is because we didn't have monitoring yet. If you do 2 operations in `t` and `t+2`, then the bug happens.

Liamsi:
The problem happens if the client misses an epoch in between two commands (and in one and the same terminal session)? Can we open an issue for that?

c633:
> Can we open an issue for that?

I reopened #90 for this (with a TODO).

Liamsi:
Is this for 0.1.0, though?

c633:
Yes!

Liamsi:
OK, I can give it a try this afternoon!

client/coniksclient/internal/cmd/run.go
```diff
@@ -0,0 +1,198 @@
...
+	"strings"
+
+	"github.com/coniks-sys/coniks-go/client"
+	"github.com/coniks-sys/coniks-go/keyserver/testutil"
+	p "github.com/coniks-sys/coniks-go/protocol"
+	"github.com/spf13/cobra"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+const help = "\n- register [name] [key]:\r\n" +
```
arlolra: Why does it only start off w/ `\n` whereas the rest are `\r\n`?

protocol/consistencychecks.go
arlolra: > when we address the persistent storage from the client side, we may need to do something with this then

Can we leave a TODO to think about the appropriate place here then?

client/coniksclient/internal/cmd/run.go
Liamsi: > Why does it only start off w/ \n whereas the rest are \r\n?

Don't know but it seems like the carriage-return wasn't neccessary here (I tested it, it worked). I though an extra newline looked nicer. But it's also fine now. Thanks @c633

c633: > I though an extra newline looked nicer

Then we should add newline for all commands. Just a minor point. If you think it's good, we should do it.

arlolra:
I think this is a significant improvement and we should land it as is.  The interactivity is already uncovering some bugs, which is great.

Liamsi:
I agree.

c633:
Merged in 9b17e1594a9c0d82b09a99dbc69415308528a03e. I also closed #127.

iesiyok:
This problem happening 

> Here are the client logs:
> 
> ```
> coniksclient run
> > register alice a_key
> [+] Name is already registered.
> > register   key
> [!] Incorrect number of args to register.
> > register "" key 
> [+] Succesfully registered name: ""
> > lookup  
> [!] Incorrect number of args to lookup.
> > lookup 
> [+] Error: [coniks] Malformed client message
> > lookup ""
> [+] Found! Key bound to name is: [key]
> > lookup alice
> [+] Found! Key bound to name is: [a_key]
> > lookup bob
> [+] Name isn't registered.
> > register bob
> [!] Incorrect number of args to register.
> > register bob 6758w834rh
> [+] Error: [coniks] The hash chain is inconsistent
> > lookup bob
> [+] Error: [coniks] The hash chain is inconsistent
> > 
> ```
> Should I restart the server so you can try to reproduce it?

I have this problem on my computer. I am running as the following:

> ./coniksclient run
>>register alice key
Error: [coniks] The hash chain is inconsistent. Maybe the client missed an epoch in between two commands, monitoring isn't supported yet.

I don't understand what this error message means. There are no errors and no warnings on the server side. But when I run the same command again :
>>register alice key
The same error message occurs on the client, but there is a warning on the server side this time "[coniks] Registering identity is already registered".

I believe that means the key has already been created with the first time, because when I run lookup:
>>lookup alice
The same error message occurs on the client, no warnings on the server side. 
>>lookup bob
But searching another name results in a warning on the server side "[coniks] Searched name not found in directory"

I couldn't find a solution to this, could you please help me?
Is there a method, a file or a folder that I can see all recorded name-key bindings?

masomel:
Hi @iesiyok ! Thank you for reporting this problem to us. Please open a new issue for this problem, so we can properly track this.

PR #140, opened 2016-12-01
Refine landing page/main Readme:

### LINKED ISSUE #139: Repository's Readme should have pointer to installation instructions
There should be a sub-section which simply points to the clients/servers installation instructions (in their Readmes).

Liamsi:
- Add pointers to the client and server Readmes
- some minor refinements: remove double apostrophes, consistent Headers
resolves #139

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,10 @@ This allows messaging clients to verify the identity of
 users automatically, and prevents malicious/compromised servers from
 hijacking secure communications without getting caught.
 
+This repository provides a [Golang](https://golang.org) implementation of the 
+CONIKS system. The implementation consists of a library, described in the 
+following section, and a standalone CONIKS-server and -client using this library.
+
 ## Golang Library
 
 The packages in this library implement the various components of the CONIKS
@@ -39,8 +43,8 @@ We recommend using a [tagged release](https://github.com/coniks-sys/coniks-go/re
 
 ## Installation
 
-The preferred way to install CONIKS is by downloading the
-[latest release](releases/latest).
+The preferred way to install CONIKS reference server or client implementation
+is by downloading the [latest release](releases/latest).
 <!-- and by following the contained instructions. -->
 
 If you prefer to build and install from source you need to have a recent Golang
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 
 http://coniks.org
 
-## Introduction
+## Overview
 
 CONIKS is a key management system that provides transparency and privacy
 for end-user public keys.
@@ -19,7 +19,8 @@ hijacking secure communications without getting caught.
 
 This repository provides a [Golang](https://golang.org) implementation of the 
 CONIKS system. The implementation consists of a library, described in the 
-following section, and a standalone CONIKS-server and -client using this library.
+following section, and a standalone CONIKS-server and -client, and a registration 
+proxy using this library.
 
 ## Golang Library
 
@@ -35,27 +36,29 @@ system and may be imported individually.
 - `protocol`: CONIKS protocols implementation/library
 - `storage`: Hooks for persistent storage backend (currently unused)
 
-## Disclaimer
-
-Please keep in mind that this CONIKS library is under active development.
-The repository may contain experimental features that aren't fully tested.
-We recommend using a [tagged release](https://github.com/coniks-sys/coniks-go/releases).
-
 ## Installation
 
 The preferred way to install CONIKS reference server or client implementation
 is by downloading the [latest release](releases/latest).
 <!-- and by following the contained instructions. -->
 
-If you prefer to build and install from source you need to have a recent 
-[Golang](https://golang.org/doc/install) version installed (1.6+). 
-If Golang is setup correctly, you can simply run `go get github.com/coniks-sys/coniks-go` in your command line.
+If you prefer to build and install from source you need to have 
+[Golang](https://golang.org/doc/install) version 1.6 or higher installed.
+If Golang is set up correctly, you can simply run 
+`go get github.com/coniks-sys/coniks-go` in your command line.
 
-In both cases, installing CONIKS installs a [CONIKS-server](keyserver/) and a
-simple command-line [client](client). Both commands are documented in their
-own package ([server here](keyserver) and [client here](client)).
+In both cases, installing coniks-go installs a [CONIKS-server](keyserver), a
+simple command-line [client](client), and the [registration-proxy](bots).
+For usage instructions, see the documentation in their respective packages
+([server here](keyserver) and [client here](client)).
 
-## API-Documentation
+## Disclaimer
+
+Please keep in mind that this CONIKS library is under active development.
+The repository may contain experimental features that aren't fully tested.
+We recommend using a [tagged release](https://github.com/coniks-sys/coniks-go/releases).
+
+## API Documentation
 
 https://godoc.org/github.com/coniks-sys/coniks-go
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# CONIKS Go Library
+# CONIKS Go
 
 [![Build Status](https://travis-ci.org/coniks-sys/coniks-go.svg?branch=master)](https://travis-ci.org/coniks-sys/coniks-go)
 [![Coverage Status](https://coveralls.io/repos/github/coniks-sys/coniks-go/badge.svg?branch=master&dummy=1)](https://coveralls.io/github/coniks-sys/coniks-go)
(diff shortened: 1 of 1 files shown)


Liamsi:
I don't really like that the added section is larger than the others. If you agree we could skip the sentence about Golang? 
Also, it might be confusing that there isn't a sentence what exactly this repo is for: the main header says "CONIKS Go Library" although it is both: a library and reference implementation of the server/client using this library. Maybe we should add a sentence or two stating exactly that in the introduction? 
Maybe sth. like: "This repository provides a Golang implementation of the CONIKS system. The implementation consists of a library described below and a standalone CONIKS-server and -client using this library."

masomel:
> I don't really like that the added section is larger than the others. If you agree we could skip the sentence about Golang? (although it might be helpful if someone not familiar with golang wants to experiment with it)

Looking at it, the "Installation" section doesn't seem to be larger than the others, e.g. the "Golang Library" section actually looks longer to me. I also think that the short paragraph on Golang is helpful, the only sentence I might remove is the one that points to the Go installation page. Indicating the supported Go version and the `get` command seems ok to me.

> Also, it might be confusing that there isn't a sentence what exactly this repo is for: the main header says "CONIKS Go Library" although it is both: a library and reference implementation of the server/client using this library. Maybe we should add a sentence or two stating exactly that in the introduction?
Maybe sth. like: "This repository provides a Golang implementation of the CONIKS system. The implementation consists of a library described below and a standalone CONIKS-server and -client using this library."

Good point. I might even change the title of the README to "CONIKS Go", too. I think the sentence you added to the intro makes sense.

README.md
```diff
@@ -5,7 +5,8 @@
 
 http://coniks.org
 
-##Introduction
+## Introduction
```
masomel: Looking at some other repos' READMEs, they use "Overview" instead of "Introduction", which I think I like a bit better for this section. What do you think?

README.md
```diff
@@ -16,28 +17,51 @@ This allows messaging clients to verify the identity of
...
+- `storage`: Hooks for persistent storage backend (currently unused)
 
 ## Disclaimer
+
 Please keep in mind that this CONIKS library is under active development.
 The repository may contain experimental features that aren't fully tested.
 We recommend using a [tagged release](https://github.com/coniks-sys/coniks-go/releases).
 
-##Documentation
+## Installation
```
masomel: I would move this section above the "Disclaimer" section (which I think should be the last section of the README).

README.md
```diff
@@ -16,28 +17,51 @@ This allows messaging clients to verify the identity of
...
+
+The preferred way to install CONIKS reference server or client implementation
+is by downloading the [latest release](releases/latest).
+<!-- and by following the contained instructions. -->
+
+If you prefer to build and install from source you need to have a recent Golang
+version installed (1.6+). See Golang's [documentation](https://golang.org/doc/install)
```
masomel: I think you can drop the "See Golang's documentation..." sentence. You already link to Golang above.

README.md
```diff
@@ -16,28 +17,51 @@ This allows messaging clients to verify the identity of
...
 We recommend using a [tagged release](https://github.com/coniks-sys/coniks-go/releases).
 
-##Documentation
+## Installation
+
+The preferred way to install CONIKS reference server or client implementation
+is by downloading the [latest release](releases/latest).
+<!-- and by following the contained instructions. -->
+
+If you prefer to build and install from source you need to have a recent Golang
```
masomel: I might change this to "...you need to have Golang version 1.6 or higher installed".

README.md
```diff
@@ -16,28 +17,51 @@ This allows messaging clients to verify the identity of
...
+
 Please keep in mind that this CONIKS library is under active development.
 The repository may contain experimental features that aren't fully tested.
 We recommend using a [tagged release](https://github.com/coniks-sys/coniks-go/releases).
 
-##Documentation
+## Installation
+
+The preferred way to install CONIKS reference server or client implementation
+is by downloading the [latest release](releases/latest).
```
masomel: Sorry if this is a silly question, but won't downloading the latest release from github just give them the source?

README.md
```diff
@@ -16,28 +17,50 @@ This allows messaging clients to verify the identity of
...
+<!-- and by following the contained instructions. -->
+
+If you prefer to build and install from source you need to have a recent 
+[Golang](https://golang.org/doc/install) version installed (1.6+).
```
masomel: Nitpicky: I might change this to "... you need to have Golang version 1.6 or higher installed". I think that makes it a little bit more explicit that we require Go 1.6+.

README.md
```diff
@@ -16,28 +17,50 @@ This allows messaging clients to verify the identity of
...
-##Documentation
+## Installation
+
+The preferred way to install CONIKS reference server or client implementation
+is by downloading the [latest release](releases/latest).
+<!-- and by following the contained instructions. -->
+
+If you prefer to build and install from source you need to have a recent 
+[Golang](https://golang.org/doc/install) version installed (1.6+). 
+If Golang is setup correctly, you can simply run `go get github.com/coniks-sys/coniks-go` in your command line.
```
masomel: s/setup/set up

README.md
```diff
@@ -16,28 +17,50 @@ This allows messaging clients to verify the identity of
...
+
+In both cases, installing CONIKS installs a [CONIKS-server](keyserver/) and a
```
masomel: Nitpicky: s/CONIKS/coniks-go

README.md
```diff
@@ -16,28 +17,50 @@ This allows messaging clients to verify the identity of
 users automatically, and prevents malicious/compromised servers from
 hijacking secure communications without getting caught.
 
+This repository provides a [Golang](https://golang.org) implementation of the 
+CONIKS system. The implementation consists of a library, described in the 
+following section, and a standalone CONIKS-server and -client using this library.
```
masomel: I think we should mention the registration proxy here, too.

README.md
```diff
@@ -16,28 +17,50 @@ This allows messaging clients to verify the identity of
...
+is by downloading the [latest release](releases/latest).
+<!-- and by following the contained instructions. -->
+
+If you prefer to build and install from source you need to have a recent 
+[Golang](https://golang.org/doc/install) version installed (1.6+). 
+If Golang is setup correctly, you can simply run `go get github.com/coniks-sys/coniks-go` in your command line.
+
+In both cases, installing CONIKS installs a [CONIKS-server](keyserver/) and a
+simple command-line [client](client). Both commands are documented in their
+own package ([server here](keyserver) and [client here](client)).
```
masomel: I would change the last sentence to something like "For usage instructions, see the documentation in their respective packages (server here and client here)". Should we mention the bots here, too?

README.md
```diff
@@ -16,28 +17,50 @@ This allows messaging clients to verify the identity of
...
+
+## API-Documentation
```
masomel: Super nitpicky: s/API-Documentation/API Documentation.

README.md
Liamsi: I think you can also possible to add binary releases (I actually don't know if we want to go this way yet).

README.md
Liamsi: I agree.

README.md
masomel: I see yeah, it would be cool if we could just include the binary, but you're right, maybe we shouldn't do that yet. So, maybe we should clarify in the README that the way to install it is from source right now? Might make it a bit easier even, since then the flow is just `go get; go install` for the reference implementations.

README.md
```diff
@@ -16,28 +17,53 @@ This allows messaging clients to verify the identity of
 users automatically, and prevents malicious/compromised servers from
 hijacking secure communications without getting caught.
 
+This repository provides a [Golang](https://golang.org) implementation of the 
+CONIKS system. The implementation consists of a library, described in the 
+following section, and a standalone CONIKS-server and -client, and a registration
```
arlolra: s/and a standalone/a standalone/

README.md
```diff
@@ -16,28 +17,53 @@ This allows messaging clients to verify the identity of
...
+- `merkletree`: Merkle prefix tree and related data structures
+- `utils`: Utility functions
+- `protocol`: CONIKS protocols implementation/library
+- `storage`: Hooks for persistent storage backend (currently unused)
+
+## Installation
+
+The preferred way to install CONIKS reference server or client implementation
+is by downloading the [latest release](releases/latest).
+<!-- and by following the contained instructions. -->
```
arlolra: We're going to be distributing platform / arch specific compiled binaries?  Interesting ...

arlolra: Whoops, sorry for the repetition.

Then can we get rid of that blurb and only describe installing from source for the time being?

c633:
LGTM.

c633:
Btw, do you want to resolve #141 here too?

arlolra:
> Btw, do you want to resolve #141 here too?

Let's keep these pulls small and separate :)

README.md
masomel: > Whoops, sorry for the repetition.

Nbd, I've done that, many times, too.

> Then can we get rid of that blurb and only describe installing from source for the time being?

That's what I would say, otherwise, it's misleading.

Liamsi:
> I made some changes in e5c469c. Feel free to merge if you approve it.

These changes look good (and they address all of @arlolra's comments). Thanks @c633!

Squashing and merging it now.

PR #132, opened 2016-11-22
Fix godoc display issues, other doc fixes

c633:

PR #143, opened 2016-12-02
Add doc for server's confs

### LINKED ISSUE #141: Document policies and other confs
On https://github.com/coniks-sys/coniks-go/blob/master/keyserver/README.md

Running `coniksserver init` spits a config.toml with,

```
database = "coniks.db"
loaded_history_length = 1000000

[policies]
  epoch_deadline = 60
  vrf_key_path = "vrf.priv"
  sign_key_path = "sign.priv"

[[addresses]]
  address = "unix:///tmp/coniks.sock"
  allow_registration = true

...
```

but I don't know what units the `epoch_deadline` is?  Seconds?

Is the `database` useful yet?

We might need something similar for the client as well.

c633:
- Resolve #141

### DIFF
--- a/keyserver/README.md
+++ b/keyserver/README.md
@@ -36,7 +36,8 @@ Use "coniksserver [command] --help" for more information about a command.
 is for the registration proxy, the second is the server's public address
 for "read-only" requests (lookups, monitoring etc).
 - Edit the configuration file as needed:
-    - Replace the `epoch_deadline` with the desired duration in seconds.
+    - Replace the `loaded_history_length` with the desired number of snapshots kept in memory.
+    - Replace the `epoch_deadline` with the desired duration in **seconds**.
     - If using a CONIKS registration proxy, replace the registration proxy `address`. Otherwise, remove the registration proxy `addresses` entry, and add `allow_registration = true` field to the public `addresses` entry.
     - In either case, replace the public `address` with the server's public CONIKS address.
 - Test setup (no registration proxy) config file example:
--- a/keyserver/coniksserver/internal/cmd/init.go
+++ b/keyserver/coniksserver/internal/cmd/init.go
@@ -53,7 +53,6 @@ func mkConfig(dir string) {
 		},
 	}
 	var conf = keyserver.ServerConfig{
-		DatabasePath:        "coniks.db",
 		LoadedHistoryLength: 1000000,
 		Addresses:           addrs,
 		Policies: &keyserver.ServerPolicies{
--- a/keyserver/server.go
+++ b/keyserver/server.go
@@ -65,7 +65,7 @@ type Address struct {
 
 // ServerPolicies contains a server's CONIKS policies configuration
 // including paths to the VRF private key, the signing private
-// key and the epoch deadline value.
+// key and the epoch deadline value in seconds.
 type ServerPolicies struct {
 	EpochDeadline protocol.Timestamp `toml:"epoch_deadline"`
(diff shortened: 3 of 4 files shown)


keyserver/README.md
```diff
@@ -36,7 +36,9 @@ Use "coniksserver [command] --help" for more information about a command.
 is for the registration proxy, the second is the server's public address
 for "read-only" requests (lookups, monitoring etc).
 - Edit the configuration file as needed:
-    - Replace the `epoch_deadline` with the desired duration in seconds.
+    - Replace the `database` with the directory location of the persistent storage. However, this value is currently unused.
```
arlolra: It's not actually unused though, because the database gets created.

Did we have ticket about temporarily removing the storage stuff before the release?  Is that something we still want to do?

Liamsi: One question: Isn't it better to skip this value if it isn't used currently?

Liamsi:
LGTM

arlolra:
I think it would be helpful to put some comments in the config.toml as well.

keyserver/README.md
c633: I submitted #145 for this. Maybe we can merge it right away.

masomel:
LGTM, too. I agree with @arlolra, comments in the config.toml would be helpful for someone who isn't familiar with toml files.

c633:
I just remembered that I have tried to encode comments to generated TOML files as well, but I haven't found any way, unfortunately.

keyserver/README.md
c633: Should we do something like #145 temporarily?

Liamsi:
Not sure if this helps: https://github.com/BurntSushi/toml-test/blob/master/tests/valid/comments-everywhere.toml

c633:
Hmm, I think it is edited manually, not generated by `toml.Encode`.

keyserver/README.md
masomel: > It just seems to be odd to me when we have WithDB in each test (which I don't really want to remove it for now :( ).

Makes sense. To a developer reading our code though, it might just be confusing to have some "random" db-related code.

masomel:
> Hmm, I think it is edited manually, not generated by toml.Encode.

I see. I'm fine not adding this (commented toml files) in for 0.1.0.

arlolra:
The library doesn't have support for it yet in the encoder.  See the open issues / pulls there.

> I'm fine not adding this (commented toml files) in for 0.1.0

Yup, let's skip it for now.

c633:
Then should we merge this now (along with #145)?

keyserver/README.md
```diff
@@ -36,7 +36,8 @@ Use "coniksserver [command] --help" for more information about a command.
 is for the registration proxy, the second is the server's public address
 for "read-only" requests (lookups, monitoring etc).
 - Edit the configuration file as needed:
-    - Replace the `epoch_deadline` with the desired duration in seconds.
```
arlolra: Wait, this was there to begin with, before #141?!

I'm not smart :(

c633: Yes, @masomel did write it in #132.

arlolra:
> Then should we merge this now (along with #145)?

LGTM, but I'm losing faith in myself.

keyserver/README.md
c633: Oops, sorry :p That definitely isn't what I meant.

arlolra: Ha, don't worry, I'm just creating noise.

PR #146, opened 2016-12-02
Logging

arlolra:
- [x] Refactor logger calls (by using our wrapper)

### DIFF
--- a/keyserver/listener.go
+++ b/keyserver/listener.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"crypto/tls"
 	"io"
-	"log"
 	"net"
 	"time"
 
@@ -35,7 +34,7 @@ func (server *ConiksServer) handleRequests(ln net.Listener, tlsConfig *tls.Confi
 			if opErr, ok := err.(*net.OpError); ok && opErr.Timeout() {
 				continue
 			}
-			log.Printf("accept client: %s", err)
+			server.logger.Error(err.Error())
 			continue
 		}
 		if _, ok := ln.(*net.TCPListener); ok {
@@ -55,7 +54,8 @@ func (server *ConiksServer) acceptClient(conn net.Conn, handler func(msg []byte)
 
 	var buf bytes.Buffer
 	if _, err := io.CopyN(&buf, conn, 8192); err != nil && err != io.EOF {
-		log.Printf("client read %v: %v", conn.RemoteAddr(), err)
+		server.logger.Error(err.Error(),
+			"address", conn.RemoteAddr().String())
 		return
 	}
 
@@ -63,13 +63,15 @@ func (server *ConiksServer) acceptClient(conn net.Conn, handler func(msg []byte)
 	// TODO: The `err` returned here is purely for logging purposes.  It
 	// would be better for `handler` not to return any error, and instead
 	// log if the error code in the `res` is not `ReqSuccess`.
-	if err != nil {
-		log.Printf("client handle %v: %v", conn.RemoteAddr(), err)
+	if err != ReqSuccess {
+		server.logger.Warn(err.Error(),
+			"address", conn.RemoteAddr().String())
 	}
 
 	_, err = conn.Write([]byte(res))
 	if err != nil {
-		log.Printf("client write %v: %v", conn.RemoteAddr(), err)
+		server.logger.Error(err.Error(),
+			"address", conn.RemoteAddr().String())
 		return
 	}
 }
@@ -120,7 +122,8 @@ func (server *ConiksServer) makeHandler(acceptableTypes map[int]bool) func(msg [
 			return malformedClientMsg(err)
 		}
 		if !acceptableTypes[req.Type] {
-			log.Printf("unacceptable message type: %q", req.Type)
+			server.logger.Error("Unacceptable message type",
+				"request type", req.Type)
 			return malformedClientMsg(ErrMalformedClientMessage)
 		}
 
--- a/keyserver/server.go
+++ b/keyserver/server.go
@@ -4,7 +4,6 @@ import (
 	"crypto/tls"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"net"
 	"net/url"
 	"os"
@@ -30,7 +29,8 @@ type ServerConfig struct {
 	// Policies contains the server's CONIKS policies configuration.
 	Policies *ServerPolicies `toml:"policies"`
 	// Addresses contains the server's connections configuration.
-	Addresses      []*Address `toml:"addresses"`
+	Addresses      []*Address          `toml:"addresses"`
+	Logger         *utils.LoggerConfig `toml:"logger"`
 	configFilePath string
 }
 
@@ -81,6 +81,8 @@ type ServerPolicies struct {
 // a mechanism to update the underlying ConiksDirectory automatically
 // at regular time intervals.
 type ConiksServer struct {
+	logger *utils.Logger
+
 	sync.RWMutex
 	dir *protocol.ConiksDirectory
 
@@ -131,6 +133,9 @@ func LoadServerConfig(file string) (*ServerConfig, error) {
 		addr.TLSCertPath = utils.ResolvePath(addr.TLSCertPath, file)
 		addr.TLSKeyPath = utils.ResolvePath(addr.TLSKeyPath, file)
 	}
+	// logger config
+	conf.Logger.Path = utils.ResolvePath(conf.Logger.Path, file)
+
 	return &conf, nil
 }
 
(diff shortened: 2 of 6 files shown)


arlolra:
This was a stray branch related to #45

keyserver/listener.go
```diff
@@ -55,7 +55,8 @@ func (server *ConiksServer) acceptClient(conn net.Conn, handler func(msg []byte)
 
 	var buf bytes.Buffer
 	if _, err := io.CopyN(&buf, conn, 8192); err != nil && err != io.EOF {
-		log.Printf("client read %v: %v", conn.RemoteAddr(), err)
+		server.logger.Error(err.Error(),
+			zap.String("address", conn.RemoteAddr().String()))
```
masomel: Is there a way we can completely wrap `zap` in our logger wrapper? It would be nice to just be able to write something like `server.logger.Error(err.Error(), "address", conn.RemoteAddr().String())`

keyserver/server.go
```diff
@@ -81,6 +82,8 @@ type ServerPolicies struct {
 // a mechanism to update the underlying ConiksDirectory automatically
 // at regular time intervals.
 type ConiksServer struct {
+	logger zap.Logger
```
masomel: Related to https://github.com/coniks-sys/coniks-go/pull/146/files#r101622894: I think it'd be nice if we could say `utils.Logger` instead of `zap.Logger`.

utils/logger.go
```diff
@@ -0,0 +1,37 @@
...
+			zap.RFC3339Formatter("timestamp"), // human-readable timestamps
+			zap.MessageKey("message"),         // customize the message key
+			zap.LevelString("level"),          // stringify the log level
+		),
+		zap.Output(f),
+		*zlevel,
+	)
+
+	return logger
+}
```
masomel: Also related to my other comments, I think it'd be nice to add wrappers for `zap.Logger.Info()`, `Warn()` and `Error()`.

keyserver/server.go
c633: Sure. I was waiting for zap's `SugaredLogger`. Hope to finish this PR soon ;)

masomel: Ok cool! I figured you had something in mind, but I wanted to start moving things along again.

c633:
PTAL. /cc @arlolra @masomel @Liamsi

keyserver/listener.go
```diff
@@ -55,21 +54,24 @@ func (server *ConiksServer) acceptClient(conn net.Conn, handler func(msg []byte)
...
-		log.Printf("client handle %v: %v", conn.RemoteAddr(), err)
+	if err != ReqSuccess {
+		server.logger.Warnw(err.Error(),
+			"address", conn.RemoteAddr().String())
 	}
 
 	_, err = conn.Write([]byte(res))
 	if err != nil {
-		log.Printf("client write %v: %v", conn.RemoteAddr(), err)
+		server.logger.Errorw(err.Error(),
```
Liamsi: Why is this an error and the above `err != ReqSuccess` a warning?

c633: I think we do not expect an error here (hence `Error`), whereas the above `err != ReqSuccess` is the return status and if there is something like `ErrDirectory` etc I think we would like to treat it as a warning.

c633:
> The calls to server.logger.Warnw server.logger.Error etc. won't work anymore if we, for whatever reasons, decide to switch the logging framework, right?

As long as the new logging framework support structured context as key-value pairs like `zap` then, it still works. Otherwise, we can wrap its APIs by JSON encoding our key-value pairs then pass this string to the logger (not ideally but I think it works).

Liamsi:
What I meant: I fail to see why wrapping the `zap` logger, by embedding it into `utils.Logger`, without wrapping the API is useful. It might be less confusing to either completely wrap the `zap` logger (with it's API), or expose that we are using the `zap`logger.

c633:
It makes sense. Done in 30450192a8366cbff64c0fd27e2b84d7f8745e89. Thanks @Liamsi !

main.go
```diff
@@ -0,0 +1,18 @@
...
+func main() {
+	logger := utils.NewLogger(&utils.LoggerConfig{
+		Environment: "development",
+		Path:        "./test.log",
+	})
+
+	url := "hello"
+
+	logger.Error(fmt.Sprintf("string %s", url))
+}
```
Liamsi: An [example](https://blog.golang.org/examples) might be better suited than a dedicated main here.

c633: Oops, sorry :(.

utils/logger.go
```diff
@@ -0,0 +1,122 @@
+package utils
+
+import (
+	"strings"
+
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+// Logger is the wrapper of zap.SugaredLogger.
```
masomel: Nitpicky: "...the wrapper of..." --> "... a wrapper for..."

utils/logger.go
```diff
@@ -0,0 +1,122 @@
...
+// which is either "development" or "production",
+// and the path of file to write the logging output to.
+// Logging is enabled at DebugLevel and above in development mode,
+// at InfoLevel and above in production mode.
+type LoggerConfig struct {
+	Environment string `toml:"env"`
+	Path        string `toml:"path"`
+}
+
+func NewLogger(conf *LoggerConfig) *Logger {
```
masomel: Add documentation for `NewLogger`?

utils/logger.go
```diff
@@ -0,0 +1,122 @@
...
+		zLevel.SetLevel(zap.InfoLevel)
+	default:
+		panic("Environment must be either development or production")
+	}
+
+	zConfig := &zap.Config{
+		Level:             zLevel,
+		Development:       strings.EqualFold("development", conf.Environment),
+		Encoding:          "json",
+		DisableStacktrace: true, // turn off stacktrace capturing
```
masomel: I'm wondering if it might be useful to enable this feature optionally in dev mode? Maybe pass in a flag that will set this to `false` if the developer chooses?

utils/logger.go
```diff
@@ -0,0 +1,122 @@
...
+
+	logger, err := zConfig.Build()
+	if err != nil {
+		panic(err)
+	}
+	return &Logger{logger.Sugar()}
+}
+
+// Debug logs a message with some additional context
+// addressed by key-value pairs.
```
masomel: The documentation for `Debug`, `Info`, `Warn` and `Error` is identical, which isn't super helpful. Is the difference between these logging levels really just the logging level itself? Maybe we could describe for which situations each logging level is intended? In other words, when does it make sense to call `Warn` vs `Info`? Alternatively, we could pass extra/less context information depending on the logging level? I guess I'm just trying to figure out a way that it's clear for developers (and also ourselves) how to choose which function to use. The same goes for `Panic` and `Fatal`.

c633:
> But I the documentation should be a bit clearer

Done in e18c093f520d58e5e5d058b8c8e622de74f45493. PTAL.

utils/logger.go
c633: Done in 9acfcd11142aa51133d222f9c86e5b70866807ff. It is turned off by default.

utils/logger.go
```diff
@@ -101,8 +108,9 @@ func (l *Logger) Error(msg string, keysAndValues ...interface{}) {
 	}
 }
 
-// Panic logs a message with some additional context
-// addressed by key-value pairs, then panics.
+// Panic logs a message that is severe error event,
```
masomel: Nitpicking: "is *a* severe error"

utils/logger.go
```diff
@@ -0,0 +1,129 @@
...
+	logger, err := zConfig.Build()
+	if err != nil {
+		panic(err)
+	}
+	return &Logger{logger.Sugar()}
+}
+
+// Debug logs a message that is most useful to debug,
+// with some additional context addressed by key-value pairs.
+func (l *Logger) Debug(msg string, keysAndValues ...interface{}) {
```
arlolra: Can you and is there any reason not to just do,

```
func (l *Logger) Debug(all ...interface{}) {
    l.zLogger.Debugw(all...)
}
```

c633: The signature helps the developers know that these could be used like:
`l.Debug(fmt.Sprintf())` or `l.Debug(msg, key, value)`, but not `l.Debug(msg1, msg2)`.

arlolra: How about,

```
func (l *Logger) Debug(msg string, keysAndValues ...interface{}) {
    l.zLogger.Debugw(msg, keysAndValues...)
}
```

arlolra: > The signature helps the developers know that these could be used like

Ah, I see, makes sense.

c633: zap's `Debugw` requires non-nil `keyAndValues`, unfortunately.

arlolra: Alrighty, not a big deal.

c633:
Merged in eafb0ca39cfb3acd8d598201e14f1c7fff324708 with some minor changes: also output log to `stderr` (`LoggerConfig.Path` can be omitted), the output format is "tab-separated" instead of JSON.

PR #148, opened 2016-12-02
Add version commands

c633:


internal/version.go
```diff
@@ -0,0 +1,4 @@
+package internal
+
+// Version of executables.
+var Version = "0.1"
```
arlolra: 0.1.0?

arlolra: Or, maybe start at "0.0.0+git" and then the commit that we tag and release will bump the version?

masomel: I like that approach. Then we don't have to worry about manually editing the file every time.

arlolra:
Merged in bc2c84b3

Issue #151, opened 2016-12-05
Implement auditing


masomel:
Auditing is a protocol necessary for detection of equivocation. It includes an auditor component, a separate entity that is supposed to check a key server's STR history. The idea is that *anybody* should be able to run an auditor (regardless of whether they also run a key server), so this implementation will require an executable.

Auditing protocol server-side:
- Key server sends every new STR to the auditors
- The auditors verify the hash chain

Auditing protocol client-side:
- Client receives new STR from the key server
- Client queries n auditors for the key server's STR for the same epoch
- Client ensures that all STRs are consistent

Open questions:
- Auditor discovery: How do servers and clients find auditors? How do auditors keep track of key servers?
    - It's probably OK to pin known auditors for now
- How many auditors do we need to ensure equivocation detection? See Appendix B of the paper.

c633:
> Key server sends every new STR to the auditors

I thought the auditors just act like a client? If not, do we need a consensus algorithm for the server and auditors?

masomel:
> I thought the auditors just act like a client? 

By this you mean that the auditors will query the key server at the beginning of every epoch like clients do?

> If not, do we need a consensus algorithm for the server and auditors?

I think we might need consensus anyway for keeping track of running auditors (much like Tor uses to keep track of running relays). But the protocol itself shouldn't require any consensus algorithm.

c633:
> By this you mean that the auditors will query the key server at the beginning of every epoch like clients do?

Yes.

> I think we might need consensus anyway for keeping track of running auditors

I think more about the consensus for the STRs between the key server and auditors. I would like to let the keyserver actively broadcast its STRs to the auditors rather than let the auditors query the key server at the beginning of each epoch. I don't know if we go this way, we can avoid the Clock synchronization issue or not ...

LongNguyenHoang89:
Can publishing STRs to a global append-only ledger like blockchain solve the problem ? 
EthIKS already did that didn't they. 
My proposal is that Identity Provider publishes only the STR in a transaction to a blockchain (not the whole data structure like EthIKS). Each client can then implement a lightweight wallet which tracks only their interested transactions from the blockchain network. 
This way, there is no need for separate auditors at all.

Liamsi:
Thanks a lot for your feedback! 

> Each client can then implement a lightweight wallet which tracks only their interested transactions from the blockchain network.

[...]

PR #155, opened 2016-12-15
Client-auditor protocol

masomel:
Part of #151 

TODO:

- [X] client-auditor requests and responses
- [x] auditor state
- [x] tests

### DIFF
--- a/protocol/auditlog.go
+++ b/protocol/auditlog.go
@@ -44,7 +44,7 @@ func NewAuditLog() *ConiksAuditLog {
 
 // IsKnownDirectory checks to see if an entry for the directory
 // address addr exists in the audit log l. IsKnownDirectory() does not
-//  validate the entries themselves. It returns true if an entry exists,
+// validate the entries themselves. It returns true if an entry exists,
 // and false otherwise.
 func (l *ConiksAuditLog) IsKnownDirectory(addr string) bool {
 	h := l.histories[addr]
@@ -54,15 +54,15 @@ func (l *ConiksAuditLog) IsKnownDirectory(addr string) bool {
 	return false
 }
 
-// FIXME: pass Request message as param
+// FIXME: pass Response message as param
 // masomel: will probably want to write a more generic function
 // for "catching up" on a history in case an auditor misses epochs
 func (l *ConiksAuditLog) Insert(addr string, signKey sign.PublicKey,
 	oldSTRs map[uint64]*m.SignedTreeRoot, latestSTR *m.SignedTreeRoot) error {
 
-	// panic if we want to create a new entry for an addr we already know
+	// error if we want to create a new entry for an addr we already know
 	if l.IsKnownDirectory(addr) {
-		panic("[protocol] Trying to add an audit log entry for an existing address")
+		return ErrAuditLog
 	}
 
 	// create the new directory history
@@ -73,11 +73,14 @@ func (l *ConiksAuditLog) Insert(addr string, signKey sign.PublicKey,
 
 	// add each old STR into the history
 	for ep := startEp; ep < endEp; ep++ {
-		if str := oldSTRs[ep]; str != nil {
-			h.snapshots[ep] = str
+		str := oldSTRs[ep]
+		if str == nil {
+			return ErrMalformedDirectoryMessage
 		}
-		return ErrMalformedDirectoryMessage
+		h.snapshots[ep] = str
 	}
+
+	l.histories[addr] = h
 
 	// FIXME: verify the consistency of each new STR
 	return nil
@@ -90,12 +93,12 @@ func (l *ConiksAuditLog) Insert(addr string, signKey sign.PublicKey,
 // check error otherwise. Update() assumes that Insert() has been called for
 // addr prior to its first call and thereby expects that an entry for addr
 // exists in the audit log l.
-// FIXME: pass Request message as param
+// FIXME: pass Response message as param
 func (l *ConiksAuditLog) Update(addr string, newSTR *m.SignedTreeRoot) error {
 
 	// panic if we want to update an entry for which we don't have
 	if !l.IsKnownDirectory(addr) {
-		panic("[protocol] Trying to update audit log entry for non-existent address")
+		return ErrAuditLog
 	}
 
 	h := l.histories[addr]
--- a/protocol/auditlog.go
+++ b/protocol/auditlog.go
@@ -128,8 +128,23 @@ func (h *directoryHistory) verifySTRConsistency(str *m.SignedTreeRoot) error {
 	return CheckBadSTR
 }
 
-// TODO add doc
-func (l *ConiksAuditLog) Query(req *AuditingRequest) (*Response, ErrorCode) {
+// GetObservedSTR gets the observed STR for the CONIKS directory address indicated
+// in the AuditingRequest req received from a CONIKS client from the auditor's latest
+// directory history entry, and returns a tuple of the form
+// (response, error).
+// The response (which also includes the error code) is supposed to
+// be sent back to the client. The returned error is used by the auditor
+// for logging purposes.
+//
+// A request without a directory address is considered
+// malformed, and causes GetObservedSTR() to return a
+// message.NewErrorResponse(ErrMalformedClientMessage) tuple.
+// GetObservedSTR() returns a message.NewObservedSTR(str) tuple.
+// str is the signed tree root the auditor has observed for the latest epoch.
+// If the auditor doesn't have any history entries for the requested CONIKS
+// directory, GetObservedSTR() returns a
+// message.NewErrorResponse(ReqUnknownDirectory) tuple.
+func (l *ConiksAuditLog) GetObservedSTR(req *AuditingRequest) (*Response, ErrorCode) {
 
 	// make sure the request is well-formed
 	if len(req.DirectoryAddr) <= 0 {
@@ -141,12 +156,29 @@ func (l *ConiksAuditLog) Query(req *AuditingRequest) (*Response, ErrorCode) {
 		return NewObservedSTR(h.latestSTR)
 	}
 
-	// FIXME: return ErrUnknownDirectory
-	return nil, ErrDirectory
+	return NewErrorResponse(ReqUnknownDirectory), ReqUnknownDirectory
 }
 
-// TODO: add doc
-func (l *ConiksAuditLog) QueryInEpoch(req *AuditingInEpochRequest) (*Response,
+// GetObservedSTRInEpoch gets the observed STR for the CONIKS directory address
+// for a prior directory history entry indicated in the
+// AuditingInEpochRequest req received from a CONIKS client,
+// and returns a tuple of the form (response, error).
+// The response (which also includes the error code) is supposed to
+// be sent back to the client. The returned error is used by the auditor
+// for logging purposes.
+//
+// A request without a directory address or with an epoch greater than the latest
+// observed epoch of this directory is considered malformed, and causes
+// GetObservedSTRInEpoch() to return a
+// message.NewErrorResponse(ErrMalformedClientMessage) tuple.
+// GetObservedSTRInEpoch() returns a message.NewObservedSTRs(strs) tuple.
+// strs is a list of STRs for the epoch range [ep,
+// l.histories[req.DirectoryAddr].latestSTR.Epoch], where ep is the past epoch
+// for which the client has requested the observed STR.
+// If the auditor doesn't have any history entries for the requested CONIKS
+// directory, GetObservedSTR() returns a
+// message.NewErrorResponse(ReqUnknownDirectory) tuple.
+func (l *ConiksAuditLog) GetObservedSTRInEpoch(req *AuditingInEpochRequest) (*Response,
 	ErrorCode) {
 
 	// make sure the request is well-formed
@@ -178,6 +210,5 @@ func (l *ConiksAuditLog) QueryInEpoch(req *AuditingInEpochRequest) (*Response,
 		return NewObservedSTRs(strs)
 	}
 
-	// FIXME: return ErrUnknownDirectory
-	return nil, ErrDirectory
+	return NewErrorResponse(ReqUnknownDirectory), ReqUnknownDirectory
 }
--- a/protocol/auditlog.go
+++ b/protocol/auditlog.go
@@ -100,7 +100,7 @@ func (l *ConiksAuditLog) Update(addr string, newSTR *m.SignedTreeRoot) error {
 
 	h := l.histories[addr]
 
-	if err := h.verifySTRConsistency(newSTR); err != nil {
+	if err := verifySTRConsistency(h.signKey, h.latestSTR, newSTR); err != nil {
 		return err
 	}
 
@@ -108,24 +108,6 @@ func (l *ConiksAuditLog) Update(addr string, newSTR *m.SignedTreeRoot) error {
 	h.snapshots[h.latestSTR.Epoch] = h.latestSTR
 	h.latestSTR = newSTR
 	return nil
-}
-
-// verifySTRConsistency checks the consistency between 2 snapshots.
-// It uses the pinned signing key in the directory history
-// to verify the STR's signature and verifies
-// the hash chain using the latestSTR stored in the history.
-// TODO: dedup this: write generic verifySTRConsistency
-func (h *directoryHistory) verifySTRConsistency(str *m.SignedTreeRoot) error {
-	// verify STR's signature
-	if !h.signKey.Verify(str.Serialize(), str.Signature) {
-		return CheckBadSignature
-	}
-	if str.VerifyHashChain(h.latestSTR) {
-		return nil
-	}
-
-	// TODO: verify the directory's policies as well. See #115
-	return CheckBadSTR
 }
 
 // GetObservedSTR gets the observed STR for the CONIKS directory address indicated
--- a/protocol/auditlog.go
+++ b/protocol/auditlog.go
@@ -152,28 +152,29 @@ func (l *ConiksAuditLog) GetObservedSTRs(req *AuditingRequest) (*Response,
 			ErrMalformedClientMessage
 	}
 
-	if h := l.histories[req.DirectoryAddr]; h != nil {
+	h := l.histories[req.DirectoryAddr]
 
-		// also make sure the epoch is well-formed
-		if req.Epoch > h.latestSTR.Epoch {
-			return NewErrorResponse(ErrMalformedClientMessage),
-				ErrMalformedClientMessage
-		}
-
-		var strs []*DirSTR
-		startEp := req.Epoch
-		endEp := h.latestSTR.Epoch
-
-		for ep := startEp; ep < endEp; ep++ {
-			str := h.snapshots[ep]
-			strs = append(strs, str)
-		}
-
-		// don't forget to append the latest STR
-		strs = append(strs, h.latestSTR)
-
-		return NewSTRHistoryRange(strs)
+	if h == nil {
+		return NewErrorResponse(ReqUnknownDirectory), ReqUnknownDirectory
 	}
 
-	return NewErrorResponse(ReqUnknownDirectory), ReqUnknownDirectory
+	// also make sure the epoch is well-formed
+	if req.Epoch > h.latestSTR.Epoch {
+		return NewErrorResponse(ErrMalformedClientMessage),
+			ErrMalformedClientMessage
+	}
+
+	var strs []*DirSTR
+	startEp := req.Epoch
+	endEp := h.latestSTR.Epoch
+
+	for ep := startEp; ep < endEp; ep++ {
+		str := h.snapshots[ep]
+		strs = append(strs, str)
+	}
+
+	// don't forget to append the latest STR
+	strs = append(strs, h.latestSTR)
+
+	return NewSTRHistoryRange(strs)
 }
--- a/protocol/auditlog.go
+++ b/protocol/auditlog.go
@@ -25,9 +25,7 @@ type directoryHistory struct {
 // public signing key enabling the auditor to verify the corresponding
 // signed tree roots, and a list with all observed snapshots in
 // chronological order.
-type ConiksAuditLog struct {
-	histories map[[crypto.HashSizeByte]byte]*directoryHistory
-}
+type ConiksAuditLog map[[crypto.HashSizeByte]byte]*directoryHistory
 
 // computeInitSTRHash is a wrapper for the digest function;
 // returns nil if the STR isn't an initial STR (i.e. str.Epoch != 0)
@@ -35,27 +33,27 @@ func computeInitSTRHash(initSTR *DirSTR) [crypto.HashSizeByte]byte {
 	var initSTRHash [crypto.HashSizeByte]byte
 
 	if initSTR.Epoch == 0 {
-		// ugh HACK: need to call copy to convert slice into the array
+		// ugh HACK: need to call copy to convert slice
+		// into the array
 		copy(initSTRHash[:], crypto.Digest(initSTR.Serialize()))
 	}
 	return initSTRHash
 }
 
-func newDirectoryHistory(dirName string, signKey sign.PublicKey, str *DirSTR) *directoryHistory {
+func newDirectoryHistory(dirName string, signKey sign.PublicKey) *directoryHistory {
 	h := new(directoryHistory)
 	h.dirName = dirName
 	h.signKey = signKey
 	h.snapshots = make(map[uint64]*DirSTR)
-	h.latestSTR = str
+	h.latestSTR = nil
 	return h
 }
 
 // NewAuditLog constructs a new ConiksAuditLog. It creates an empty
 // log; the auditor will add an entry for each CONIKS directory
 // the first time it observes an STR for that directory.
-func NewAuditLog() *ConiksAuditLog {
-	l := new(ConiksAuditLog)
-	l.histories = make(map[[crypto.HashSizeByte]byte]*directoryHistory)
+func NewAuditLog() ConiksAuditLog {
+	l := make(map[[crypto.HashSizeByte]byte]*directoryHistory)
 	return l
 }
 
@@ -64,20 +62,26 @@ func NewAuditLog() *ConiksAuditLog {
 // in the audit log l. IsKnownDirectory() does not
 // validate the entries themselves. It returns true if an entry exists,
 // and false otherwise.
-func (l *ConiksAuditLog) IsKnownDirectory(dirInitHash [crypto.HashSizeByte]byte) bool {
-	h := l.histories[dirInitHash]
+func (l ConiksAuditLog) IsKnownDirectory(dirInitHash [crypto.HashSizeByte]byte) bool {
+	h := l[dirInitHash]
 	if h != nil {
 		return true
 	}
 	return false
 }
 
+// updateLatestSTR inserts a new STR into a directory history;
+// assumes the STR has been validated by the caller
+func (h *directoryHistory) updateLatestSTR(newLatest *DirSTR) {
+	h.snapshots[newLatest.Epoch] = newLatest
+	h.latestSTR = newLatest
+}
+
 // Insert creates a new directory history for the key directory addr
 // and inserts it into the audit log l.
 // The directory history is initialized with the key directory's
-// signing key signKey, a list of STRs representing the
-// directory's prior history oldSTRs, and the directory's latest STR
-// latestSTR.
+// signing key signKey, and a list of STRs representing the
+// directory's STR history so far (as a map of epochs to STRs).
 // Insert() returns an ErrAuditLog if the auditor attempts to create
 // a new history for a known directory, an ErrMalformedDirectoryMessage
 // if oldSTRs is malformed, and nil otherwise.
@@ -86,20 +90,16 @@ func (l *ConiksAuditLog) IsKnownDirectory(dirInitHash [crypto.HashSizeByte]byte)
 // FIXME: pass Response message as param
 // masomel: will probably want to write a more generic function
 // for "catching up" on a history in case an auditor misses epochs
-func (l *ConiksAuditLog) Insert(addr string, signKey sign.PublicKey,
-	oldSTRs map[uint64]*DirSTR, latestSTR *DirSTR) error {
+func (l ConiksAuditLog) Insert(addr string, signKey sign.PublicKey,
+	hist map[uint64]*DirSTR) error {
 
-	// make sure we have oldEpochs unless we're inserting right
-	// at the start of a directory's history
-	if oldSTRs == nil && latestSTR.Epoch > 0 {
+	// make sure we're getting an initial STR at the very least
+	if len(hist) < 1 && hist[0].Epoch != 0 {
 		return ErrMalformedDirectoryMessage
 	}
 
 	// compute the hash of the initial STR
-	dirInitHash := computeInitSTRHash(latestSTR)
-	if oldSTRs != nil {
-		dirInitHash = computeInitSTRHash(oldSTRs[0])
-	}
+	dirInitHash := computeInitSTRHash(hist[0])
 
 	// error if we want to create a new entry for a directory
 	// we already know
@@ -108,28 +108,30 @@ func (l *ConiksAuditLog) Insert(addr string, signKey sign.PublicKey,
 	}
 
 	// create the new directory history
-	h := newDirectoryHistory(addr, signKey, latestSTR)
+	h := newDirectoryHistory(addr, signKey)
 
 	startEp := uint64(0)
-	endEp := latestSTR.Epoch
+	endEp := uint64(len(hist))
 
-	// add each old STR into the history
-	// This loop automatically catches if oldSTRs is malformed
-	// (i.e. oldSTRs starts at epoch 0 and
-	// len(oldSTRs) != latestSTR.Epoch-1)
+	// add each STR into the history
+	// This loop automatically catches if hist is malformed
+	// (i.e. hist is missing and epoch)
+	// FIXME: verify the consistency of each new STR before inserting
+	// into the audit log
 	for ep := startEp; ep < endEp; ep++ {
-		str := oldSTRs[ep]
+		str := hist[ep]
 		if str == nil {
 			return ErrMalformedDirectoryMessage
 		}
 		h.snapshots[ep] = str
 	}
 
-	// don't forget to add the latest STR
-	h.snapshots[endEp] = latestSTR
-	l.histories[dirInitHash] = h
+	// Make sure to update the latestSTR
+	// in this particular call, the latestSTR has already been
+	// inserted into the snapshots map in the loop above
+	h.updateLatestSTR(hist[endEp-1])
+	l[dirInitHash] = h
 
-	// FIXME: verify the consistency of each new STR
 	return nil
 }
 
@@ -141,22 +143,21 @@ func (l *ConiksAuditLog) Insert(addr string, signKey sign.PublicKey,
 // addr prior to its first call and thereby expects that an entry for addr
 // exists in the audit log l.
 // FIXME: pass Response message as param
-func (l *ConiksAuditLog) Update(dirInitHash [crypto.HashSizeByte]byte, newSTR *DirSTR) error {
+func (l ConiksAuditLog) Update(dirInitHash [crypto.HashSizeByte]byte, newSTR *DirSTR) error {
 
 	// error if we want to update the entry for an addr we don't know
 	if !l.IsKnownDirectory(dirInitHash) {
 		return ErrAuditLog
 	}
 
-	h := l.histories[dirInitHash]
+	h := l[dirInitHash]
 
 	if err := verifySTRConsistency(h.signKey, h.latestSTR, newSTR); err != nil {
 		return err
 	}
 
 	// update the latest STR
-	h.snapshots[h.latestSTR.Epoch] = h.latestSTR
-	h.latestSTR = newSTR
+	h.updateLatestSTR(newSTR)
 	return nil
 }
 
@@ -178,7 +179,7 @@ func (l *ConiksAuditLog) Update(dirInitHash [crypto.HashSizeByte]byte, newSTR *D
 // If the auditor doesn't have any history entries for the requested CONIKS
 // directory, GetObservedSTRs() returns a
 // message.NewErrorResponse(ReqUnknownDirectory) tuple.
-func (l *ConiksAuditLog) GetObservedSTRs(req *AuditingRequest) (*Response,
+func (l ConiksAuditLog) GetObservedSTRs(req *AuditingRequest) (*Response,
 	ErrorCode) {
 
 	// make sure we have a history for the requested directory in the log
@@ -186,7 +187,7 @@ func (l *ConiksAuditLog) GetObservedSTRs(req *AuditingRequest) (*Response,
 		return NewErrorResponse(ReqUnknownDirectory), ReqUnknownDirectory
 	}
 
-	h := l.histories[req.DirInitSTRHash]
+	h := l[req.DirInitSTRHash]
 
 	// make sure the request is well-formed
 	if req.EndEpoch > h.latestSTR.Epoch || req.StartEpoch > req.EndEpoch {
@@ -195,13 +196,10 @@ func (l *ConiksAuditLog) GetObservedSTRs(req *AuditingRequest) (*Response,
 	}
 
 	var strs []*DirSTR
-	for ep := req.StartEpoch; ep < req.EndEpoch; ep++ {
+	for ep := req.StartEpoch; ep <= req.EndEpoch; ep++ {
 		str := h.snapshots[ep]
 		strs = append(strs, str)
 	}
-
-	// don't forget to append the STR for EndEpoch
-	strs = append(strs, h.snapshots[req.EndEpoch])
 
 	return NewSTRHistoryRange(strs)
 }
--- a/protocol/auditlog.go
+++ b/protocol/auditlog.go
@@ -123,41 +123,9 @@ func (l *ConiksAuditLog) Update(addr string, newSTR *DirSTR) error {
 	return nil
 }
 
-// GetObservedSTR gets the observed STR for the CONIKS directory address
-// indicated in the AuditingRequest req received from a CONIKS client from
-// the auditor's latest
-// directory history entry, and returns a tuple of the form
-// (response, error).
-// The response (which also includes the error code) is supposed to
-// be sent back to the client. The returned error is used by the auditor
-// for logging purposes.
-//
-// A request without a directory address is considered
-// malformed, and causes GetObservedSTR() to return a
-// message.NewErrorResponse(ErrMalformedClientMessage) tuple.
-// GetObservedSTR() returns a message.NewObservedSTR(str) tuple.
-// str is the signed tree root the auditor has observed for the latest epoch.
-// If the auditor doesn't have any history entries for the requested CONIKS
-// directory, GetObservedSTR() returns a
-// message.NewErrorResponse(ReqUnknownDirectory) tuple.
-func (l *ConiksAuditLog) GetObservedSTR(req *AuditingRequest) (*Response, ErrorCode) {
-
-	// make sure the request is well-formed
-	if len(req.DirectoryAddr) <= 0 {
-		return NewErrorResponse(ErrMalformedClientMessage),
-			ErrMalformedClientMessage
-	}
-
-	if h := l.histories[req.DirectoryAddr]; h != nil {
-		return NewObservedSTR(h.latestSTR)
-	}
-
-	return NewErrorResponse(ReqUnknownDirectory), ReqUnknownDirectory
-}
-
-// GetObservedSTRInEpoch gets the observed STR for the CONIKS directory
-// address for a prior directory history entry indicated in the
-// AuditingInEpochRequest req received from a CONIKS client,
+// GetObservedSTRs gets the observed STR for the CONIKS directory
+// address for a directory history entry indicated in the
+// AuditingRequest req received from a CONIKS client,
 // and returns a tuple of the form (response, error).
 // The response (which also includes the error code) is supposed to
 // be sent back to the client. The returned error is used by the auditor
@@ -165,16 +133,17 @@ func (l *ConiksAuditLog) GetObservedSTR(req *AuditingRequest) (*Response, ErrorC
 //
 // A request without a directory address or with an epoch greater than
 // the latest observed epoch of this directory is considered malformed,
-// and causes GetObservedSTRInEpoch() to return a
+// and causes GetObservedSTRs() to return a
 // message.NewErrorResponse(ErrMalformedClientMessage) tuple.
-// GetObservedSTRInEpoch() returns a message.NewObservedSTRs(strs) tuple.
+// GetObservedSTRs() returns a message.NewSTRList(strs) tuple.
 // strs is a list of STRs for the epoch range [ep,
-// l.histories[req.DirectoryAddr].latestSTR.Epoch], where ep is the past
-// epoch for which the client has requested the observed STR.
+// l.histories[req.DirectoryAddr].latestSTR.Epoch], where ep is the epoch for
+// which the client has requested the observed STR; i.e. if ep == the latest epoch,
+// the list returned is of length 1.
 // If the auditor doesn't have any history entries for the requested CONIKS
-// directory, GetObservedSTR() returns a
+// directory, GetObservedSTRs() returns a
 // message.NewErrorResponse(ReqUnknownDirectory) tuple.
-func (l *ConiksAuditLog) GetObservedSTRInEpoch(req *AuditingInEpochRequest) (*Response,
+func (l *ConiksAuditLog) GetObservedSTRs(req *AuditingRequest) (*Response,
 	ErrorCode) {
 
 	// make sure the request is well-formed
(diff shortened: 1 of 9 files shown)


c633:
My most concern here is how will this fit well with other auditor types (like Cosi and Blockchain). But I'm fine with this PR and happy to reorganize our code as things come later.

masomel:
> My most concern here is how will this fit well with other auditor types (like Cosi and Blockchain).

Right, all of these auditor types are very distinct, but I think CoSi auditors and blockchain auditors are more extensions to the auditing protocol instead of complete replacements. 

> But I'm fine with this PR and happy to reorganize our code as things come later.

I'm not sure what you have in mind. Is your issue that the auditor protocol is part of the core protocol? Because as I said above, I think it should be. If/when we have more auditing options, like you said, we can re-think whether this makes sense.

c633:
> Is your issue that the auditor protocol is part of the core protocol? Because as I said above, I think it should be.

Yes, I agree that too.

protocol/auditlog.go
```diff
@@ -0,0 +1,196 @@
...
+// and false otherwise.
+func (l *ConiksAuditLog) IsKnownDirectory(addr string) bool {
+	h := l.histories[addr]
+	if h != nil {
+		return true
+	}
+	return false
+}
+
+// FIXME: pass Request message as param
```
masomel: Ah, you're right, this should be `Response`. I also like the idea of having a generic `HandleDirectorySTRs` function since there is a lot of potentially shared STR verification code between the auditor and the client.  But `Insert`, `Update` and `InsertRange` are specific to the auditor log, though, so I think it makes sense to keep those separate from `consistencychecks.go`.

protocol/auditlog.go
```diff
@@ -0,0 +1,196 @@
...
+	ErrorCode) {
+
+	// make sure the request is well-formed
+	if len(req.DirectoryAddr) <= 0 {
+		return NewErrorResponse(ErrMalformedClientMessage),
+			ErrMalformedClientMessage
+	}
+
+	if h := l.histories[req.DirectoryAddr]; h != nil {
+
```
masomel: Can you please point to code in `directory.go` where this happens?

protocol/message.go
```diff
@@ -216,6 +260,40 @@ func NewMonitoringProof(ap []*m.AuthenticationPath,
...
+func NewObservedSTRs(str []*m.SignedTreeRoot) (*Response, ErrorCode) {
+	return &Response{
+		Error: ReqSuccess,
+		DirectoryResponse: &ObservedSTRs{
+			STR: str,
+		},
+	}, ReqSuccess
+}
+
 func (msg *Response) validate() error {
```
masomel: Yes, I think this makes sense, especially if we're going to have some generic `HandleSTRs` function.

protocol/auditlog.go
```diff
@@ -0,0 +1,196 @@
...
+// enabling the auditor to verify the corresponding signed
+// tree roots.
+type ConiksAuditLog struct {
+	histories map[string]*directoryHistory
+}
+
+func newDirectoryHistory(signKey sign.PublicKey, str *m.SignedTreeRoot) *directoryHistory {
+	h := new(directoryHistory)
+	h.signKey = signKey
+	h.snapshots = make(map[uint64]*m.SignedTreeRoot)
```
masomel: Is there a big advantage to including the initially observed `str` in `snapshots` right away?

masomel:
@c633 @Liamsi @arlolra Can you please TAL at this?

protocol/auditlog.go
c633: > But Insert, Update and InsertRange are specific to the auditor log, though, so I think it makes sense to keep those separate from consistencychecks.go.

Yes, of course.

protocol/auditlog.go
c633: Hmm, no. I just thought it should be here at the first glance. But it doesn't matter :p

protocol/auditlog.go
c633: Here we go: https://github.com/coniks-sys/coniks-go/blob/master/protocol/directory.go#L201-L215. I meant we return `NewErrorResponse` earlier.

protocol/consistencychecks.go
```diff
@@ -100,6 +100,50 @@ func (cc *ConsistencyChecks) HandleResponse(requestType int, msg *Response,
 	return CheckPassed
 }
 
+// handleDirectorySTRs is supposed to be used by CONIKS clients to
+// handle auditor responses, and by CONIKS auditors to handle directory responses.
```
c633: Hmm, I'm not sure if this is supposed to be here since the file header states `Implements all consistency check operations done by a CONIKS client on data received from a CONIKS directory.`. Maybe we should move `verifySTR` and `verifySTRConsistency` to `str.go`? If you agree, we can add a TODO here.

masomel: I realized this as well. I've started working on creating a more generic `auditor` module that contains all the functionality clients and auditors need to verify/audit the STR consistency. What do you think? But I also think this belongs in a different PR since it requires for a `verifySTRConsistencyInEpoch` to be implemented. 

I know you were already working on this when implementing the monitoring protocol. Do you think it would make sense to implement all the STR-related verification separate from the monitoring?

c633: Yeah, I agree! I'll separate these code and open PR for each of them.

masomel: I've started this in https://github.com/coniks-sys/coniks-go/tree/generic-auditor What do you think?

protocol/auditlog.go
```diff
@@ -0,0 +1,199 @@
...
+	h := l.histories[addr]
+	if h != nil {
+		return true
+	}
+	return false
+}
+
+// FIXME: pass Response message as param
+// masomel: will probably want to write a more generic function
+// for "catching up" on a history in case an auditor misses epochs
```
Liamsi: Before we merge this, the comments should follow golang's documentation convention (start with `Insert`). The `FIXME` can stay but shouldn't be the first thing one gets to read. The comment should probably briefly state what the difference/relation between `Insert` and `Update` is.

protocol/auditlog.go
```diff
@@ -0,0 +1,199 @@
...
+// the directory addr, and inserts the newSTR into addr's directory history
+// if the checks (i.e. STR signature and hash chain verifications) pass.
+// Update() returns nil if the checks pass, and the appropriate consistency
+// check error otherwise. Update() assumes that Insert() has been called for
+// addr prior to its first call and thereby expects that an entry for addr
+// exists in the audit log l.
+// FIXME: pass Response message as param
+func (l *ConiksAuditLog) Update(addr string, newSTR *m.SignedTreeRoot) error {
+
+	// panic if we want to update an entry for which we don't have
```
Liamsi: "... for which we don't know the directory" (?). Also, we don't really `panic` here, right?

protocol/auditlog_test.go
```diff
@@ -0,0 +1,220 @@
...
+	aud := NewAuditLog()
+	err := aud.Insert("test-server", pk, nil, d.LatestSTR())
+	if err != nil {
+		t.Fatal("Error inserting new server history")
+	}
+
+	res, err := aud.GetObservedSTR(&AuditingRequest{
+		DirectoryAddr: "test-server"})
+	obs := res.DirectoryResponse.(*ObservedSTR)
+	if err != ReqSuccess {
```
Liamsi: nitpicking: directly catch the error after it occurs (move this one line up)

protocol/consistencychecks.go
```diff
@@ -2,6 +2,7 @@
 // on data received from a CONIKS directory.
 // These include data binding proof verification,
 // and non-equivocation checks.
+// TODO: move all STR-verifying functionality to a separate module
```
Liamsi: Could we open an issue for that so we won't forget about it? Or is this planned to happen in this PR?

protocol/consistencychecks.go
```diff
@@ -100,6 +101,50 @@ func (cc *ConsistencyChecks) HandleResponse(requestType int, msg *Response,
 	return CheckPassed
 }
 
+// handleDirectorySTRs is supposed to be used by CONIKS clients to
```
Liamsi: Case doesn't match with method signature (handleDirectorySTRs vs HandleDirectorySTRs).

protocol/auditlog.go
masomel: Ah yes, thanks for pointing this out!

protocol/auditlog.go
masomel: No, we don't. I forgot to update the comment here.

protocol/consistencychecks.go
masomel: @c633 and I are already working on this in a separate branch, but we haven't opened the PR yet.

protocol/message.go
```diff
@@ -229,6 +309,16 @@ func (msg *Response) validate() error {
 	case *DirectoryProofs:
 		// TODO: also do above assertions here
 		return nil
+	case *ObservedSTR:
+		if df.STR == nil {
+			return ErrMalformedAuditorMessage
+		}
+		return nil
+	case *ObservedSTRs:
+		if df.STR == nil || len(df.STR) < 1 {
```
c633: This can be simplified: `if len(df.STR) == 0`

protocol/consistencychecks.go
```diff
@@ -140,12 +189,15 @@ func (cc *ConsistencyChecks) verifySTR(str *DirSTR) error {
...
 // verifySTRConsistency checks the consistency between 2 snapshots.
-// It uses the pinned signing key in cc
-// to verify the STR's signature and should not verify
-// the hash chain using the STR stored in cc.
-func (cc *ConsistencyChecks) verifySTRConsistency(savedSTR, str *DirSTR) error {
+// It uses the signing key signKey to verify the STR's signature.
+// The signKey param either comes from a client's
+// pinned signing key in cc, or an auditor's pinned signing key
+// in its history.
+// In the case of a client-side consistency check, verifySTRConsistency()
```
c633: We can remove this line of comment since we're no longer used `cc` anymore.

protocol/consistencychecks.go
```diff
@@ -140,12 +189,15 @@ func (cc *ConsistencyChecks) verifySTR(str *DirSTR) error {
 }
 
 // verifySTRConsistency checks the consistency between 2 snapshots.
-// It uses the pinned signing key in cc
-// to verify the STR's signature and should not verify
-// the hash chain using the STR stored in cc.
-func (cc *ConsistencyChecks) verifySTRConsistency(savedSTR, str *DirSTR) error {
+// It uses the signing key signKey to verify the STR's signature.
+// The signKey param either comes from a client's
+// pinned signing key in cc, or an auditor's pinned signing key
```
c633: Should just be `pinned signing key, or an ...` (no more `cc` here).

protocol/consistencychecks.go
```diff
@@ -100,6 +101,51 @@ func (cc *ConsistencyChecks) HandleResponse(requestType int, msg *Response,
 	return CheckPassed
 }
 
+// HandleDirectorySTRs is supposed to be used by CONIKS clients to
+// handle auditor responses, and by CONIKS auditors to handle directory
+// responses.
+func HandleDirectorySTRs(requestType int, msg *Response, signKey sign.PublicKey,
```
c633: Since it is being used nowhere and we are working on this in #170, I would like to remove it from this PR.

protocol/message.go
```diff
@@ -90,6 +92,34 @@ type MonitoringRequest struct {
 	EndEpoch   uint64 `json:"end_epoch"`
 }
 
+// An AuditingRequest is a message with a CONIKS key directory's address
+// as a string that a CONIKS client sends to a CONIKS auditor, or a CONIKS auditor
```
c633: Is this supposed to be used for a auditor to request to a directory? If so, the `DirectoryAddr` field seems to be redundant.

protocol/message.go
```diff
@@ -90,6 +92,34 @@ type MonitoringRequest struct {
...
+// must send an AuditingInEpochRequest.
+//
+// The response to a successful request is an ObservedSTR.
+type AuditingRequest struct {
+	DirectoryAddr string `json:"directory_addr"`
+}
+
+// An AuditingInEpochRequest is a message with a key directory's address
+// as a string and an epoch as a uint64 that a CONIKS client sends to
+// a CONIKS auditor, or a CONIKS auditor sends to a CONIKS directory,
```
c633: Same as https://github.com/coniks-sys/coniks-go/pull/155/files#r105599786.

protocol/message.go
```diff
@@ -216,6 +262,40 @@ func NewMonitoringProof(ap []*m.AuthenticationPath,
 	}, ReqSuccess
 }
 
+// NewObservedSTR creates the response message a CONIKS auditor
+// sends to a client, or a CONIKS directory sends to an auditor,
```
c633: If we use a pub-sub architecture, will this be used between an auditor and a directory?

protocol/message.go
masomel: When I first wrote this, it was supposed to be used by both clients and auditors, but this doesn't really make sense. We should have separate message types for auditor-server communication and client-auditor communication.

protocol/message.go
masomel: Won't this throw a null pointer error if we don't assert `df.STR != nil`?

Liamsi:
As long as `df` is not `nil` there would be no exception as `len` returns 0
if the argument is `nil`
https://golang.org/pkg/builtin/#len

Am 04.04.2017 11:05 nachm. schrieb "Marcela Melara" <
notifications@github.com>:

protocol/message.go
masomel: I think in either case, pub-sub or request-response, the implementation of the corresponding auditor-directory message would be the same, so it would make sense to use the same function to return a list of requested STRs. Similarly, the response message type for a client-auditor request and an auditor-server request for STRs should be the same type. This being said, since this PR is only for client-auditor communication, I won't add relevant auditor-directory documentation yet.

protocol/message.go
```diff
@@ -90,6 +91,18 @@ type MonitoringRequest struct {
...
 
+// An AuditingRequest is a message with a CONIKS key directory's address
+// as a string and an epoch as a uint64 that a CONIKS client sends to
+// a CONIKS auditor to request the given directory's STR at the given
+// epoch.
+//
+// The response to a successful request is an ObservedSTRs with
+// a list of STRs covering the epoch range [Epoch, d.LatestSTR().Epoch].
+type AuditingRequest struct {
+	DirectoryAddr string `json:"directory_addr"`
```
c633: Please remove the field tags as we did in 77350b65b486ad6b46b6604d49606319f7e23b28.

protocol/message.go
```diff
@@ -124,15 +137,23 @@ type DirectoryProofs struct {
 	STR []*DirSTR
 }
 
+// An STRList response includes a list of signed tree roots
+// STR. A CONIKS auditor returns this DirectoryResponse type upon an
+// AudutingRequest.
+type STRList struct {
```
c633: Why did you change this? I think `ObservedSTRs` sounds better to me :D

protocol/message.go
masomel: Thanks for the reminder. I'm in the process of rebasing onto master, so I'll remove the tags as part of that.

protocol/message.go
masomel: I also liked `ObservedSTRs` a bit better :) but I changed it because of https://github.com/coniks-sys/coniks-go/pull/155/files/4ec8fb73df460e31d2ab3e35da49da2c6fcaaede#diff-2d3b2346117a9f1fc92ea4a92e285e69

This struct will also be used by the directory to return STRs to the auditor, so I wanted to create a more generic data structure.

c633: Yaa, naming is hard :( Hopefully we can find a better name at the end :D

masomel: Yup :) `STRList` isn't super meaningful, although it's an accurate description of the contents of the message. Because it's a range in the STR history, I was thinking something like `STRHistoryRange` could be a bit more meaningful?

c633: I think it would be better if we have something match with current `DirectoryProof(s)`. The reason we chose this name was to show the purpose of the response (I'm not sure 100% about this). Here are some come to my mind: `STRChain` or `ConsistencyProof`.

protocol/auditlog.go
c633: @masomel: Please fix this.

masomel: PTAL at https://github.com/coniks-sys/coniks-go/pull/155/commits/42da6b31949e2fa5a097b42e2d755b1b588f21db

protocol/auditlog_test.go
masomel: Actually, I don't think that would work because I use `res` below for other operations.

c633: Seems like this has been removed?

protocol/str.go
```diff
@@ -19,7 +19,7 @@ func NewDirSTR(str *merkletree.SignedTreeRoot) *DirSTR {
 
 // Serialize overrides merkletree.SignedTreeRoot.Serialize
 func (str *DirSTR) Serialize() []byte {
-	return append(str.SerializeInternal(), str.Policies.Serialize()...)
+	return append(str.SignedTreeRoot.SerializeInternal(), str.Policies.Serialize()...)
```
c633: Why did you make this change? I think it can be simplified to `str.SerializeInternal()`.

c633:
Just one more question and I think we're good.

protocol/str.go
masomel: I was getting an error during testing on my machine. I changed how I was testing and the error no longer appears so I'll change it back.

protocol/auditlog_test.go
masomel: Yes, this has been updated.

masomel:
@arlolra Any comments before we merge?

protocol/auditlog.go
```diff
@@ -0,0 +1,180 @@
...
+// message.NewErrorResponse(ErrMalformedClientMessage) tuple.
+// GetObservedSTRs() returns a message.NewSTRHistoryRange(strs) tuple.
+// strs is a list of STRs for the epoch range [ep,
+// l.histories[req.DirectoryAddr].latestSTR.Epoch], where ep is the epoch for
+// which the client has requested the observed STR; i.e. if ep == the latest epoch,
+// the list returned is of length 1.
+// If the auditor doesn't have any history entries for the requested CONIKS
+// directory, GetObservedSTRs() returns a
+// message.NewErrorResponse(ReqUnknownDirectory) tuple.
+func (l *ConiksAuditLog) GetObservedSTRs(req *AuditingRequest) (*Response,
```
c633: @masomel: I think for this request, the auditor should return a single STR indicated by `ep`, this way, the client could avoid the case where the auditor returns a more recent STR. But it is also hard for the client to know whether it has the latest STR.
And does the client need to compare every STR in the received range from the server with the auditors or just the latest STR?

protocol/auditlog.go
```diff
@@ -0,0 +1,180 @@
...
+	latestSTR *DirSTR
+}
+
+// A ConiksAuditLog maintains the histories
+// of all CONIKS directories known to a CONIKS auditor.
+// Each history includes the directory's public signing key
+// enabling the auditor to verify the corresponding signed
+// tree roots.
+type ConiksAuditLog struct {
+	histories map[string]*directoryHistory
```
c633: Instead of indexing by key server's address, can we use the hash of the key server's public signing key?

masomel: I hadn't thought of that, but I like this idea. My only concern is that indexing by the hash instead of the server address could have an impact on the auditor's performance (i.e. needing to do a hash for every lookup).

protocol/auditlog.go
masomel: In an earlier iteration we had two separate requests, one for the most recent STR and one for the `STR in epoch`. But  we then discussed that it made sense for the client to use the same `Request` message to either obtain the latest STR or catch up on missed epochs. (I couldn't find where we had that discussion, not sure if it was by email or on Github). 

> this way, the client could avoid the case where the auditor returns a more recent STR. But it is also hard for the client to know whether it has the latest STR.

Is the concern here that the client could receive an STR from the auditor than the STR it received from the server during the last monitoring? The only way this could happen is if the client finishes with monitoring very close to/at the end of an epoch. We wouldn't want the client to trigger a new monitoring cycle every time it receives a mroe recent STR from the auditor since a malicious server could use that to DoS the client. The solution to this issue to me would be to ensure that the client has a way of ensuring it does monitoring at the beginning of en epoch.

> And does the client need to compare every STR in the received range from the server with the auditors or just the latest STR?

Since the client only stores the latest STR from the server, it would do a hash chain verification on the STR range from the auditor and just compare the latest STRs with each other.

protocol/auditlog.go
c633: I meant the client also requests using the hash, the address would be used to establish the connection only. Hence, there should have no hashing for every lookup. Does it make sense?

protocol/auditlog.go
c633: > In an earlier iteration we had two separate requests, [...]

I remember this, too. But the main reason I suggested this because we have #173 now. 

> Is the concern here that the client could receive an STR from the auditor than the STR it received from the server during the last monitoring?

Yes.

> The only way this could happen is if the client finishes with monitoring very close to/at the end of an epoch. [...]

I think it could also happen if the `EpochDeadline` is very small, say, 5s, and the client has a high network latency, even if it does monitoring at the beginning of en epoch. But I agree that we shouldn't make things more complicated. What you said makes sense to me.

> Since the client only stores the latest STR from the server, it would do a hash chain verification on the STR range from the auditor and just compare the latest STRs with each other.

Do you mean the client fetches a range of STR from the server, verifies them, and compare the newest STR in the range with the auditors?

protocol/auditlog.go
masomel: Got it. This seems fine to me.

protocol/auditlog.go
masomel: > I think it could also happen if the EpochDeadline is very small, say, 5s, and the client has a high network latency, even if it does monitoring at the beginning of en epoch. But I agree that we shouldn't make things more complicated. What you said makes sense to me.

This is a really good point, though. So, maybe what makes sense is to change the `AuditingRequest` message to have a `startEpoch` and `endEpoch` instead, and if the client wants one STR, it sets both params to the same epoch? And if it needs to catch up, it can also indicate this with this message format.

> Do you mean the client fetches a range of STR from the server, verifies them, and compare the newest STR in the range with the auditors?

Right, although this isn't the full picture. The full flow is: (1) the client fetches an STR range from the server. (2) the client does the hash chain check, and if it passes the check, the client accepts and caches the latest STR from the server. (3) the client requests the same STR range from an auditor. (4) the client verifies that range as in step 2. (5) if the client independently accepts the latest STRs from both, it then compares only the latest STR. If there has been a fork at any point in time, the STRs *will* differ because somewhere along the way, the `prevHash` will be different. So there's no need for the client to compare every STR in the range. It's also important to note that this flow will only happen if the client misses epochs, so it should be a less common operation.

c633: > Right, although this isn't the full picture. The full flow is: [snip]

Since the latest STR commits to the entire history, so I think (3) could be: the client requests the latest STR from an auditor. Does this compromise the security? 
But then the client doesn't know at which epoch the server has equivocated (even in the full flow).

masomel: > (3) could be: the client requests the latest STR from an auditor. Does this compromise the security?

It does since the client won't detect if the server has tried to "unfork" a forked history, i.e. the client won't check if the history is linear, for any given observer. The comparison of the two latest STRs then still helps the client make sure that the server hasn't equivocated about the latest STR.

> But then the client doesn't know at which epoch the server has equivocated (even in the full flow).

We thought it would be more important (and enough) to detect *that* a server had equivocated, and less when this happened. On the other hand, having a specific point in time for an equivocation could help servers recover from an equivocation (for example, in the case of a single bad employee). This would be a task for a separate issue, I think.

c633: Cool! Thank you!

> This is a really good point, though. So, maybe what makes sense is to change the AuditingRequest message to have a startEpoch and endEpoch instead, and if the client wants one STR, it sets both params to the same epoch? And if it needs to catch up, it can also indicate this with this message format.

Yes, I agree.

protocol/auditlog.go
masomel: > can we use the hash of the key server's public signing key?

I've been thinking about this some more, and on second thought, I'm not sure we should index the STRs in an audit log by the hash of a server's signing key. If we do, we would break the continuity of a server's history when the server changes its signing key. This doesn't seem ideal since it might allow a server to equivocate using its two different signing keys. Can you please explain the purpose of this change?

arlolra: Maybe the index should be the hash for the initial STR?

c633: I agree.

masomel: That's a much more elegant solution.

protocol/auditlog.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	h := l.histories[addr]
+
+	if err := verifySTRConsistency(h.signKey, h.latestSTR, newSTR); err != nil {
+		return err
+	}
+
+	// update the latest STR
+	h.snapshots[h.latestSTR.Epoch] = h.latestSTR
```
arlolra: Shouldn't this be,

```
h.snapshots[newSTR.Epoch] = newSTR
```

or swap these two lines.  You've already set the snapshot for the previous `latestSTR` in L102-103 above.

masomel: `Insert` is only called once per directory, when the auditor first observes a directory's history. On subsequent epochs, the auditor only calls `Update()` so it needs to update the `latestSTR` on every epoch. But maybe it's misleading to separate the `latestEpoch` from the rest of the history in `Insert()`. Might make sense to only pass a list of STRs when initializing an observed history.

arlolra: The question I'm asking is when should the newSTR be added to the map of snapshots.  At Update, or at the next call to Update.  Because right now it's only on subsequent calls.  Except for at Insert, when it gets added right away.

masomel: Ah. The main reason for doing this was to make it a bit faster (and the code a bit cleaner) to access the `latestSTR` whenever a client does a lookup. An alternative approach would be to keep track of the current epoch and fetch the latestSTR from the map whenever a lookup happens.

masomel: On second thought, the use of `latestSTR` in the `directoryHistory` data structure may just be cause confusion. Happy to change the implementation.

arlolra: Ok, then remove L102-103 as redundant.

protocol/auditlog.go
```diff
@@ -0,0 +1,187 @@
...
+	snapshots map[uint64]*DirSTR
+	latestSTR *DirSTR
+}
+
+// A ConiksAuditLog maintains the histories
+// of all CONIKS directories known to a CONIKS auditor.
+// Each history includes the directory's public signing key
+// enabling the auditor to verify the corresponding signed
+// tree roots.
+type ConiksAuditLog struct {
```
arlolra: Does this need to be struct?  Or, can it just be,

```
type ConiksAuditLog map[string]*directoryHistory
```

protocol/auditlog.go
masomel: That'll be gone anyway once I remove the `latestSTR` field from the `directoryHistory` struct and we start using `latestEpoch`.

protocol/auditlog.go
```diff
@@ -0,0 +1,187 @@
...
+
+	// make sure the request is well-formed
+	if len(req.DirectoryAddr) <= 0 || req.StartEpoch > req.EndEpoch {
+		return NewErrorResponse(ErrMalformedClientMessage),
+			ErrMalformedClientMessage
+	}
+
+	h := l.histories[req.DirectoryAddr]
+
+	if h == nil {
```
arlolra: You seem to use IsKnownDirectory for this above.

protocol/auditlog.go
```diff
@@ -0,0 +1,187 @@
...
+		return NewErrorResponse(ReqUnknownDirectory), ReqUnknownDirectory
+	}
+
+	// also make sure the epoch range is well-formed
+	if req.StartEpoch > h.latestSTR.Epoch || req.EndEpoch > h.latestSTR.Epoch {
```
arlolra: Only need to check req.EndEpoch here.

protocol/auditlog.go
```diff
@@ -0,0 +1,187 @@
...
+		return NewErrorResponse(ErrMalformedClientMessage),
+			ErrMalformedClientMessage
+	}
+
+	var strs []*DirSTR
+	for ep := req.StartEpoch; ep < req.EndEpoch; ep++ {
+		str := h.snapshots[ep]
```
arlolra: This is problematic if latestSTR isn't in the snapshots.  See the discussion above.

protocol/auditlog.go
```diff
@@ -0,0 +1,187 @@
...
+		strs = append(strs, str)
+	}
+
+	// don't forget to append the STR for EndEpoch
+	strs = append(strs, h.snapshots[req.EndEpoch])
```
arlolra: Why wouldn't you just do `<=` in the loop above?

protocol/auditlog.go
c633: Can we merge this check with the above one (L160)?

protocol/auditlog.go
```diff
@@ -0,0 +1,187 @@
...
+	h := newDirectoryHistory(signKey, latestSTR)
+
+	startEp := uint64(0)
+	endEp := latestSTR.Epoch
+
+	// add each old STR into the history
+	// This loop automatically catches if oldSTRs is malformed
+	// (i.e. oldSTRs starts at epoch 0 and
+	// len(oldSTRs) != latestSTR.Epoch-1)
+	for ep := startEp; ep < endEp; ep++ {
```
c633: I'd like to have a `TODO: verify the prior STR history when it first joins` here.

protocol/auditlog.go
masomel: This check depends on `h`, so I wanted to fail early on L160 and not even retrieve `h` from the log if the request is malformed.

protocol/auditlog.go
masomel: There sort of already is one on L107. But that FIXME is a bit buried at the bottom of the function.

protocol/auditlog.go
c633: I'd like to keep `latestSTR` here for 1), it is a bit faster to access the `latestSTR` which I assume should be a frequent operation; and 2), we have this in `auditorState` as in #170 (used for generic auditing functionality).
And, I think the newSTR should be added to the map at `Update()`.

masomel: Yes, this was my original reasoning. I think the code around `latestSTR` can still be cleaned up a bit. But `latestSTR` *is* in the log, just not in the internal map of snapshots until the subsequent call to `Updated()`

protocol/auditlog.go
c633: OK, I feel that it would make the reader understand easier why we only need to check req.EndEpoch here. But either is fine.

protocol/auditlog.go
c633: I think the main purpose of keeping `latestSTR` is the performance, and for the consistency, I suggest we should update the `latestSTR` whenever we insert it into the map, and vice versa, so it may not be forgotten. Please also see https://github.com/coniks-sys/coniks-go/pull/155/files/f63dbb0b265021712fbdcbc78b8d9f84aee6b52f#r124956103.

protocol/auditlog_test.go
```diff
@@ -0,0 +1,247 @@
...
+		t.Fatal("Unable to get latest observed STR")
+	}
+
+	obs := res.DirectoryResponse.(*STRHistoryRange)
+	if len(obs.STR) != 1 {
+		t.Fatal("Expect returned STR to be not nil")
+	}
+	if obs.STR[0].Epoch != d.LatestSTR().Epoch {
+		t.Fatal("Unexpected epoch for returned STR")
+	}
```
c633: Please add a testcase for https://github.com/coniks-sys/coniks-go/pull/155/files/f63dbb0b265021712fbdcbc78b8d9f84aee6b52f#r124956103.

protocol/auditlog.go
masomel: That's a good point. I will make this change

protocol/auditlog.go
masomel: > update the latestSTR whenever we insert it into the map, and vice versa

The `latestSTR` is always updated on an `Insert()` (implicitly in `newDirectoryHistory()` and explicitly in `Update()`, what do you mean by "vice versa"?

c633: I meant we would have a function 
```
func updateSTR() {
h.snapshots[newSTR.Epoch] = newSTR
h.latestSTR = newSTR
}
```
and it would be called everywhere the `latestSTR` is updated. I mean I'd like to see these statements go together.

masomel: Ah, got it. Makes sense!

masomel:
@c633 @arlolra Addressed your comments, PTAL

protocol/auditlog.go
```diff
@@ -0,0 +1,205 @@
...
+
+// A ConiksAuditLog maintains the histories
+// of all CONIKS directories known to a CONIKS auditor,
+// indexing the histories by the hash of a directory's initial
+// STR.
+// Each history includes the directory's domain name as a string, its
+// public signing key enabling the auditor to verify the corresponding
+// signed tree roots, and a list with all observed snapshots in
+// chronological order.
+type ConiksAuditLog map[[crypto.HashSizeByte]byte]*directoryHistory
```
c633: Nitpicking: Can we use `map[string]*directoryHistory` with `str := hex.EncodeToString(crypto.Digest(initSTR.Signature))`?

protocol/auditlog.go
```diff
@@ -0,0 +1,205 @@
...
+
+	startEp := uint64(0)
+	endEp := uint64(len(hist))
+
+	// add each STR into the history
+	// This loop automatically catches if hist is malformed
+	// (i.e. hist is missing and epoch)
+	// FIXME: verify the consistency of each new STR before inserting
+	// into the audit log
+	for ep := startEp; ep < endEp; ep++ {
```
c633: My thought here is we will verify the consistency of the hash chain, so it maybe better if the index starts from `startEp+1` (since the `hist[startEp]` is the initial STR), thus should we insert the initial STR into the map snapshot in `newDirectoryHistory`? It can be:
```
+func newDirectoryHistory(dirName string, signKey sign.PublicKey, str *DirSTR) *directoryHistory {
+	h := new(directoryHistory)
+	h.dirName = dirName
+	h.signKey = signKey
+	h.snapshots = make(map[uint64]*DirSTR)
+      h.snapshots[computeInitSTRHash(str)] = str
+	h.latestSTR = str
+	return h
+} 
```

protocol/auditlog_test.go
```diff
@@ -0,0 +1,200 @@
...
+		t.Fatal("Unable to get latest observed STR")
+	}
+
+	obs := res.DirectoryResponse.(*STRHistoryRange)
+	if len(obs.STR) == 0 {
+		t.Fatal("Expect returned STR to be not nil")
+	}
+	if obs.STR[0].Epoch != d.LatestSTR().Epoch {
+		t.Fatal("Unexpected epoch for returned latest STR")
+	}
```
c633: I'd like to have some calls `d.Update()` here and re-request again.

c633:
Please rebase with master and make travis happy.

protocol/auditlog.go
masomel: It took me a long time to get this to work :P ...but I was assuming that we are hashing the entire STR. Hashing the signature is much simpler.

protocol/auditlog.go
masomel: I think that the initial STR should be in `snapshots` since it's also pinned in the client, and the client verifies its directory's history with this pinned STR. So the auditor should mirror the entire history for each directory that it tracks. But I do like your idea that `newDirectoryHistory` should take the initial STR as a parameter.

protocol/auditlog_test.go
masomel: So, just repeating this test multiple times?

protocol/auditlog.go
c633: Yes, since we are doing the same for `str.PreviousSTRHash`.

protocol/auditlog.go
c633: Sure, what I was suggesting is we should insert the initial STR at `newDirectoryHistory` and start verifying the str consistency with `startEp+1`.

protocol/auditlog_test.go
c633: Oops, sorry. I mean we call `d.Update` several times, then call `aud.Update`, then test the auditing request again, to address the problem Arlo has mentioned before.

protocol/auditlog.go
masomel: I got it. It makes total sense!

protocol/auditlog_test.go
masomel: I think what you're looking for is in `TestGetEntireSTRHist()`. Since the latestSTR is always inserted into the `snapshots` on each `aud.Update()`, the problem Arlo mentioned should be taken care of now.

masomel: I think I understand what you mean now, you want there to be a call to `GetObservedSTR()` *after* the audit log has been updated with a new STR 👍

c633: Yes, yes!

protocol/auditlog.go
```diff
@@ -0,0 +1,214 @@
...
+// Each history includes the directory's domain name as a string, its
+// public signing key enabling the auditor to verify the corresponding
+// signed tree roots, and a map with the snapshots for each observed
+// epoch.
+type ConiksAuditLog map[string]*directoryHistory
+
+// computeInitSTRHash is a wrapper for the digest function;
+// returns empty string if the STR isn't an initial STR (i.e. str.Epoch != 0)
+func computeInitSTRHash(initSTR *DirSTR) string {
+	if initSTR.Epoch != 0 {
```
c633: I think we should either remove this check or panic here.

c633: I refactored this function in https://github.com/coniks-sys/coniks-go/pull/155/commits/92cee27f21ab8299483b4d43712cfbd05c8308a6 (it would be used by the clients, too). 
Feel free to revert if you don't like it.

masomel: This refactoring LGTM.

protocol/auditlog.go
```diff
@@ -0,0 +1,202 @@
...
+// a new history for a known directory, an ErrMalformedDirectoryMessage
+// if oldSTRs is malformed, a CheckBadSignature or CheckBadSTR if there
+// is an inconsistency in the history given in hist, and nil otherwise.
+// Insert() only creates the initial entry in the log for addr. Use Update()
+// to insert newly observed STRs for addr in subsequent epochs.
+// FIXME: pass Response message as param
+// masomel: will probably want to write a more generic function
+// for "catching up" on a history in case an auditor misses epochs
+func (l ConiksAuditLog) Insert(addr string, signKey sign.PublicKey,
+	hist map[uint64]*DirSTR) error {
```
arlolra: `hist` is confusing here ... maybe `snaps` instead?

protocol/auditlog.go
```diff
@@ -0,0 +1,202 @@
...
+
+	// make sure we're getting an initial STR at the very least
+	if len(hist) < 1 && hist[0].Epoch != 0 {
```
arlolra: `hist` is a map so it's possible that the length is greater than 1 but `hist[0]` returns `nil`.  Since you're using it in several places below, cache it as `initSTR, ok := hist[0]` and then check `ok` here.

protocol/common.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+// ComputeDirectoryIdentity returns the hash of
+// the directory's initial STR as a string.
+// It panics if the STR isn't an initial STR (i.e. str.Epoch != 0).
+func ComputeDirectoryIdentity(str *DirSTR) string {
+	if str.Epoch != 0 {
+		panic(fmt.Sprintf("[coniks] Expect epoch 0, got %x", str.Epoch))
+	}
+	return hex.EncodeToString(crypto.Digest(str.Signature))
```
arlolra: Why the hex encoding here?  We do `bytes.Equal` with the `str.PreviousSTRHash` in other places.  Maybe it'd be better to keep all these `crypto.Digest(Signature)` operations the same.

protocol/auditlog.go
```diff
@@ -0,0 +1,202 @@
...
+			return err
+		}
+
+		h.snapshots[ep] = str
+	}
+
+	// Make sure to update the latestSTR
+	// in this particular call, the latestSTR has already been
+	// inserted into the snapshots map in the loop above
+	h.updateLatestSTR(hist[endEp-1])
```
arlolra: I think it would be cleared if you moved this to the last step of the loop, so that rather than

```
h.snapshots[ep] = str
```

you have,

```
h.updateLatestSTR(str)
```

then you won't need the clarifying comment, and you won't be doing it redundantly when `endEp == 1`.  It does mean a few unnecessary pointer updates, but I think it's worth it.

protocol/auditlog.go
```diff
@@ -0,0 +1,202 @@
...
+	l[dirInitHash] = h
```
arlolra: Maybe don't expose the internals of `l` and add a method, `l.Add(dirInitHash, h)` or something.

protocol/auditlog.go
```diff
@@ -0,0 +1,202 @@
...
+// exists in the audit log l.
+// FIXME: pass Response message as param
+func (l ConiksAuditLog) Update(dirInitHash string, newSTR *DirSTR) error {
+
+	// error if we want to update the entry for an addr we don't know
+	if !l.IsKnownDirectory(dirInitHash) {
+		return ErrAuditLog
+	}
+
+	h := l[dirInitHash]
```
arlolra: Same thing here, `l.Get(dirInitHash)` or similar.

protocol/auditlog.go
```diff
@@ -0,0 +1,202 @@
...
+	// (i.e. hist is missing an epoch between 0 and the latest given)
+	for ep := startEp; ep < endEp; ep++ {
+		str := hist[ep]
+		if str == nil {
+			return ErrMalformedDirectoryMessage
+		}
+
+		// verify the consistency of each new STR before inserting
+		// into the audit log
+		err := verifySTRConsistency(signKey, h.snapshots[ep-1], str)
```
arlolra: See the comment below, but this would then become `(signKey, h.lastestSTR, str)`

protocol/auditlog.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	h := l[dirInitHash]
+
+	if err := verifySTRConsistency(h.signKey, h.latestSTR, newSTR); err != nil {
+		return err
+	}
+
+	// update the latest STR
+	h.updateLatestSTR(newSTR)
```
arlolra: With the refactor suggested above, these two operations (`verifySTRConsistency`, `updateLatestSTR`) can be combined into an `h.internalUpdate(newSTR)`.

protocol/auditlog.go
```diff
@@ -0,0 +1,202 @@
...
+// message.NewErrorResponse(ReqUnknownDirectory) tuple.
+func (l ConiksAuditLog) GetObservedSTRs(req *AuditingRequest) (*Response,
+	ErrorCode) {
+
+	// make sure we have a history for the requested directory in the log
+	if !l.IsKnownDirectory(req.DirInitSTRHash) {
+		return NewErrorResponse(ReqUnknownDirectory), ReqUnknownDirectory
+	}
+
+	h := l[req.DirInitSTRHash]
```
arlolra: Or, maybe we should embrace that it's a map and always just use `h, ok :=` and drop the `IsKnownDirectory` method.

protocol/auditlog.go
c633: Good catch! Can we change it to a slice?

protocol/auditlog.go
```diff
@@ -0,0 +1,202 @@
...
+	return l
+}
+
+// IsKnownDirectory checks to see if an entry for the directory
+// (indexed by the hash of its initial STR dirInitHash) exists
+// in the audit log l. IsKnownDirectory() does not
+// validate the entries themselves. It returns true if an entry exists,
+// and false otherwise.
+func (l ConiksAuditLog) IsKnownDirectory(dirInitHash string) bool {
+	h := l[dirInitHash]
```
masomel: Still getting used to some Go idioms...

protocol/auditlog.go
```diff
@@ -0,0 +1,202 @@
...
+// allowing CONIKS clients to audit the CONIKS directories.
+
+package protocol
+
+import (
+	"github.com/coniks-sys/coniks-go/crypto/sign"
+)
+
+type directoryHistory struct {
+	name      string
```
masomel: This used to be `dirName` for directory domain name (i.e. its address), which is why we assign it `addr` below. My plan right now is to use this field when auditors query the directories they log every epoch.

protocol/auditlog.go
masomel: A slice would also make sense. Then we would have to check that the epoch for the STR in each element corresponds to the index, which seems fine to me.

protocol/auditlog.go
masomel: I was considering making this change just today when I was working on #170. I totally agree with you that this step would be clearer if we just call `updateLatestSTR()` even if it means extra pointer updates.

protocol/auditlog.go
masomel: Because of what I'm doing for #170, I need to see if it makes sense to refactor these operations both on the client and auditor. I'll probably hold off on that for now.

protocol/common.go
masomel: Using a string simplifies indexing for the `AuditLog`. Before this commit, the index needed was represented as a 32-byte array, which worked fine but ended up being a bit messy.

protocol/auditlog.go
masomel: We should, see https://github.com/coniks-sys/coniks-go/pull/155/files#r125792869

protocol/auditlog.go
```diff
@@ -0,0 +1,196 @@
...
+// (i.e. the hash of the initial STR) dirInitHash in the ConiksAuditLog.
+func (l ConiksAuditLog) Set(dirInitHash string, dirHistory *directoryHistory) {
+	l[dirInitHash] = dirHistory
+}
+
+// Get retrieves the directory history for the given directory identifier
+// dirInitHash from the ConiksAuditLog.
+// Get() also returns a boolean indicating whether the requested dirInitHash
+// is present in the log.
+func (l ConiksAuditLog) Get(dirInitHash string) (*directoryHistory, bool) {
```
c633: Can we unexport this function? Same with `Set`.

protocol/auditlog.go
```diff
@@ -0,0 +1,196 @@
...
+// Insert() only creates the initial entry in the log for addr. Use Update()
+// to insert newly observed STRs for addr in subsequent epochs.
+// FIXME: pass Response message as param
+// masomel: will probably want to write a more generic function
+// for "catching up" on a history in case an auditor misses epochs
+func (l ConiksAuditLog) Insert(addr string, signKey sign.PublicKey,
+	snaps []*DirSTR) error {
+
+	// make sure we're getting an initial STR at the very least
+	if len(snaps) < 1 && snaps[0].Epoch != 0 {
```
c633: Should it be `len(snaps) < 1 || snaps[0].Epoch != 0`?

protocol/auditlog.go
```diff
@@ -0,0 +1,196 @@
...
+	// create the new directory history
+	h = newDirectoryHistory(addr, signKey, snaps[0])
+
+	// add each STR into the history
+	// start at 1 since we've inserted the initial STR above
+	// This loop automatically catches if snaps is malformed
+	// (i.e. snaps is missing an epoch between 0 and the latest given)
+	for i := 1; i < len(snaps); i++ {
+		str := snaps[i]
+		if str == nil || str.Epoch != uint64(i) {
```
c633: The `verifySTRConsistency` would make sure that `str.Epoch == i`, so we can remove the check here.

protocol/auditlog.go
```diff
@@ -0,0 +1,196 @@
...
+			return ErrMalformedDirectoryMessage
+		}
+
+		// verify the consistency of each new STR before inserting
+		// into the audit log
+		err := verifySTRConsistency(signKey, h.latestSTR, str)
```
c633:
```
if err := verifySTRConsistency(signKey, h.latestSTR, str); err != nil {
  return err
}
```

protocol/auditlog.go
masomel: Oops yes, this should be an OR. Thank you for catching that.

protocol/auditlog.go
masomel: This check is only needed because `snaps` is a slice, and it's being checked in the loop. This property isn't inherent in the consistency check. Plus, `verifySTRConsistency` already does this check implicitly for one STR when it calls `VerifyHashchain`.

c633: I do think that this property is inherent in the consistency check, since we are verify using the initial STR, and it would fail if `str.Epoch != latestSTR.Epoch + 1`, which means `str.Epoch == i` has to be true.

masomel: What I meant is that `i == str.Epoch` is only meaningful because we expect the `snaps` slice to have this property. `i` isn't meaningful to consistency checks in the general sense, so I don't see why we should impose this particular assumption on the consistency checks. And `VerifyHashchain` already does the `str.Epoch != latestSTR.Epoch + 1` check.

PR #157, opened 2016-12-19
Use protocol.DirSTR for policy unmarshalling

### LINKED ISSUE #135: Use protocol.DirSTR for policy unmarshalling
Started with commit https://github.com/coniks-sys/coniks-go/pull/93/commits/3c6603393499ebf56ce063ba6e4839e2e3753df9. Also see https://github.com/coniks-sys/coniks-go/blob/94d368df307139f4f9681adbe72e32adfb269523/merkletree/str.go#L35

arlolra:
Closes #135


protocol/str.go
```diff
@@ -0,0 +1,31 @@
...
+		*str,
+		str.Ad.(*Policies),
+	}
+}
+
+// VerifyHashChain computes the hash of savedSTR's signature,
+// and compares it to the hash of previous STR included
+// in the issued STR. The hash chain is valid if
+// these two hash values are equal and consecutive.
+func (str *DirSTR) VerifyHashChain(savedSTR *DirSTR) bool {
```
c633: Could we make this a `merkletree.SignedTreeRoot` method? It seems like we will need a more elegant way to handle the `savedSTR` parameter.

c633: I made some changes in 959c496e7061297d1f3690eb5957e5a97d68a698. Hopefully it works for you.

protocol/str.go
```diff
@@ -0,0 +1,19 @@
+package protocol
+
+import "github.com/coniks-sys/coniks-go/merkletree"
+
+type DirSTR struct {
+	merkletree.SignedTreeRoot
```
c633: Could we use a pointer here instead?

protocol/str.go
arlolra: I'm not sure I agree with this change.

c633: My point is I want to make `merkletree` be an independent and complete package and I think `merklettree.SignedTreeRoot` also needs a verifying method. Please feel free to discard this change :)

protocol/str.go
arlolra: Maybe

protocol/str.go
c633: > we should restore the original merkletree/str_test.go I removed and keep the new protocol/str_test.go as well.

I totally agree.

arlolra: Restored in a2f119b2

protocol/str.go
c633: This won't be changed?

c633:
LGTM! Thank you, Arlo!

protocol/str.go
arlolra: Is there a compelling reason to?

c633: I think we can save memory by having one pointer to, say, the latest STR for all requests in the same epoch, instead of allocating new one for each request. Is that reasonable?

arlolra: 1c8418de seems to work.

But, I'm not sure I follow what you're saying.  Are `DirSTR`s retained?  They'll just be collected, no?  It seems like that just saves some copies.

c633: Yes, they will be gc'd. But I think saving some copies can be huge save when the server needs to handle multiple requests? I'm so nitpicking :(

arlolra:
Merged in 8d42521d

PR #158, opened 2016-12-19
Resolve #154

### LINKED ISSUE #154: Force the bot to delete older messages
If the bot account is compromised, the attackers wouldn't be able to see a list of all the accounts that had registered in the past and when.
This is an improvement for our privacy-preserving effort.

c633:


bots/twitterbot.go
```diff
@@ -100,6 +100,8 @@ func NewTwitterBot(path string) (Bot, error) {
 	bot.coniksAddress = conf.CONIKSAddress
 	bot.handle = conf.Handle
 
+	bot.deleteOldDMs()
```
masomel: Is `deleteOldDMs` a safeguard in case `deleteRequestDMs` doesn't manage to delete all DMs? I'm wondering how often this function will actually be necessary given that `deleteRequestDMs` already deletes response DMs.

c633: Yeah, it is. As you said, just a safeguard!

c633:
Any thoughts on this before this gets merged, @arlolra @Liamsi?

bots/twitterbot.go
```diff
@@ -197,10 +202,54 @@ func (bot *TwitterBot) HandleRegistration(username string, msg []byte) string {
...
-	_, _, err := bot.client.DirectMessages.New(params)
-	return err
+	dm, _, err := bot.client.DirectMessages.New(params)
+	return dm, err
+}
+
+// deleteOldDMs deletes all prior DMs before the bot runs.
+func (bot *TwitterBot) deleteOldDMs() {
+	log.Println("[registration bot] Deleting old DMs ...")
+	params := &twitter.DirectMessageGetParams{Count: 200} // up to a maximum of 200.
```
Liamsi: Why delete the last 200 messages (and not all for instance)? As you discussed with @masomel above it seems possible that `deleteRequestDMs` isn't able to delete all messages. But then, is 200 a strict upper bound (if so why)? Also, I'm not sure we would want to delete those missed messages on startup.

c633: 200 is the upper bound limited by Twitter API. 
`deleteRequestDMs` deletes the message using its ID (which is returned by the API). But the dev can terminate the bot before it deletes all the messages (There is a timer before calling `deleteRequestDMs`). That's why I think we need `deleteOldDMs` to clean up remaining messages.

c633: Please see this for Twitter API limit: https://dev.twitter.com/rest/reference/get/direct_messages

Liamsi: I see. Makes sense! Maybe add a comment about the 200 mentioning the API limit. (maybe even linking to: https://dev.twitter.com/rest/reference/get/direct_messages if you think that makes sense)

c633:
Merged in e499520246f993d646929295a30ec4f8ff830c49.

PR #159, opened 2016-12-21
Print timestamp along with the result for debugging

c633:
- Also add a "q" command for exiting the program.

### DIFF
--- a/client/coniksclient/internal/cmd/common.go
+++ b/client/coniksclient/internal/cmd/common.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/coniks-sys/coniks-go/client"
 	"github.com/spf13/cobra"
@@ -39,5 +40,5 @@ func loadConfigOrExit(cmd *cobra.Command) *client.Config {
 
 // append "\r\n" to msg and then write to terminal in raw mode.
 func writeLineInRawMode(term *terminal.Terminal, msg string) {
-	term.Write([]byte(msg + "\r\n"))
+	term.Write([]byte("<" + time.Now().Format("15:04:05.999999999") + "> " + msg + "\r\n"))
 }
--- a/client/coniksclient/internal/cmd/common.go
+++ b/client/coniksclient/internal/cmd/common.go
@@ -39,6 +39,9 @@ func loadConfigOrExit(cmd *cobra.Command) *client.Config {
 }
 
 // append "\r\n" to msg and then write to terminal in raw mode.
-func writeLineInRawMode(term *terminal.Terminal, msg string) {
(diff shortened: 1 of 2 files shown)


client/coniksclient/internal/cmd/common.go
```diff
@@ -39,5 +40,5 @@ func loadConfigOrExit(cmd *cobra.Command) *client.Config {
 
 // append "\r\n" to msg and then write to terminal in raw mode.
 func writeLineInRawMode(term *terminal.Terminal, msg string) {
-	term.Write([]byte(msg + "\r\n"))
+	term.Write([]byte("<" + time.Now().Format("15:04:05.999999999") + "> " + msg + "\r\n"))
```
Liamsi: Example output:
`<15:43:01.090076899> [+] Are you trying to update your binding? Unfortunately, KeyChange isn't supported yet.`
One question to the others: Should we make this optional? 
Otherwise I'm fine with this pull!

c633: How about having commands `enable/disable timestamp`?

Liamsi: I would prefer to have it in the config. But if its easier/more convenient a command like you propose is fine, too.

Liamsi: True, it's not interesting for anything else than the test-client. OK, you convinced me. Let's see if there are other opinions.

Liamsi:
That's a "hidden" command (not shown in the help text) and it's disabled by default. Still LGTM!

c633:
I added in 159/commits/ca00620784fdfa5a0fa6ea7c41b3d886517c79c7. Thanks!

client/coniksclient/internal/cmd/common.go
masomel: This seems like a debugging feature in the test client, maybe we could add a `-d/--debug` flag to the `run` command?

client/coniksclient/internal/cmd/run.go
```diff
@@ -17,6 +17,10 @@ const help = "- register [name] [key]:\r\n" +
 	"	Register a new name-to-key binding on the CONIKS-server.\r\n" +
 	"- lookup [name]:\r\n" +
 	"	Lookup the key of some known contact or your own bindings.\r\n" +
+	"- enable timestamp:\r\n" +
+	"	Print timestamp of format <15:04:05.999999999> along with the result.\r\n" +
+	"- disable timestamp:\r\n" +
+	"	Disable timestamp printing.\r\n" +
```
masomel: I still think a debug flag would be more elegant, but at the same time, I also kinda like the option to toggle on/off the timestamp at runtime.

c633:
Merged in 6827a2177cc4faa2edd5b85279a6a0707580dfe9.

PR #160, opened 2016-12-23
Hotfix: STR.Ad unmarshalling

c633:


arlolra:
Can you explain what the problem is here?

The client should only be performing verification on the DirSTR, no?

c633:
Sure.

When the client receives the next STR, it calls `verifySTRConsistency` which calls `str.Serialize()`, which is the underlying method of `merkletree.SignedTreeRoot`. But due to JSON unmarshalling, `str.Ad` would be nil, then `str.Ad.Serialize()` call (https://github.com/coniks-sys/coniks-go/blob/master/merkletree/str.go#L64) would panic. This patch fixes this bug by re-assigning `str.Policies` to `str.Ad` after unmarshalling the response.

arlolra:
Hmm, so this goes back to our discussion here, https://github.com/coniks-sys/coniks-go/pull/157#discussion_r92979261

c633:
Actually, it panics in https://github.com/coniks-sys/coniks-go/blob/master/protocol/consistencychecks.go#L148, before calling `VerifyHashChain`. I think the problem is due to unmarshalling.

arlolra:
Right, I edited my comment to that effect.

client/encoding_test.go
```diff
@@ -32,3 +33,12 @@ func TestUnmarshalMalformedErrorResponse(t *testing.T) {
 			"got", res.Error)
 	}
 }
+
+func TestUnmarshalSampleMessage(t *testing.T) {
+	msg := `{"Error":100,"DirectoryResponse":{"AP":{"TreeNonce":"dmriJjZCIs451/6e6OWdEhukqf35szGK0MDP5tqStoU=","PrunedTree":[[248,166,186,56,249,219,214,188,80,58,222,92,254,175,220,201,255,212,82,86,184,177,28,122,242,237,252,205,37,56,178,164]],"LookupIndex":"HzdAcxzkVDEdCnrc0Rq6KZ7zewuqgy6BZrBBXkbPiL4=","VrfProof":"ZIbhTZFn0l6K8NISjBpOo+L4GyrvRBxIN2QlHpchewxSlK+ok0NNzvQuD7qKDXMUbj/aPMBXerGwdur1obNiAqPsCgoO1y7rxxHXdXJ3BL08W8RBTb2XlfyempBWMWVA","Leaf":{"Level":1,"Index":"HzdAcxzkVDEdCnrc0Rq6KZ7zewuqgy6BZrBBXkbPiL4=","Value":"a2V5ZWtlcg==","IsEmpty":false,"Commitment":{"Salt":"NIKu0xbTVEsBREihmzxiLqWKmv6m0pu5mapCuHw2YBA=","Value":"b0uh+7ssMwvp7xcN+lm97tWsRABJr1cjS+dAUNty5Us="}}},"STR":{"TreeHash":"QgJG9j1yq3z50PhddRKkA34V+8ucEqWWjzDnyv6y960=","Epoch":3,"PreviousEpoch":2,"PreviousSTRHash":"6txKUQyf+TCLYkrwS0qEaN4LI6Pn2rV7oavWgUaXEeY=","Signature":"PfJvZuT4MPSBj1lJ2iU6qlD81r6VMh4KtH83a5HqGsQd9b4bi7CFKTzDjlxy0TgK+HE+lv+4shZ222NgZJdiCg==","Policies":{"Version":"0.3","HashID":"SHAKE128","VrfPublicKey":"Csmr5OOqALXvRq/rND+A6pYaEQ0tA7x0sSMPML+N8vQ=","EpochDeadline":60}}}}`
```
arlolra: This string is likely to go stale as we update the code.  It would be preferable to generate the response.

arlolra:
Hmm, maybe DirSTR needs to override the Serialize method instead?  Just thinking out loud.

c633:
I did that change, but it turned out that https://github.com/coniks-sys/coniks-go/blob/master/protocol/consistencychecks.go#L136 would return fail because of the difference between `str.Ad` and `savedSTR.Policies`. I'm not really happy with this patch, but because the problem happens when unmarshalling, I think it would be the best place to patch.

arlolra:
The reflect.DeepEqual issue seems only relevant to testing, because it practice all the SavedSTR will be coming from unmarshalling.

client/coniksclient/internal/cmd/run.go
```diff
@@ -190,7 +190,7 @@ func keyLookup(cc *p.ConsistencyChecks, conf *client.Config, name string) string
 		return ("Invalid config!")
 	}
 
-	response := client.UnmarshalResponse(p.RegistrationType, res)
+	response := client.UnmarshalResponse(p.KeyLookupType, res)
```
arlolra: This seems unrelated.  Can you push it directly?

c633: Done.

c633:
Another possible patch is in 4b80d07cdc07f59a68c6b9ecddcbfe104b15e97b.

arlolra:
I pushed 6fab2ff as an alternative since it seems a little safer.

arlolra:
> Another possible patch is in 4b80d07, as Arlo suggested.

I find the replication there unacceptable.

c633:
I'm so sorry for digging into it again, I pushed some changes in 380ba5c87c4e4aaafbbf4c6cc225d7d5e0366e8f. What do you think?

arlolra:
I think I prefer things as they are ... not sure.

c633:
Since we have always been trying to avoid mixing marshalling with the core library :confused:...

arlolra:
Good point ... it's the `WithAd` that I'm finding distasteful.   Maybe if you switched it to `Serialize` and `SerializeWithoutAd` or, `SerializeInternal` or something.

c633:
Fixed in a9b6ccab6a77139da22f32efc32cb05bc259d3aa. Thanks for your suggestion!

arlolra:
Yeah, that seems ok.

PR #164, opened 2017-01-19
[WIP] Make benchmarks more comparable to those of the orig paper

### LINKED ISSUE #163: Benchmark re-computation of the tree exactly like in the orig. paper (Figure 7)
As discussed with @masomel:
- Create directory/pad with X users; including computation of the tree (our `pad.Update` operation)
- Insert 1000 additional entries
- benchmark the following: only re-computation of the tree (`pad.Update`) on tree with (X+1000)  entries/users

Liamsi:
see section 5.3 of the CONIKS paper
resolves #163

### DIFF
--- a/protocol/directory.go
+++ b/protocol/directory.go
@@ -126,7 +126,7 @@ func (d *ConiksDirectory) NewTB(name string, key []byte) *TemporaryBinding {
 // snapshot.
 // In any case, str is the signed tree root for the latest epoch.
 // If Register() encounters an internal error at any point, it returns
-// a message.NewErrorRespose(ErrDirectory) tuple.
+// a message.NewErrorResponse(ErrDirectory) tuple.
 func (d *ConiksDirectory) Register(req *RegistrationRequest) (
 	*Response, ErrorCode) {
 	// make sure the request is well-formed
--- a/merkletree/pad_test.go
+++ b/merkletree/pad_test.go
@@ -345,10 +345,15 @@ func benchPADUpdate(b *testing.B, entries uint64) {
 			b.Fatal(err)
 		}
 	}
+	// clone current PAD's state:
+	orgTree := pad.tree.Clone()
 	b.ResetTimer()
 
 	// now benchmark re-hashing the tree:
 	for i := 0; i < b.N; i++ {
+		b.StopTimer()
+		pad.tree = orgTree.Clone()
+		b.StartTimer()
 		pad.Update(nil)
 	}
 }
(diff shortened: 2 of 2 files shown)


Liamsi:
Note to myself: This PR still isn't really equivalent to what the java implementation did: We add 1000 users once but the inner "benchmark-loop" considers the newly added users in the first iteration only. 
Investigate if internally resetting 1000 hashes to `nil` works instead ([it should](https://github.com/coniks-sys/coniks-go/blob/0bedfc6e83b710377aee19011a90f48e57c1584d/merkletree/node.go#L78-L96)).

protocol/directory.go
```diff
@@ -126,7 +126,7 @@ func (d *ConiksDirectory) NewTB(name string, key []byte) *TemporaryBinding {
 // snapshot.
 // In any case, str is the signed tree root for the latest epoch.
 // If Register() encounters an internal error at any point, it returns
-// a message.NewErrorRespose(ErrDirectory) tuple.
+// a message.NewErrorResponse(ErrDirectory) tuple.
```
c633: Fixed this in `master` branch in a2dbdeaca62a9f46e588b4ca2219fd53a8066559.

c633:
> We add 1000 users once but the inner "benchmark-loop" considers the newly added users in the first iteration only.

I tried with this:
```
func benchPADUpdate(b *testing.B, entries uint64) {
	[...]
	// build the tree once:
	pad.Update(nil)

	// Insert 1000 additional entries (as described in section 5.3):
	var i uint64
	for i = 0; i < 1000; i++ {
		key := keyPrefix + string(i+entries)
		value := append(valuePrefix, byte(i+entries))
		if err := pad.Set(key, value); err != nil {
			b.Fatal(err)
		}
	}
	// clone current PAD's state:
	orgTree := pad.tree.Clone()
	b.ResetTimer()

	// now benchmark re-hashing the tree:
	for i := 0; i < b.N; i++ {
		b.StopTimer()
		pad.tree = orgTree.Clone()
		b.StartTimer()
		pad.Update(nil)
	}
}
```

There is the result of both versions:
```
# before
$ go test -bench BenchmarkPADUpdate100K
BenchmarkPADUpdate100K-4   	      10	 199899731 ns/op
PASS
ok  	github.com/coniks-sys/coniks-go/merkletree	215.545s

# after
$ go test -bench BenchmarkPADUpdate100K
BenchmarkPADUpdate100K-4   	      30	  67303366 ns/op
PASS
ok  	github.com/coniks-sys/coniks-go/merkletree	313.979s
```

What do you think @Liamsi ?

masomel:
@c633 I like your solution!

Liamsi:
I agree! Elegant and simple solution.

PR #169, opened 2017-03-07
Init tree with an empty node as root node

c633:


merkletree/merkletree_test.go
```diff
@@ -15,74 +13,137 @@ var vrfPrivKey1, _ = vrf.GenerateKey(bytes.NewReader(
...
+// 			"expected", expect,
+// 			"get", root.leftHash)
+// 	}
+
+// 	r = m.Get([]byte("abc"))
+// 	if r.Leaf.Value != nil {
+// 		t.Error("Invalid look-up operation:", key)
+// 		return
+// 	}
+// }
```
Liamsi: Will these test be reactivated? In other words is this pull WIP or ready for review?

c633: We still need to decide if we go this way. I added this to our meeting agenda.

Liamsi: Thanks. Right, we started an email tread about this pull. Sorry, I was not sure anymore.

Issue #87, opened 2016-09-19
How should the client reporting proof validation errors (aka blow the whistle)?


masomel:
The client needs to provide a mechanism for the user to blow the whistle on the key server if any proof verification fails. Some suggested mechanisms: Post to Twitter (if account linked), post to public web page, disseminate via auditors.

PR #170, opened 2017-03-12
Refactor common STR verification functionality

masomel:
Since the client and auditor both need to perform some common consistency checks on the STRs, all this functionality should be refactored into a generic auditor module. This will also require the implementation of STR verification over a range of epochs.

See https://github.com/coniks-sys/coniks-go/pull/155#discussion_r104835245


Liamsi:
We will merge #155 first

protocol/auditlog.go
```diff
@@ -29,19 +42,23 @@ type ConiksAuditLog map[[crypto.HashSizeByte]byte]*directoryHistory
 
 // updateLatestSTR inserts a new STR into a directory history;
```
c633: s/updateLatestSTR/updateVerifiedSTR

protocol/auditlog.go
```diff
@@ -178,7 +195,7 @@ func (l ConiksAuditLog) GetObservedSTRs(req *AuditingRequest) (*Response,
 	}
 
 	// make sure the request is well-formed
-	if req.EndEpoch > h.latestSTR.Epoch || req.StartEpoch > req.EndEpoch {
+	if req.EndEpoch > h.verifiedSTR.Epoch || req.StartEpoch > req.EndEpoch {
```
c633: Please replace `h.verifiedSTR` with `h.VerifiedSTR()`

c633:
Generally, I think this can be:
- have a generic interface for different auditor types (i.e. `Auditor` interface)
- have a default auditor (`AudState`) which implements `Auditor` interface, and unexport functions such as `CompareWithVerified` or `VerifySTRConsistency`, etc.
- make `AudState` a nested struct of `AuditLog` and `ConsistencyChecks`, and extend the interface implementations as needed.

masomel:
> have a default auditor (AudState) which implements Auditor interface, and unexport functions such as `CompareWithVerified` or `VerifySTRConsistency`, etc.

I don't fully understand what we gain by doing this. Can you please elaborate on this idea?

c633:
I think with #173, client- and auditor-side would share (very much) the same logic code of `Auditor` interface implementation (e.g., `Audit` function). All they do are:
- fetch the latest (or a range of) STR(s) from the server
- verify the consistency & update the state.

By doing that, we can let each side extend the interface as needed, and reuse the verifying logic from `AudState`, much like what we did with `SignedTreeRoot` and `DirSTR.Serialize()`: https://github.com/coniks-sys/coniks-go/blob/master/protocol/str.go#L21-L23.

masomel:
> with #173, client- and auditor-side would share (very much) the same logic code of `Auditor` interface implementation (e.g., `Audit` function) All they do are:
>
> * fetch the latest (or a range of) STR(s) from the server
> * verify the consistency & update the state.

Is this because after #173 the clients will start sending the separate STRs requests to the servers as auditors do now? I agree that the fetching of STRs from client-->auditor and auditor-->server is very similar, but the actual `Audit`ing isn't that similar:
* Clients need to: (1) Verify the STR history from the server and save the latest STR, and (2) Verify the STR history from the auditor and compare the latest against the verified STR.
* Auditors need to: Verify the STR history from the server and save all STRs.

There is definitely functionality that can be refactored here, and I think I've done that so far, but I don't think we can efficiently use the same `Audit` function for both. But I do agree that we can definitely refactor the default functionality some more.

c633:
I agreed with what you said. But I think some functions like `CompareWithVerified` and `VerifySTRConsistency` are used only for *verify the STR history from the server*, and that is what I mean "share the same logic code". The other parts could be done by extending the interface implementation in separate functions.

For this pull, I think you can refactor the STR verification functionality, and use it in the `AuditLog`, since most of the code of `ConsistencyChecks` would be rewritten towards #173 (I'm doing it now).

masomel:
I understand `ConsistencyChecks` will change as part of #173,  but I'll still update it for now to demonstrate how the refactored logic could be used. You can then change it however you see fit for #173.

protocol/auditlog.go
```diff
@@ -40,11 +38,12 @@ func newDirectoryHistory(addr string, signKey sign.PublicKey, initSTR *DirSTR) *
...
-// assumes the STR has been validated by the caller
-func (h *directoryHistory) updateVerifiedSTR(newVerified *DirSTR) {
-	h.snapshots[newVerified.Epoch] = newVerified
-	h.verifiedSTR = newVerified
+// updateVerifiedSTR inserts a new range of STRs into a directory history;
+// assumes the STRs have been validated by the caller
+func (h *directoryHistory) updateVerifiedSTR(newVerified []*DirSTR) {
+	h.Update(LatestSTRInRange(newVerified))
+	h.snapshots = append(h.snapshots, newVerified...)
+
```
c633: nitpicking: please remove this empty line.

protocol/auditor.go
```diff
@@ -12,17 +12,19 @@ import (
 	"github.com/coniks-sys/coniks-go/crypto/sign"
 )
 
+// Auditor provides a generic interface allowing different
+// auditor types to implement specific auditing functionality.
+type Auditor interface {
+	AuditDirectory([]*DirSTR) error
```
c633: 👍 I really like this name.

protocol/auditlog.go
masomel: Oops, must have forgotten to remove it when I was done debugging.

protocol/auditlog.go
```diff
@@ -11,10 +11,21 @@ import (
...
 	addr      string
-	signKey   sign.PublicKey
-	snapshots map[uint64]*DirSTR
-	latestSTR *DirSTR
+	snapshots []*DirSTR
+}
+
+// caller validates that initSTR is for epoch 0
+func newDirectoryHistory(addr string, signKey sign.PublicKey, initSTR *DirSTR) *directoryHistory {
+	a := NewAuditor(signKey, initSTR)
```
c633: Please use either `a` or `aud` for the consistency. You used `aud` in https://github.com/coniks-sys/coniks-go/pull/170/files?diff=unified#diff-8eead7b3ac1a29e7ad9e067fb3d9aaf8R46

protocol/auditlog.go
```diff
@@ -11,10 +11,21 @@ import (
...
+	h := &directoryHistory{
```
c633: I would like to make this field:value initializers, i.e. 
```
h := &directoryHistory{
		AudState:  a,
		addr:      addr,
		snapshots: make([]*DirSTR, 0),
	}
```

protocol/auditor.go
```diff
@@ -0,0 +1,174 @@
...
+
+// AudState verifies the hash chain of a specific directory.
+type AudState struct {
+	signKey     sign.PublicKey
+	verifiedSTR *DirSTR
+}
+
+var _ Auditor = (*AudState)(nil)
+
+// New instantiates a new auditor state from a persistance storage.
```
c633: s/New/NewAuditor

protocol/auditlog.go
```diff
@@ -100,6 +116,7 @@ func (l ConiksAuditLog) Insert(addr string, signKey sign.PublicKey,
 	// create the new directory history
 	h = newDirectoryHistory(addr, signKey, snaps[0])
 
+	// FIXME: remove this check --> caller calls Audit() before this function
```
c633: Since we will pass Response message as param, should we assume that the callers will call Audit() before this function? Or is it better to embed the Audit() call in this function?

protocol/auditlog.go
```diff
@@ -100,6 +116,7 @@ func (l ConiksAuditLog) Insert(addr string, signKey sign.PublicKey,
 	// create the new directory history
 	h = newDirectoryHistory(addr, signKey, snaps[0])
```
c633: Please verify snaps[0] with the `pinnedSTR` before inserting.

protocol/auditlog.go
```diff
@@ -112,26 +129,28 @@ func (l ConiksAuditLog) Insert(addr string, signKey sign.PublicKey,
...
-		if err := verifySTRConsistency(signKey, h.latestSTR, str); err != nil {
+		if err := h.verifySTRConsistency(snaps[i-1], str); err != nil {
 			return err
 		}
-
-		h.updateLatestSTR(snaps[i])
 	}
 
 	// Finally, add the new history to the log
+	if len(snaps) > 1 {
```
c633: Please add a test to coverage this path.

c633: Should we move this check into `updateVerifiedSTR`, and just call `h.updateVerifiedSTR(snaps)`?

protocol/auditlog.go
```diff
@@ -141,12 +160,14 @@ func (l ConiksAuditLog) Update(dirInitHash [crypto.HashSizeByte]byte, newSTR *Di
 		return ErrAuditLog
 	}
 
-	if err := verifySTRConsistency(h.signKey, h.latestSTR, newSTR); err != nil {
+	// FIXME: remove this check --> caller calls Audit() before this function
```
c633: Same as above.

protocol/auditor.go
```diff
@@ -0,0 +1,174 @@
...
+// CompareWithVerified checks whether the received STR is the same as
+// the verified STR in the AudState using reflect.DeepEqual().
+func (a *AudState) compareWithVerified(str *DirSTR) error {
+	if reflect.DeepEqual(a.verifiedSTR, str) {
+		return nil
+	}
+	return CheckBadSTR
+}
+
+// CheckSTRAgainstVerified checks an STR str against the a.verifiedSTR.
```
c633: s/CheckSTRAgainstVerified/checkSTRAgainstVerified

protocol/auditor.go
```diff
@@ -0,0 +1,174 @@
...
+// the two STRs.
+// CheckSTRAgainstVerified() returns nil if the check passes,
+// or the appropriate consistency check error if any of the checks fail,
+// or str's epoch is anything other than the same or one ahead of
+// a.verifiedSTR.
+func (a *AudState) checkSTRAgainstVerified(str *DirSTR) error {
+	// FIXME: check whether the STR was issued on time and whatnot.
+	// Maybe it has something to do w/ #81 and client transitioning between epochs.
+	// Try to verify w/ what's been saved
+
```
c633: I think we should add a FIXME: 
```
	// FIXME: we are returning the error immediately
	// without saving the inconsistent STR
	// see: https://github.com/coniks-sys/coniks-go/pull/74#commitcomment-19804686
```
here or in `AuditDirectory`.

protocol/auditor.go
```diff
@@ -0,0 +1,174 @@
...
+// CheckSTRAgainstVerified checks an STR str against the a.verifiedSTR.
+// If str's Epoch is the same as the verified, CheckSTRAgainstVerified()
+// compares the two STRs directly. If str is one epoch ahead of the
+// a.verifiedSTR, CheckSTRAgainstVerified() checks the consistency between
+// the two STRs.
+// CheckSTRAgainstVerified() returns nil if the check passes,
+// or the appropriate consistency check error if any of the checks fail,
+// or str's epoch is anything other than the same or one ahead of
+// a.verifiedSTR.
+func (a *AudState) checkSTRAgainstVerified(str *DirSTR) error {
```
c633: Please reorder these functions:
```
compareWithVerified()
verifySTRConsistency()
checkSTRAgainstVerified()
verifySTRRange()
```

protocol/auditor.go
```diff
@@ -0,0 +1,174 @@
...
+	// TODO: verify the directory's policies as well. See #115
+	return CheckBadSTR
+}
+
+// VerifySTRHistory checks the consistency of a range
+// of a directory's STRs. It begins by verifying the STR consistency between
+// the given prevSTR and the first STR in the given range, and
+// then verifies the consistency between each subsequent STR pair.
+func (a *AudState) verifySTRRange(prevSTR *DirSTR, strs []*DirSTR) error {
+
```
c633: Nitpicking: can we remove this empty line? Here and some other places.

protocol/auditor.go
```diff
@@ -0,0 +1,174 @@
...
+	prev := prevSTR
+	for i := 0; i < len(strs); i++ {
+		str := strs[i]
+		if str == nil {
+			// FIXME: if this comes from the auditor, this
+			// should really be an ErrMalformedAuditorMessage
+			return ErrMalformedDirectoryMessage
```
c633: I also encountered this problem, and I think we should just change this error and `ErrMalformedClientMessage` to `ErrMalformedResquest` and `ErrMalformedResponse`.

c633: I can open a separate pull later, since I already have this.

protocol/consistencychecks.go
```diff
@@ -45,18 +43,48 @@ func NewCC(savedSTR *DirSTR, useTBs bool, signKey sign.PublicKey) *ConsistencyCh
 	if !useTBs {
 		panic("[coniks] Currently the server is forced to use TBs")
 	}
+	aud := NewAuditor(signKey, savedSTR)
 	cc := &ConsistencyChecks{
```
c633: It is easier for reader when using field:value initializers.

protocol/consistencychecks.go
```diff
@@ -45,18 +43,48 @@ func NewCC(savedSTR *DirSTR, useTBs bool, signKey sign.PublicKey) *ConsistencyCh
...
 
+// CheckEquivocation checks for possible equivocation between
+// an auditors' observed STRs and the client's own view.
+// CheckEquivocation() first verifies the STR range received
+// in msg if msg contains more than 1 STR, and
+// then checks the most recent STR in msg against
+// the cc.verifiedSTR.
+// CheckEquivocation() is called when a client receives a response to a
+// message.AuditingRequest from an auditor.
+func (cc *ConsistencyChecks) CheckEquivocation(msg *Response) error {
```
c633: IMHO, we should open a separate pull for this feature, since I think we should cover something like the number of chosen auditors, and design a set of APIs for the `/protocol` level.

c633:
If it's possible, please remove the empty line after the function name. 
Please update the documentation in the comments (some function names are uppercase, or not matched with the new name).

protocol/auditlog.go
masomel: Good point. I think it makes more sense to separate the logic, so we should keep the function signature as-is and only pass the Response to the `Audit()` function.

protocol/auditlog.go
masomel: `Audit()` is going to be called before `Insert()`, so `snaps[0]` will have been verified at this point.

protocol/auditlog.go
masomel: This check only applies to `Insert`s because `snaps[0]` is inserted into `h.snapshots` in `newDirectoryHistory` and the rest of the range is added separately, while in all other cases (i.e. Update) we will always add the entire range at once. So moving this check into `updateVerifiedSTR` would have to distinguish between an`Insert` and an `Update`, which I think makes things messy.

protocol/auditor.go
masomel: Why is this a FIXME? Why should we be saving inconsistent STRs?

protocol/auditor.go
masomel: Sounds good to me.

protocol/consistencychecks.go
masomel: I agree. This isn't currently consistent in all of the code, so we should fix this everywhere.

protocol/consistencychecks.go
masomel: The number of chosen auditors is independent of the core equivocation check protocol -- that's what this is. What other APIs do we need for equivocation checking on the client?

protocol/auditlog.go
masomel: Also, this path is already covered in tests that include https://github.com/coniks-sys/coniks-go/blob/generic-auditor/protocol/auditlog_test.go#L29 via https://github.com/coniks-sys/coniks-go/blob/generic-auditor/protocol/testutil.go#L51

protocol/auditlog.go
```diff
@@ -11,10 +11,21 @@ import (
 )
 
 type directoryHistory struct {
+	*AudState
 	addr      string
-	signKey   sign.PublicKey
-	snapshots map[uint64]*DirSTR
-	latestSTR *DirSTR
+	snapshots []*DirSTR
```
c633: Can we keep using the map as before? Is there any advantages over using a map? 
I'm afraid that we may accidentally insert 2 same DirSTR into snapshots if we use slice.

protocol/consistencychecks.go
c633: How about taking an array of auditors' `Response` and passing them to `CheckEquivocation`? I have no strong opinion about this, though.

protocol/auditlog.go
c633: OK, I got it. 
> snaps[0] is inserted into h.snapshots in newDirectoryHistory and the rest of the range is added separately

If we switch to use a map of snapshots as before, this could be avoided.

protocol/auditor.go
c633: This makes me think we should do:
> Any STR that passes the checks in updateSTR will be stored and used by the client, regardless of whether the rest of the checks pass / fail, and that's fine since it contains proof of being issued.

I think we should have a TODO/FIXME here so we can come back later, when we address client recovery & whistle-blowing issues.

masomel: I generally agree with this, but I think auditors should save inconsistent STRs in a separate list. I also agree that this is a whistleblowing and recovery issue. We should open a new issue for this if we don't have one already.

c633: We have #87 and #171. What I was saying is we add a TODO here so we don't forget it.

protocol/consistencychecks.go
masomel: I thought about doing this, although that I think that already imposes a requirement on the client to check multiple auditors. I think that detail should be implemented at the application layer to make the use of auditors more flexible in the client.

protocol/auditlog.go
masomel: This is a possibility, but the slice makes it much easier to append a whole batch of STRs. I wanted to avoid having to loop through a range twice, once for the verification, and one for the insertion into `snapshots`. But maybe this wouldn't be a huge performance issue?

c633: Looking at current code, there are insertion and search operations, which are O(1), so I think it wouldn't be a huge performance issue.

protocol/consistencychecks.go
```diff
@@ -45,18 +43,48 @@ func NewCC(savedSTR *DirSTR, useTBs bool, signKey sign.PublicKey) *ConsistencyCh
...
+	if len(strs.STR) > 1 {
+		if err := cc.verifySTRRange(strs.STR[0], strs.STR[1:]); err != nil {
+			return err
+		}
+	}
+
+	// TODO: if the auditor has returned a more recent STR,
+	// should the client update its savedSTR? Should this
+	// force a new round of monitoring?
+	return cc.checkSTRAgainstVerified(LatestSTRInRange(strs.STR))
```
c633: Last-minute nitpicking: should we remove `LatestSTRInRange` and just do
```
latestSTRInRange := strs[len(strs)-1]
return cc.checkSTRAgainstVerified(latestSTRInRange)
```
?
(or even pass `strs[len(strs)-1]` directly to checkSTRAgainstVerified)

protocol/auditlog.go
c633: If we keep the function signature as-is, the caller has to de-serialize the Response twice (in `Audit` and before calling `Insert`), then wouldn't it make things more complicated? I mean the implementor has to do
```
Audit(response)
deSerialize(response)
Insert(...)
```
You also said that `Audit` will "finally update the snapshots if the checks pass", which means you cannot call `Audit` before this function (actually we cannot call `Audit` before this function since we call `newDirectoryHistory` inside this function). 
I think this correct logic is 
```
Insert() {
  h := newDirectoryHistory()
  h.Audit()
  h.updateVerifiedSTR() //optional if we don't do this in Audit
  log.set(h)
}
```

protocol/auditlog.go
c633: Please see my comment: https://github.com/coniks-sys/coniks-go/pull/170#discussion_r132661384. I think we still need to verify `snaps[0]` when creating a new directory history. We could include this verification in `newDirectoryHistory`. What do you think?

protocol/auditor.go
```diff
@@ -0,0 +1,176 @@
...
+	return nil
+}
+
+// AuditDirectory validates a range of STRs received from a CONIKS directory.
+// AuditDirectory() checks the consistency of the oldest STR in the range
+// against the verifiedSTR, and verifies the remaining
+// range if the message contains more than one STR.
+// AuditDirectory() returns the appropriate consistency check error
+// if any of the checks fail, or nil if the checks pass.
+func (a *AudState) AuditDirectory(strs []*DirSTR) error {
```
c633: @masomel @arlolra @Liamsi Should we check null parameters in the function or assume that the caller will validate the parameters before calling?

protocol/auditor.go
```diff
@@ -0,0 +1,176 @@
...
+	"fmt"
+	"reflect"
+
+	"github.com/coniks-sys/coniks-go/crypto"
+	"github.com/coniks-sys/coniks-go/crypto/sign"
+)
+
+// Auditor provides a generic interface allowing different
+// auditor types to implement specific auditing functionality.
+type Auditor interface {
```
c633: Maybe we should add a TODO "this interface is used nowhere, we should use it everywhere in ConsistencyChecks and AuditLog instead of AudState." ?

masomel: I agree that `ConsistencyChecks` and `AuditLog` should "extend" `Auditor` rather than `AudState`, but to me this problem is more about how we can express what we mean. 

Ideally, what I would have liked to do here is declare a generic auditor type/interface, as one would in Java for example, and have the `AuditLog` and `ConsistencyChecks` extend that class and implement something like an abstract `AuditDirectory` method. I think what we have now comes closest to this. Maybe the solution would be to rename `AudState` to `Auditor` and rename `Auditor` to something like `AbstractAuditor`?

masomel: Also, thinking ahead to implementing a CoSi-compatible auditor, AFAICT, it should be able to directly use `AudState` (or whatever we call it) and implement `AuditDirectory()` like we've done so far.

protocol/consistencychecks.go
masomel: `LatestSTRInRange` is only used here, right? I added this when I using a slice for the `snapshots` instead of a map, but it's no longer necessary now. I'll remove it.

protocol/auditor.go
c633: I agree, though I have no idea what it should be :(. Thus, I suggest we add a TODO here or open an issue so we can address it later.

masomel: Tbh, the way it's implemented right now was according to a suggestion you made to make `AudState` implement `Auditor` and have the other two use `AudState` so that other kinds of auditors could implement something different. I'm a bit confused why you're suggesting now that we should change this again ;)

I need to think about it some more, but I think that the only new type of auditor which would have to implement the `Auditor` interface would be a blockchain based auditor. So I don't think we need to change what we have when it already expresses what we intended. Maybe @arlolra and @Liamsi can weigh in on this, too?

protocol/auditor.go
masomel: Hmm, I generally assume that the caller hasn't validated the parameters. In the case of `AuditDirectory`, both `ConsistencyChecks` and `AuditLog`, both will call it inside of `Audit`, so in that case I think it would be safe to assume that the caller did a validation beforehand. OTOH, adding a parameter validation here is trivial, so I think we can add it here to be safe.

protocol/auditlog.go
masomel: I think there are multiple ways to approach this. `Insert` shouldn't do anything but do an actual insertion into the audit log. I don't understand why `newDirectoryHistory` *must* be called before `Audit`, those are independent operations. The current logic is: "Did the audit pass? Yes, so let's insert the snapshots into our log/update our log". This totally fits with my previous statement, what it's saying is that insertion/updates to the log should happen *after* auditing checks pass. This *is* the logic that's currently implemented.

Now, I do agree that passing the response message to `Insert` would require the implementor to de-serialize twice, and that's not elegant. My intention was to keep the function signatures consistent with the ones in `directory.go`, where `Insert` also takes the `Response`. but I never felt like this really made sense in the auditor's case.

So, given my previous statement about how `Audit()` should update the snapshots if the checks pass, `Audit()` should really be calling `Insert()` and `Update`, not the other way around. In fact, at the application level, it would really make more sense to have the auditor simply call `Audit()` on a new server response and have the insertion/update be abstracted away from the application layer. Then, we can unexport `Insert` and `Update`, and pass the snapshots directly instead of the `Response`. What do you think?

protocol/auditlog.go
masomel: Similar to my comment in https://github.com/coniks-sys/coniks-go/pull/170#discussion_r132728501, I don't think that it's `newDirectoryHistory`'s job to verify `snaps[0]`, especially if it's being called in the context of `Audit()`.

protocol/auditlog.go
masomel: So, what I'm suggesting is

```
Audit(Reponse r) {
  pass = AuditDirectory()
  if pass {
    if new directory
      insert(snaps)
    else
      update(snaps)
  }
}
```
I haven't implemented this yet, because it's part of the server-auditor protocol.

c633: > I don't understand why newDirectoryHistory must be called before Audit, those are independent operations.

It's "Go" things, you are declaring `Audit` as a method of `directoryHistory`, you cannot call Audit without instantiating a variable of `directoryHistory`.

> The current logic is: "Did the audit pass? Yes, so let's insert the snapshots into our log/update our log".

Yes, that's true. I totally agree, but could we say "Do you want to insert into/update our log? Let's check the audit first."

> In fact, at the application level, it would really make more sense to have the auditor simply call Audit() on a new server response and have the insertion/update be abstracted away from the application layer. Then, we can unexport Insert and Update, and pass the snapshots directly instead of the Response

Yes, we are trying to achieve a same thing 🎉 . How can you separate insertion/update if they are abstracted?

c633: Anyways, your suggestion makes a lot of senses. Sorry for wasting your time 😕

masomel: > It's "Go" things, you are declaring Audit as a method of directoryHistory, you cannot call Audit without instantiating a variable of directoryHistory.

Oh gotcha! Good point, sorry for being slow :(

masomel: Implementing this as part of #182

protocol/auditlog.go
masomel: Also addressing this as part of #182

protocol/auditor.go
```diff
@@ -0,0 +1,174 @@
...
+// AudState verifies the hash chain of a specific directory.
+type AudState struct {
+	signKey     sign.PublicKey
+	verifiedSTR *DirSTR
+}
+
+var _ Auditor = (*AudState)(nil)
+
+// NewAuditor instantiates a new auditor state from a persistance storage.
+func NewAuditor(signKey sign.PublicKey, verified *DirSTR) *AudState {
```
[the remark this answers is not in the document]
c633: We discussed about this at https://github.com/coniks-sys/coniks-go/pull/170#discussion_r132664234. For now, `AudState` has some methods which shouldn't belong to the `Auditor` interface.

protocol/auditlog.go
```diff
@@ -27,21 +38,25 @@ type directoryHistory struct {
 // chronological order.
 type ConiksAuditLog map[[crypto.HashSizeByte]byte]*directoryHistory
 
-// updateLatestSTR inserts a new STR into a directory history;
-// assumes the STR has been validated by the caller
-func (h *directoryHistory) updateLatestSTR(newLatest *DirSTR) {
-	h.snapshots[newLatest.Epoch] = newLatest
-	h.latestSTR = newLatest
+// updateVerifiedSTR inserts a new range of STRs into a directory history;
```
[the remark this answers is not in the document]
c633: done.

protocol/auditlog.go
```diff
@@ -27,21 +38,25 @@ type directoryHistory struct {
...
-// caller validates that initSTR is for epoch 0
-func newDirectoryHistory(addr string, signKey sign.PublicKey, initSTR *DirSTR) *directoryHistory {
-	h := new(directoryHistory)
-	h.addr = addr
-	h.signKey = signKey
-	h.snapshots = make(map[uint64]*DirSTR)
-	h.updateLatestSTR(initSTR)
-	return h
+// Audit checks that a directory's STR history
+// is linear and updates the audtor's state
```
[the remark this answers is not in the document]
c633: done.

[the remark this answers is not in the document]
masomel: Thanks @c633 for taking care of this!

masomel:
Am I right to assume that this is ready to be merged? No new comments in over a week.

Issue #165, opened 2017-01-25
HTTP requests instead of (or in addition to) TCP sockets


jpeisenbarth:
Hi,

I work at INRIA Nancy Grand Est (France) and our team is interested in the work you are leading on CONIKS.

We plan to use CONIKS in a context  of an web application of collaborative editing in a peer-to-peer network. In this context, the communication with the CONIKS server would be simplified if it listened to HTTP requests (HTTPS in fact, we could still use TLS). 

Currently the client and the server listen and use TCP sockets (or Unix
sockets) and I am opening this issue to discuss about the changing of the TCP sockets to HTTP requests (or maybe having both solutions).

What are you opinions about this ? (If you are ok I can develop this feature and suggest a PR :)

Thank you in advance !

masomel:
@jpeisenbarth Thank you for your question! Could you please elaborate on this:

> the communication with the CONIKS server would be simplified if it listened to HTTP requests

Can you also give us some details on how you would implement this feature?

jpeisenbarth:
@masomel At the beginning it could be really simple and just replace the TCP sockets with HTTP.

Concretely : 
Server side -> `http.ListenAndServeTLS` and `http.HandleFunc` instead of `net.ListenTCP` 
Client side -> `http.NewRequest` and `http.Client.Do` instead of `tls.Client`and `tls.Conn.Write`

I have a working prototype but is it a little dirty because with TCP you handle manually the sockets (listening, accepting and handling) but with HTTPS server it is automatic, only the handler function is needed. For this reason I bypass some code and it is not clean.

Secondly for a more elaborated HTTPS handling we could think about a REST API but this needs a lot of additional discussions

masomel:
I think your suggested changes make a lot of sense, communicating over HTTPS would definitely simplify the interface. We've briefly discussed using a REST API on our end as a viable alternative to our current communication interface, and as you said, implementing this would certainly require more discussion.

We'll make sure to discuss this more during our next developer meeting. On our end, we are currently prioritizing finishing missing core features. It seems like you've dedicated more time to this change, so we are more than happy to support you and discuss all the necessary design changes with you if you would like to lead the effort to implement communication over HTTPS. 

For now, when you think that your current prototype is ready for review, we'd be happy to review it and consider merging those changes! As a second step, we can work on developing a REST API later.

c633:
I think we need to do some refactor for the connection layer code. Related to #77. 
[...]

PR #174, opened 2017-04-24
HTTP communication

jpeisenbarth:
Hello,

the HTTP communication feature discussed in #165 has been added. 

For now we can still use TCP or Unix sockets, it depends on the config file (replace `tcp://127.0.0.1:3000` by `https://127.0.0.1:3000` for example)

I hope this is OK but do not hesitate to tell me if there is something wrong with the code

### DIFF
--- a/keyserver/server.go
+++ b/keyserver/server.go
@@ -1,10 +1,12 @@
 package keyserver
 
 import (
+	"context"
 	"crypto/tls"
 	"fmt"
 	"io/ioutil"
 	"net"
+	"net/http"
 	"net/url"
 	"os"
 	"os/signal"
@@ -93,6 +95,7 @@ type ConiksServer struct {
 	configFilePath string
 	reloadChan     chan os.Signal
 	epochTimer     *time.Timer
+	httpServer     *http.Server
 }
 
 // LoadServerConfig loads the ServerConfig for the server from the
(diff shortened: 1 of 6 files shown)


keyserver/server.go
```diff
@@ -281,9 +307,23 @@ func resolveAndListen(addr *Address) (ln net.Listener,
...
+	perms[protocol.MonitoringType] = true
+	perms[protocol.RegistrationType] = addr.AllowRegistration
+	return perms
+}
+
 // Shutdown closes all of the server's connections and shuts down the server.
 func (server *ConiksServer) Shutdown() error {
 	close(server.stop)
+	if server.httpServer != nil {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
c633: Can we move this scope to line 199 (https://github.com/coast-team/coniks-go/blob/92c82f41423983acc2e0d2e60fb0858527b5e6e2/keyserver/server.go#L199) as
```
go func() {
  <-server.stop
  // shutdown httpServer [...]
}()
```
and remove `httpServer` from `ConiksServer`?

jpeisenbarth: Yes, it is definitely a better solution !

keyserver/server.go
```diff
@@ -193,10 +192,15 @@ func (server *ConiksServer) Run(addrs []*Address) {
 				Handler:   mux,
 				TLSConfig: tlsConfig,
 			}
-			server.httpServer = httpSrv
 			go func() {
 				httpSrv.Serve(ln)
 			}()
+			go func() {
```
c633: Please add `server.waitStop.Add(1)` before this routine and `server.waitStop.Done()` after `httpSrv.Shutdown()`.

PR #167, opened 2017-02-05
[WIP] VXEdDSA

Liamsi:
resolves #128 
working version which still needs some cleanup:
- [ ] increase code readability (and more golang idiomatic)
- [ ] use cached public-key
- [ ] cleanup ed25519 package (unexport funcs, keep only one elligator2 implementation, test-vectors for elligator)
- [ ] adopt to google's keytransparency vrf interface (?)
- [ ] asserts
- [ ] TODOs, XXXs

PR #176, opened 2017-05-14
Vrf uniqueness was violated (quickfix)

### LINKED ISSUE #175: Uniqueness of VRF is violated
An attacker can choose a different rs when computing s. Rather than
s = H2(h, [r]G, [r]H), the attacker computes
s = H2(h, [a]G, [b]H) and claims that the VRF output is
[(b-a)/s + k] H rather than[k]H and t = a-sk.

This checks out because the verifier checks that
s == H2(h, [t]G + [s]\([k]G), [t]H + [s]VRF).
s == H2(h, [a]G, [b]H)
Because
[t]G+[s]\([k]G) = [a]G.
[t]H + [s]VRF = [a-sk]H + [b-a+sk]H = [b]H.
The attack succeeds at producing a non-unique value for VRF.

http://eprint.iacr.org/2012/577.pdf Page 4:

> Goldberg and Reyzin (March 2017) discovered that if one does not hash the unique identifier when computing the challenge of the proof system, the uniqueness of the VRF from DDH assumption is violated. Therefore, when using the VRF, it is important to hash the unique identifier as well.

Thanks to Sharon Goldberg and Leonid Reyzin for reporting this bug.

Liamsi:
Adopt VRF to update in CONIKS paper  …
- add unique identifier: also hash h=H1(m), g^x and h^x (the VRF output) and not only with random nonce r (g^r, h^r); additionally hash m into H2 (as before)
- like in the paper add the base-point and public-key
- skip outdated test-vectors
- renamed some vars according to naming in the paper
- unrelated: remove formatting version of test-output where no placeholder is used

fixes #175

### DIFF
--- a/crypto/internal/ed25519/edwards25519/const.go
+++ b/crypto/internal/ed25519/edwards25519/const.go
@@ -23,6 +23,20 @@ var A = FieldElement{
 	486662, 0, 0, 0, 0, 0, 0, 0, 0, 0,
 }
 
+var extendedBaseEl = ExtendedGroupElement{
+	FieldElement{25485296, 5318399, 8791791, -8299916, -14349720, 6939349, -3324311, -7717049, 7287234, -6577708},
+	FieldElement{-758052, -1832720, 13046421, -4857925, 6576754, 14371947, -13139572, 6845540, -2198883, -4003719},
+	FieldElement{-947565, 6097708, -469190, 10704810, -8556274, -15589498, -16424464, -16608899, 14028613, -5004649},
+	FieldElement{6966464, -2456167, 7033433, 6781840, 28785542, 12262365, -2659449, 13959020, -21013759, -5262166},
+}
+
+// BaseBytes can be used to hash the base point if necessary
+var BaseBytes [32]byte
+
+func init() {
+	extendedBaseEl.ToBytes(&BaseBytes)
+}
+
 var bi = [8]PreComputedGroupElement{
 	{
 		FieldElement{25967493, -14356035, 29566456, 3660896, -12694345, 4014787, 27544626, -11754271, -6079156, 2047605},
--- a/crypto/vrf/vrf.go
+++ b/crypto/vrf/vrf.go
@@ -120,8 +120,8 @@ func hashToCurve(m []byte) *edwards25519.ExtendedGroupElement {
 // same as returned by Compute(m).
 func (sk PrivateKey) Prove(m []byte) (vrf, proof []byte) {
 	x, skhr := sk.expandSecret()
-	var cH, rH [64]byte
-	var r, c, minusC, t, grB, hrB, iiB [32]byte
+	var sH, rH [64]byte
+	var r, s, minusS, t, grB, hrB, iiB [32]byte
 	var ii, gr, hr edwards25519.ExtendedGroupElement
 
 	hm := hashToCurve(m)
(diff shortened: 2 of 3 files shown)


maditya:
cc @rishiguro

r2ishiguro:
@Liamsi , I just wonder if you had considered this IETF draft? https://tools.ietf.org/html/draft-goldbe-vrf-00
The proposed hash is different than ours. Since we're sharing the code, maybe it's a good idea to keep the code in sync and the draft could be a good one to refer to...?

Liamsi:
Yes, we (the CONIKS team) are currently discussing to switch to this IETF draft (an alternative would be OWS's [vxeddsa](https://whispersystems.org/docs/specifications/xeddsa/#vxeddsa). I think it's a good idea to implement the IETF version soon. If we do so, sure, let's keep the code in sync!

Liamsi:
I would propose to implement two versions independently and share test-vectors. What do you think of this idea? I still need to discuss with my team-members, though.

r2ishiguro:
You mean, we implement the same algorithm (probably EC-VRF-ED25519-SHA256?) independently and compare the results? Sounds good to me, but after that I'd like to share one implementation. Maybe we can create a new repository just for VRF in GitHub, or we can refer to the coniks-sys or coname repository.

Liamsi:
> Maybe we can create a new repository just for VRF in GitHub, or we can refer to the coniks-sys or coname repository.

I think a single dedicated repo makes sense as the VRF might become useful for other projects, too. Did you also have a look on vxeddsa BTW?

r2ishiguro:
I looked it over. It looks the same except the way the hash value is constructed. The draft looks simpler and if it doesn't require SHA3, it might have a benefit a little bit.

r2ishiguro:
Here's a result from my implementation:

alpha: 6d657373616765
x: 1fcce948db9fc312902d49745249cfd287de1a764fd48afb3cd0bdd0a8d74674885f642c8390293eb74d08cf38d3333771e9e319cfd12a21429eeff2eddeebd2
P: 885f642c8390293eb74d08cf38d3333771e9e319cfd12a21429eeff2eddeebd2
pi: 033236f8991d15d073ff8754645f0a11071e3092227d85c243833d006b5cfe9dc0a89beb44a930f357073f49aaddc8f2d708a943c27b552a61f3594e39a59a8d49dc393cd22e48c0fbde5628c1c298924d
vrf: 3236f8991d15d073ff8754645f0a11071e3092227d85c243833d006b5cfe9d
h: 025120c3e33ae1f9829486fa11653d2dd33ce79b9d7321b912c487eac515d27261

I used golang.org/x/crypto/ed25519.GenerateKey() to generate (x, P).
h = ECVRF_hash_to_curve(alpha, g^x = P)  as specified in the draft. Sometimes I needed to iterate more than 100 to get a valid curve.. For Ed25519, there should be more efficient way to generate a curve, I guess, but for now I just followed the draft.

As for implementation, I used "golang.org/x/crypto/ed25519/internal/edwards25519" but it's internal so I couldn't import it. I copied it as-is into my repository for now.. Maybe we should ask Google to incorporate VRF into ed25519...?

andres-erbsen:
I tried, as far as I remember the pull request never got looked at.

Liamsi:
> I looked it over. It looks the same except the way the hash value is constructed. The draft looks simpler and if it doesn't require SHA3, it might have a benefit a little bit.

Vxeddsa doesn't use SHA3 either. Also, vxeddsa's hash_to_curve might be faster. We have a work in progress PR on that, too: https://github.com/coniks-sys/coniks-go/pull/167
Another thing to currently consider: the draft is still a draft and it might change.

masomel:
@r2ishiguro @andres-erbsen Would either of you be willing to review this PR? It LGTM, but it'd be great to have a more expert review before we merge this.

r2ishiguro:
LGTM, too. FYI, I'm exchanging emails with Sharon and Leonid about the test vectors. They seem to be "tweaking" the spec right now. I'll keep you informed.

masomel:
Thanks @r2ishiguro! And sounds good, hopefully we can get the ball rolling on that soon.

Issue #177, opened 2017-06-19
Create Test Vectors


gdbelvin:
A set of test vectors for 
- Tree hashing
- VRF
- Other crypto functions

Would be helpful for assessing cross-project and cross-language compatibility. 
I've begun doing this in google/keytransparency#613

c633:
Thanks @gdbelvin, I've also begun doing this in #178. 

Note that there is a difference between your tree hashing and ours: we use `treeNonce []byte` instead of `treeID uint64`.

gdbelvin:
Would you be interested in putting a maximum length on the treeNonce for compatibility?  
Would be awesome to have a common interface.

c633:
> Would you be interested in putting a maximum length on the treeNonce for compatibility?

Sure, it's is a 32-byte array.

c633:
@gdbelvin 

> Would you be interested in putting a maximum length on the treeNonce for compatibility?

In our implementation, it is a 32-byte array. Do you have any idea how should the interface be for compatibility?

/cc @masomel @arlolra @Liamsi

gdbelvin:
Hmm, I wonder if we could hash the TreeID along with extra context onto the
32 byte nonce?

gdbelvin:
After some more thought, I'd love to jump on a call to discuss this.  
An 8 byte nonce would give us 32 bits of security before two tree nonces accidentally collided with each other. 2^32 trees would indeed be a lot of CONIKS and Key Transparency servers 👍. Is there a reason for more than 32 bits of security in the tree nonce?

Liamsi:
As far as I remember this was discussed quite a while ago and we've decided to reuse [HashSizeByte](https://github.com/coniks-sys/coniks-go/blob/5e626ab2f2610256d328c1273b3cf26f68e762aa/crypto/util.go#L12) for simplicity (not introducing another constant) and some extra-bits of security. Another related reason was that we've decided to always hash the output of the system's randomness (and we use a 32-byte hash for that). But I agree that a 32-bit tree-nonce should be sufficient.

c633:
I think I don't want to let users (the developer/operator) supply the tree ID. As stated somewhere, maintaining the uniqueness of the counter (in this case the map/tree ID) is nontrivial.
As discussed at https://github.com/google/keytransparency/issues/65#issuecomment-137265950, I'd love to know the reason why KT still use the tree nonce.

gdbelvin:
[...]

PR #178, opened 2017-06-19
Create generic interface for hash functions

c633:
Create PADHasher & TreeHasher interface to allow developers to use their own hash function.

* Part of #6 and #177


gdbelvin:
Would there be any interest in trying to syncronize this interface with the one in google/trillian#670 ?

crypto/hasher/hasher_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+func TestHashLeafVectors(t *testing.T) {
+	for _, tc := range []struct {
+		treeNonce [32]byte // treeNonce is treeID in KT, it should be a 32-byte array for cross-project compatibility
+		index     []byte
+		depth     uint32
+		leaf      []byte
+		want      Hash
+	}{
+		{treeNonce: [32]byte{0}, index: []byte("foo"), depth: 128, leaf: []byte("leaf"), want: h2h("65e7f29787a6168affd016656bb1f4f03af91cf7416270f5015005f8594d3eb6")},
```
gdbelvin: Question about the index since I ran into this too. 
Should indexes be required to be the full length of the hash function? This would change the test vectors.

gdbelvin: See google/trillian#691 and google/keytransparency#614

Liamsi:
I like @gdbelvin's idea. It makes it easier to sync up and might increase compatibility. Be aware that the interface described in google/trillian#670  isn't fully implemented yet (but will be soon).

c633:
> Would there be any interest in trying to syncronize this interface with the one in google/trillian#670 ?

Definitely.

crypto/hasher/hasher_test.go
c633: For `HashLeaf`, I think yes. But the length of the index of empty nodes vary.

crypto/hasher/hasher.go
```diff
@@ -0,0 +1,94 @@
...
+// The passed slices won't be mutated.
+func (ch *coniksHasher) Digest(ms ...[]byte) []byte {
+	h := ch.New()
+	for _, m := range ms {
+		h.Write(m)
+	}
+	return h.Sum(nil)
+}
+
+// ID returns the name of the cryptographic hash function in string.
```
Liamsi: s/in string/as a string

crypto/hasher/hasher_test.go
gdbelvin: Just a note that in trillian we're using `int64` for treeID.

merkletree/pad.go
```diff
@@ -5,6 +5,7 @@ import (
 	"errors"
 
 	"github.com/coniks-sys/coniks-go/crypto"
+	conikshasher "github.com/coniks-sys/coniks-go/crypto/hasher/coniks"
```
Liamsi: I would rather prefer a shorter import alias. What do you think about `hasher`? As this whole project is called 'coniks', I feel `hasher` would be enough.

c633: Ideally, this package should be imported using a blank import name. I used `conikshasher` for now because we import both `hasher` and `hasher/coniks` sometimes.

Liamsi: Makes sense! I would still prefer a shorter import. It would saves some typing when using it and it seems more consistent with the generally shorter package names in golang. But it isn't that important.

c633: SGTM. Done.

crypto/hasher/hasher.go
```diff
@@ -0,0 +1,53 @@
...
+	ID() string
+	// Size returns the size of the hash output in bytes.
+	Size() int
+	// Digest hashes all passed byte slices. The passed slices won't be mutated.
+	Digest(ms ...[]byte) []byte
+	treeHasher
+}
+
+// treeHasher provides hash functions for tree implementations.
+type treeHasher interface {
```
Liamsi: On first sight, it's a bit confusing that the `treeHasher` is unexported but embedded in the exported `PADHasher` interface. Maybe there is a good reason which I don't see yet? OK, the 'tree-hashing logic' is separate from the rest (ID, Size, ...) but  if one wants to implement a different `PADHasher`, one needs to find the private interface `treeHasher`. I think it makes sense to collapse both and a public interface should be self-contained (or embed other public interfaces if there are really good reasons). LMK what you think.

crypto/hasher/hasher.go
```diff
@@ -0,0 +1,53 @@
...
+// RegisterHasher registers a hasher for use.
+func RegisterHasher(h string, f func() PADHasher) {
+	if _, ok := hashers[h]; ok {
+		panic(fmt.Sprintf("%s is already registered", h))
+	}
+	hashers[h] = f()
+}
+
+// Hasher returns a PADHasher.
+func Hasher(h string) (PADHasher, error) {
```
Liamsi: Maybe rename to `NewHasher` or `GetHasher`?

crypto/hasher/hasher.go
```diff
@@ -0,0 +1,53 @@
...
+
+// RegisterHasher registers a hasher for use.
+func RegisterHasher(h string, f func() PADHasher) {
+	if _, ok := hashers[h]; ok {
+		panic(fmt.Sprintf("%s is already registered", h))
+	}
+	hashers[h] = f()
+}
+
+// Hasher returns a PADHasher.
```
Liamsi: returns a registered PADHasher identified by the given string. If no such PADHasher exists, it returns an error.

crypto/hasher/coniks/coniks.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+func init() {
+	hasher.RegisterHasher(CONIKSHasher, New)
+}
+
+const (
+	// CONIKSHasher is the identity of the hashing algorithm
+	// specified in the CONIKSHasher paper.
+	CONIKSHasher = "CONIKS Hasher"
```
Liamsi: I wonder, if there will be different hashers, shouldn't be a central place of identifiers? I always thought we are only swapping the underlying hash function, but all these hashers will be "CONIKS Hashers".

c633: The idea is each hasher implementation should have its own ID declared in its own package. But you're right, we should have a central place of identifiers which includes our "standard" hash schemes, and "CONIKS Hasher" should be replaced by something like "SHA256-xxxx".

crypto/hasher/hasher.go
c633: I don't remember why I didn't publish `treeHasher` interface but I can't find any reasons for not doing that.

> I think it makes sense to collapse both and a public interface should be self-contained (or embed other public interfaces if there are really good reasons).

Can you elaborate?

Liamsi: I think this should be one interface instead of two. At least both should be public if there is a good reason for having two ifaces (which I don't see).

c633: Done.

gdbelvin:
Glad to see this PR moving forward!  Thanks @Liamsi and @c633.
I'll ask the Trillian team if we can use a `[32]byte` under the hood for consistency.  Perhaps we can get some test vectors for cross verification?

c633:
>  Perhaps we can get some test vectors for cross verification?

We would *love* to do that.

Issue #173, opened 2017-04-06
Keeping the client in sync


arlolra:
This proposal addresses #144, #119, and various other things we've been calling the time skew issue. It's basically a summary of what was discussed in the hangout today.

The main idea being that the epoch increases in lock step with monitoring, and all subsequent requests happen _in epoch_, which avoids the edge cases where a request may happen in an epoch that has yet to be verified.  Consequently, only monitoring updates the saved STR and the API for looking up in the latest epoch should be eliminated.

The first thing the client should do at startup is monitor, sending the epoch it knows about so that the server only needs to return STRs in the interval.  The server can also send the time until the next epoch, for the client to set a timer.  The client should be sure to check that the time is less than or equal to that attested to in the latest STR.

There's a bit of trickiness around the registration request (since it also needs to happen _in epoch_, as we said above). If the client is yet to register, the monitoring request at startup should be checking proof of absence. Since registration has to happen in the latest epoch, if the server receives a request to register in an outdated epoch, it should return an error, and give the client a chance to defer / re-request with exponential backoff.  Alternatively, the client can set a `monitoring` flag to `true` and prematurely queue other requests until that's done.

Thoughts?

c633:
> the API for looking up in the latest epoch should be eliminated.

What if the looking up key is in TBs queue? It happens when the registration and the key lookup are in the same epoch or when we support the key change operation. 

> The server can also send the time until the next epoch, for the client to set a timer.

You mean the expected time the next epoch will be published, right?

> Alternatively, the client can set a monitoring flag to true and prematurely queue other requests until that's done.

Queuing seems better to me. 

My main consideration is how can we put this flow into the library. I think most of these have to be done from the application level (but we can temporarily ignore this for now).

arlolra:
> What if the looking up key is in TBs queue?

Very good point; this needs some thought.  Initial ideas could be,

1. Return the TB if the epoch being looked up in happens to be the latest;
2. Keep a storage for TBs so that they can always be returned for an epoch.

> You mean the expected time the next epoch will be published, right?

Yes; well, it would be `time_to_next_epoch = time_of_next_epoch - current_time` so that what's returned can be put in a timer for when to perform the next monitoring operation.  `time_to_next_epoch` should be less than or equal to the `EpochDeadline` (as it's currently defined, an interval). Somewhat related is #81

c633:
[...]

PR #179, opened 2017-07-07
Merge `DirectoryProof` and `DirectoryProofs` into one

c633:
A series of pulls towards #173.

### DIFF
--- a/protocol/consistencychecks.go
+++ b/protocol/consistencychecks.go
@@ -76,12 +76,8 @@ func (cc *ConsistencyChecks) HandleResponse(requestType int, msg *Response,
 		return err.(ErrorCode)
 	}
 	switch requestType {
-	case RegistrationType, KeyLookupType:
+	case RegistrationType, KeyLookupType, KeyLookupInEpochType, MonitoringType:
 		if _, ok := msg.DirectoryResponse.(*DirectoryProof); !ok {
-			return ErrMalformedDirectoryMessage
-		}
-	case MonitoringType, KeyLookupInEpochType:
-		if _, ok := msg.DirectoryResponse.(*DirectoryProofs); !ok {
 			return ErrMalformedDirectoryMessage
 		}
 	default:
@@ -105,7 +101,7 @@ func (cc *ConsistencyChecks) updateSTR(requestType int, msg *Response) error {
 	var str *DirSTR
 	switch requestType {
 	case RegistrationType, KeyLookupType:
-		str = msg.DirectoryResponse.(*DirectoryProof).STR
+		str = msg.DirectoryResponse.(*DirectoryProof).STR[0]
 		// First response
 		if cc.SavedSTR == nil {
 			cc.SavedSTR = str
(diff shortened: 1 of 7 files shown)


protocol/message.go
```diff
@@ -176,8 +164,8 @@ func NewRegistrationProof(ap *m.AuthenticationPath, str *DirSTR,
 	return &Response{
 		Error: e,
 		DirectoryResponse: &DirectoryProof{
-			AP:  ap,
-			STR: str,
+			AP:  append([]*m.AuthenticationPath{}, ap),
+			STR: append([]*DirSTR{}, str),
```
masomel: We should update the documentation to specify that the length of `AP` and `STR` is expected to be equal to `1`. Same goes for `NewKeyLookupProof` below.

protocol/consistencychecks.go
```diff
@@ -199,8 +195,8 @@ func (cc *ConsistencyChecks) verifyRegistration(msg *Response,
 func (cc *ConsistencyChecks) verifyKeyLookup(msg *Response,
 	uname string, key []byte) error {
 	df := msg.DirectoryResponse.(*DirectoryProof)
-	ap := df.AP
-	str := df.STR
+	ap := df.AP[0]
+	str := df.STR[0]
```
masomel: I think we need to validate `df`at this point and fail early if `len(df.AP) != 1 || len(df.STR) != 1`. For now, we should do the same in `verifyRegistration` although this will probably change when we verify the prior history on a registration.

c633: It is validated in https://github.com/coniks-sys/coniks-go/blob/667a6e0367cf8f8fa028feb166b9137d7075e070/protocol/consistencychecks.go#L75

protocol/message.go
c633: Done.

protocol/consistencychecks.go
masomel: `message.validate()` only checks that the lengths are > 0. I was asking about an additional registration- and lookups-specific validation that checks the lengths are == 1.

c633: I got it. I think even a malicious server has no reasons to such an invalid response. I will add a TODO here, since I fixed it in the next pulls.

protocol/message.go
```diff
@@ -186,6 +175,7 @@ func NewRegistrationProof(ap *m.AuthenticationPath, str *DirSTR,
 // NewKeyLookupProof creates the response message a CONIKS directory
 // sends to a client upon a KeyLookupRequest,
 // and returns a Response containing a DirectoryProof struct.
+// The length of `AP` and `STR` must to be equal to 1.
```
arlolra: Why keep them as arrays then?

masomel: This is only the case for registrations and key lookups. Monitoring and `LookupInEpoch`s return more than one auth path and/or STR. And even registrations are going to return multiple auth paths and STRs when we implement prior history verification.

c633:
Can we merge this now?

c633:
Please open new issue if you have any further comments.

PR #182, opened 2017-08-18
Implement auditor-server protocol

masomel:
Part of #151 

TODO:

- [X] auditor-server request message
- [X] implement Audit() and change auditlog initialization
- [x] auditor-server tests
- [x] add auditor request handling in directory

### DIFF
--- a/protocol/auditlog.go
+++ b/protocol/auditlog.go
@@ -45,6 +45,15 @@ func (h *directoryHistory) updateVerifiedSTR(newVerified *DirSTR) {
 	h.snapshots[newVerified.Epoch] = newVerified
 }
 
+// insertRange inserts the given range of STRs snaps
+// into the directoryHistory h.
+// insertRange() assumes that snaps has been audited by Audit().
+func (h *directoryHistory) insertRange(snaps []*DirSTR) {
+	for i := 0; i < len(snaps); i++ {
+		h.updateVerifiedSTR(snaps[i])
+	}
+}
+
 // Audit checks that a directory's STR history
 // is linear and updates the auditor's state
 // if the checks pass.
@@ -53,10 +62,34 @@ func (h *directoryHistory) updateVerifiedSTR(newVerified *DirSTR) {
 // and then verifies the remaining STRs in msg, and
 // finally updates the snapshots if the checks pass.
 // Audit() is called when an auditor receives new STRs
-// from a directory.
+// from a specific directory.
 func (h *directoryHistory) Audit(msg *Response) error {
-	// TODO: Implement as part of the auditor-server protocol
-	return CheckPassed
+	if err := msg.validate(); err != nil {
+		return err.(ErrorCode)
+	}
+
+	strs := msg.DirectoryResponse.(*STRHistoryRange)
+
+	// Since the str[0] is pinned in the audit log
+	// expect that STR[0].Epoch is at least 1
+	if strs.STR[0].Epoch < 1 {
+		return ErrMalformedDirectoryMessage
+	}
+
+	// audit the STRs
+	// if strs.STR is somehow malformed or invalid (e.g. strs.STR
+	// contains old STRs), AuditDirectory() will detect this
+	// and throw and error
+	if err := h.AuditDirectory(strs.STR); err != nil {
+		return err
+	}
+
+	// TODO: we should be storing inconsistent STRs nonetheless
+	// so clients can detect inconsistencies -- or auditors
+	// should blow the whistle and not store the bad STRs
+	h.insertRange(strs.STR)
+
+	return nil
 }
 
 // NewAuditLog constructs a new ConiksAuditLog. It creates an empty
@@ -68,7 +101,8 @@ func NewAuditLog() ConiksAuditLog {
 
 // set associates the given directoryHistory with the directory identifier
 // (i.e. the hash of the initial STR) dirInitHash in the ConiksAuditLog.
-func (l ConiksAuditLog) set(dirInitHash [crypto.HashSizeByte]byte, dirHistory *directoryHistory) {
+func (l ConiksAuditLog) set(dirInitHash [crypto.HashSizeByte]byte,
+	dirHistory *directoryHistory) {
 	l[dirInitHash] = dirHistory
 }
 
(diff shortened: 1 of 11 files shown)


protocol/auditlog.go
```diff
@@ -53,10 +62,34 @@ func (h *directoryHistory) updateVerifiedSTR(newVerified *DirSTR) {
...
-	return CheckPassed
+	if err := msg.validate(); err != nil {
+		return err.(ErrorCode)
+	}
+
+	strs := msg.DirectoryResponse.(*STRHistoryRange)
+
+	// Since the str[0] is pinned in the audit log
+	// expect that STR[0].Epoch is at least 1
+	if strs.STR[0].Epoch < 1 {
```
c633: Wouldn't this be covered in `AuditDirectory()`? And I suppose an auditor should be able to request an STR range starting from epoch 0.

protocol/auditor.go
```diff
@@ -105,7 +106,7 @@ func (a *AudState) checkSTRAgainstVerified(str *DirSTR) error {
 		return CheckBadSTR
 	}
 
-	return nil
```
c633: I prefer to return nil here.

protocol/auditlog.go
masomel: You're right, this case is covered in the `str.Epoch == verifiedSTR.Epoch` case in `AuditDirectory()`. What I wanted to avoid is re-inserting `str[0]`, but if the checks pass, it wouldn't matter anyway.

protocol/auditor.go
masomel: Do you mean only for this function because it's only a helper function, or in all the cases in which I replaced `nil` with `CheckPassed`? I feel like it would definitely make sense to return `CheckPassed` from `AuditDirectory()` at least.

c633: I mean only for this function ;)

protocol/auditlog.go
```diff
@@ -45,18 +45,45 @@ func (h *directoryHistory) updateVerifiedSTR(newVerified *DirSTR) {
 	h.snapshots[newVerified.Epoch] = newVerified
 }
 
+// insertRange inserts the given range of STRs snaps
+// into the directoryHistory h.
+// insertRange() assumes that snaps has been audited by Audit().
+func (h *directoryHistory) insertRange(snaps []*DirSTR) {
+	for i := 0; i < len(snaps); i++ {
```
c633: Why not `h.updateVerifiedSTR(snaps[lens(snaps)-1]`?

protocol/auditlog.go
```diff
@@ -45,18 +45,45 @@ func (h *directoryHistory) updateVerifiedSTR(newVerified *DirSTR) {
...
+		h.updateVerifiedSTR(snaps[i])
+	}
+}
+
 // Audit checks that a directory's STR history
-// is linear and updates the auditor's state
+// is linear and updates the audtor's state
```
c633: typo

protocol/auditlog.go
```diff
@@ -45,18 +45,45 @@ func (h *directoryHistory) updateVerifiedSTR(newVerified *DirSTR) {
...
 // and then verifies the remaining STRs in msg, and
 // finally updates the snapshots if the checks pass.
 // Audit() is called when an auditor receives new STRs
-// from a directory.
+// from a specific directory.
 func (h *directoryHistory) Audit(msg *Response) error {
-	// TODO: Implement as part of the auditor-server protocol
-	return CheckPassed
+	if err := msg.validate(); err != nil {
+		return err.(ErrorCode)
```
c633: It could be just `return err`.

protocol/auditlog.go
```diff
@@ -81,24 +109,21 @@ func (l ConiksAuditLog) get(dirInitHash [crypto.HashSizeByte]byte) (*directoryHi
...
-// FIXME: pass Response message as param
-// masomel: will probably want to write a more generic function
-// for "catching up" on a history in case an auditor misses epochs
-func (l ConiksAuditLog) Insert(addr string, signKey sign.PublicKey,
+// InitHistory() is called by an auditor when it initializes its state
+// from disk (either first-time startup, or after reboot).
+func (l ConiksAuditLog) InitHistory(addr string, signKey sign.PublicKey,
 	snaps []*DirSTR) error {
 	// make sure we're getting an initial STR at the very least
 	if len(snaps) < 1 || snaps[0].Epoch != 0 {
```
c633: I think we should compare `snaps[0]` with the pinning STR, no?

protocol/consistencychecks.go
```diff
@@ -133,7 +133,7 @@ func (cc *ConsistencyChecks) updateSTR(requestType int, msg *Response) error {
 		// The initial STR is pinned in the client
 		// so cc.verifiedSTR should never be nil
 		// FIXME: use STR slice from Response msg
-		if err := cc.AuditDirectory([]*DirSTR{str}); err != nil {
+		if err := cc.AuditDirectory([]*DirSTR{str}); err != CheckPassed {
```
c633: I will remove this `CheckPassed` later (`nil` is more idiomatic Go), so I think we can leave it as it was.

protocol/directory.go
```diff
@@ -321,3 +321,44 @@ func (d *ConiksDirectory) Monitor(req *MonitoringRequest) (
...
+// latest epoch of this directory, or a start epoch greater than the
+// end epoch (if included) is considered malformed, and causes
+// GetSTRHistory() to return a
+// message.NewErrorResponse(ErrMalformedAuditorMessage) tuple.
+// GetSTRHistory() returns a message.NewSTRHistoryRange(strs) tuple.
+// strs is a list of STRs for
+// the epoch range [startEpoch, endEpoch], where startEpoch
+// and endEpoch are the epoch range endpoints indicated in the client's
+// request. If req.endEpoch is greater than d.LatestSTR().Epoch or
+// omitted in req, the end of the range will be set to d.LatestSTR().Epoch.
```
c633: We haven't checked if req.endEpoch is omitted in "MonitoringRequest", so should we change the range request to something like `[Epoch, Range]` where `Epoch` is `StartEpoch` and `Range` is the maximum number of epochs to return?

protocol/message.go
```diff
@@ -135,7 +149,8 @@ type DirectoryProof struct {
 // STR representing a range of the STR hash chain. If the range only
 // covers the latest epoch, the list only contains a single STR.
 // A CONIKS auditor returns this DirectoryResponse type upon an
-// AudutingRequest.
+// AudutingRequest from a client, and a CONIKS directory returns
```
c633: typo

protocol/auditlog.go
masomel: `snaps[0]` *is* the pinned STR. As the documentation describes, this function is called when the auditor first starts up, so the very first time that auditor is launched, it will read the pinned STR from disk and pass it into `InitHistory` through the `snaps` parameter.

protocol/consistencychecks.go
masomel: I think for CONIKS, though, `CheckPassed` is a clearer return value than nil.

protocol/auditlog.go
c633: Ahh sorry, I missed that part. Thanks!

masomel: No problem! I can make this clearer in the documentation.

PR #183, opened 2017-08-22
/protocol organizational refactoring

c633:
Merge #170 & #182 first.


protocol/directory/directory.go
```diff
@@ -6,14 +6,15 @@
...
-package protocol
+package directory
 
 import (
 	"bytes"
 
 	"github.com/coniks-sys/coniks-go/crypto/sign"
 	"github.com/coniks-sys/coniks-go/crypto/vrf"
 	"github.com/coniks-sys/coniks-go/merkletree"
+	p "github.com/coniks-sys/coniks-go/protocol"
```
masomel: Nit pick: We should probably be consistent with how we import some other coniks-go packages. We always import `protocol` as `p` which I think is fine, but `merkletree` sometimes is imported as `m` (e.g. https://github.com/coniks-sys/coniks-go/blob/master/protocol/consistencychecks.go#L13), and other times (like here) not. I think I've seen this with `directory` as well (`dir` vs the full name).

masomel:
@c633 I went ahead and moved around the new test cases I added in #182 to make the tests pass again. There are, however, two tests in consistency_checks.go that no longer pass (`TestVerifyRegistrationWithTB` and `TestVerifyFulfilledPromise`). I don't know if you encountered this issue before.

protocol/directory/directory.go
c633: Ok, I am fixing this.

protocol/client/consistencychecks.go
```diff
@@ -2,15 +2,16 @@
 // on data received from a CONIKS directory.
 // These include data binding proof verification,
 // and non-equivocation checks.
-// TODO: move all STR-verifying functionality to a separate module
 
-package protocol
+package client
```
c633: @masomel @Liamsi @arlolra I would like to have your inputs about this name. It is collided with our application client package (see https://github.com/coniks-sys/coniks-go/pull/183/files#diff-50a5a7c3748e6d839ee365d7f4291a1cR13). A name I think of is `verifier`. What do you think?

c633:
Our error codes are a mess for now. Another PR will follow up.

c633:
All tests are fixed. PTAL.

protocol/client/consistencychecks.go
masomel: Hmm. I think the auditor can also be considered a type of `verifier`, so I don't think this is the best name for this package. 

How about we change the client CLI package name? (Btw, we'll have to find a name other than `auditor` for the auditor CLI, too). Can we call it something like `client-bin` or `client-app`?

c633:
> Are we going to go back to using import . for the test cases that used to have cyclical dependencies?

I think yes.

protocol/client/consistencychecks.go
c633: What do you think about https://github.com/coniks-sys/coniks-go/pull/183/commits/425366425260796134bdd9baef82e6fbc6933602?

masomel:
In the spirit of refactoring: Is there an application layer API that we can refactor from the CLIs? I haven't spent much time looking into this, but it could be useful to have an application layer API for each component as well. This could then help developers integrate things like a client with their existing applications. What do you think?

c633:
> Is there an application layer API that we can refactor from the CLIs?

Yes, it is. Actually, every module that doesn't belong to `cli` package should be used to provide APIs for developers. We do have plan for this, please see #101, I think it belongs to milestone 0.3 and 0.5.

PR #191, opened 2017-10-25
Refactor testings code

c633:
Part of #18: 

- Remove some redundant tests in /protocol level. 

- Use deterministic tests.

- Refactor tests in /merkletree based on work of @Liamsi 

Note:

- Tests in `auditlog_test.go` and `consistencychecks_test.go` should be rewritten and integrated with the directory tests, after our APIs are complete.


protocol/auditor/common_test.go
```diff
@@ -35,9 +32,18 @@ func TestComputeDirectoryIdentity(t *testing.T) {
...
-				t.Errorf("ComputeDirectoryIdentity() = %v, want %v", got, tt.want)
+			if got, want := ComputeDirectoryIdentity(tt.str), tt.want; !bytes.Equal(got[:], want) {
+				t.Errorf("ComputeDirectoryIdentity() = %#v, want %#v", got, want)
 			}
 		})
 	}
 }
+
+// decode hex string to byte array
+func dh(h string) []byte {
```
masomel: This is another one of those functions that could be exposed in a `testutil` module. I would also change the name from `dh` to something more meaningful.

protocol/directory/directory_test.go
```diff
@@ -290,25 +33,74 @@ func TestPoliciesChanges(t *testing.T) {
...
-	d, _ := NewTestDirectory(t, true)
-
-	d.Update()
-
-	res := d.GetSTRHistory(&protocol.STRHistoryRequest{
-		StartEpoch: uint64(4),
-		EndEpoch:   uint64(2)})
+func TestDirectoryKeyLookupInEpochBadEpoch(t *testing.T) {
+	d := NewTestDirectory(t)
+	tests := []struct {
```
masomel: See my comment above about the `mockTest` struct.

protocol/directory/directory_test.go
```diff
@@ -290,25 +33,74 @@ func TestPoliciesChanges(t *testing.T) {
...
+		name    string
+		startEp uint64
+		endEp   uint64
+		want    error
+	}{
+		{"bad end epoch", 4, 2, protocol.ErrMalformedMessage},
+		{"out-of-bounds", 6, d.LatestSTR().Epoch, protocol.ErrMalformedMessage},
+	}
+	for _, tt := range tests {
```
masomel: I like this structure for our error condition testing. Since so many of our tests follow this structure, is it possible to further refactor this into a generic function that we can use everywhere instead of having to copy/paste this code everywhere (this will make writing future tests easier, too). Here's what I have in mind: We have a generic `TestError` function:

```
func TestError(t *testing.T, tests, testFunction, testRequestType) {
       for _, tt := range tests {
		res := testFunction(newTestRequest(testRequestType, tt))
		if res.Error != tt.want {
			t.Errorf("Expect %s for %s", tt.error, tt.name)
	       }
 	}
 }
```

Which then gets called as `TestError(t, tests, d.GetSTRHistory, protocol.STRType)` in `TestBadRequestGetSTRHistory`. Writing new tests is then only a matter of filling the appropriate test parameter struct and calling `TestError` with the right parameters. What do you think?

crypto/testutil.go
```diff
@@ -0,0 +1,29 @@
...
+	sk, err := vrf.GenerateKey(bytes.NewReader(
+		[]byte("deterministic tests need 256 bit")))
+	if err != nil {
+		t.Fatal(err)
+	}
+	return sk
+}
+
+// StaticSigning returns a static private signing key for _tests_.
+func StaticSigning(t *testing.T) sign.PrivateKey {
```
masomel: I think we should use consistent naming and include the word `Test` in these function names just to make it absolutely clear to the reader that these are used for testing.

merkletree/testutil.go
```diff
@@ -0,0 +1,39 @@
...
+	"testing"
+
+	"github.com/coniks-sys/coniks-go/crypto"
+)
+
+var keyPrefix = "key"
+var valuePrefix = []byte("value")
+
+// StaticPAD returns a pad with a static initial STR for _tests_.
+func StaticPAD(t *testing.T, ad AssocData) *PAD {
```
masomel: See my comment above regarding `crypto.StaticVRF` and such.

protocol/tests/client_auditlog_test.go
```diff
@@ -0,0 +1 @@
+package tests
```
masomel: I'm assuming you plan to move tests from `auditlog_test` etc into this module later on?

c633: Yes, for instance, please see https://github.com/coniks-sys/coniks-go/pull/192/files#diff-4b27042e76fbd702dc90612e2d9e9372.

protocol/directory/directory_test.go
c633: I'm not sure about this since I think it doesn't really save us much. Besides we can _sometimes_ run sub-tests in parallel (see https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks) using `t.Run`.

merkletree/testutil.go
```diff
@@ -0,0 +1,39 @@
...
+func staticTree(t *testing.T) *MerkleTree {
+	m, err := NewMerkleTree()
+	if err != nil {
+		t.Fatal(err)
+	}
+	m.nonce = []byte{}
+	return m
+}
+
+func newTestTree(t *testing.T) *MerkleTree {
```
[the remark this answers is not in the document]
c633: It means it is intentionally used for tests only (people can freely call this function within the package - the parameter `t` here is to prevent this.)

protocol/auditor/auditor.go
```diff
@@ -91,10 +91,6 @@ func (a *AudState) CheckSTRAgainstVerified(str *protocol.DirSTR) error {
 	// Maybe it has something to do w/ #81 and client
 	// transitioning between epochs.
 	// Try to verify w/ what's been saved
-
-	// FIXME: we are returning the error immediately
```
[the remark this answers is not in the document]
c633: We don't do any "saving" action here, we do it in the callers (e.g., https://github.com/coniks-sys/coniks-go/blob/master/protocol/auditlog/auditlog.go#L85-L87)

merkletree/proof_test.go
```diff
@@ -2,192 +2,145 @@ package merkletree
...
-	if err := m.Set(index1, key1, val1); err != nil {
-		t.Fatal(err)
-	}
+var N uint64 = 3 // number of inclusions.
+
+func setup(t *testing.T) (*MerkleTree, []*mockTest) {
+	m := newTestTree(t)
+
+	tuple := []*mockTest{}
+	for i := uint64(0); i < uint64(N); i++ {
```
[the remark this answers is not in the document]
c633: Insertion and lookups depends on test cases, so IMO we don't really need to refactor them into smaller functions.

merkletree/proof_test.go
```diff
@@ -2,192 +2,145 @@ package merkletree
...
-	if ap2.Leaf.Value == nil {
-		t.Error("Cannot find key:", key2)
-		return
-	}
-	ap3 := m.Get(index3)
-	if ap3.Leaf.Value == nil {
-		t.Error("Cannot find key:", key3)
-		return
-	}
+type mockTest struct {
```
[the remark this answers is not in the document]
c633: Same as https://github.com/coniks-sys/coniks-go/pull/191#discussion_r147870279, `mockTest` data structure depends on test cases. But yes, I should find a better name.

c633:
Thanks for your first round of reviews! I will push documentation/naming changes tomorrow.

merkletree/proof_test.go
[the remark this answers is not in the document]
masomel: That makes sense. But there still is a lot of repeated test structure, and that's what I'm suggesting we refactor. It will make the tests cleaner and a lot easier to write new ones. I think it would be more valuable and helpful to us and other developers in the future if we could abstract away most of the internals of testing and just had to pass in the test cases we want to test. If it doesn't take too much effort, we should take care of that in this PR.

protocol/directory/directory_test.go
masomel: See my comment https://github.com/coniks-sys/coniks-go/pull/191#discussion_r148081877. If we have cases in which we run parallel sub-tests, I think we could still write two separate functions, one for parallel testing and sequential testing, but still abstract away all the common testing internals.

merkletree/proof_test.go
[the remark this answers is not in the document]
masomel: Right, each test case has its own data structure, but it looks like the data structure has some common fields in all test cases, and what changes really only depends on the message type/protocol you're testing. What I was trying to point out is that you created a named data structure for the lookup tests, but not for all other tests, and I think it would be useful to explicitly have named test parameter structures for all test cases.

crypto/testutil.go
c633: I still prefer to keep this name since it reflects exactly what the function does. Besides, the docstring states it clearly and the parameter requires `testing.T` so I don't think this will cause any trouble.

masomel: My point was a bit more about consistency in naming throughout all of our tests, we should agree on a convention. You're definitely right, if the docstring states the test clearly, the name doesn't need to include `Test`.

c633: `Test` is used as prefix only for testcases in `_test.go` files, so I think this doesn't violate our convention ;)

masomel: This is true for the test cases. At the same time, for this function I would have expected a name more like `NewTestSigningKey` since most other similar `testutil` helpers follow the `NewTest* convention.

c633: What do you think about https://github.com/coniks-sys/coniks-go/pull/191/commits/e06784bb925b4e2b560af6f80a77dcf3db9a4e83?

merkletree/pad_test.go
```diff
@@ -407,22 +366,31 @@ func benchPADLookup(b *testing.B, entries uint64) {
 // The STR will get updated every epoch defined by every multiple of
 // `updateEvery`. If `updateEvery > N` createPAD won't update the STR.
 func createPad(N uint64, keyPrefix string, valuePrefix []byte, snapLen uint64,
-	updateEvery uint64) (*PAD, error) {
-	pad, err := NewPAD(TestAd{""}, signKey, vrfPrivKey1, snapLen)
+	afterCreateCB func(pad *PAD),
```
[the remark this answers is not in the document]
c633: done.

merkletree/proof_test.go
```diff
@@ -2,192 +2,145 @@ package merkletree
...
-
-	key1 := "key1"
-	index1 := vrfPrivKey1.Compute([]byte(key1))
-	val1 := []byte("value1")
-	if err := m.Set(index1, key1, val1); err != nil {
-		t.Fatal(err)
-	}
+var N uint64 = 3 // number of inclusions.
+
+func setup(t *testing.T) (*MerkleTree, []*mockTest) {
```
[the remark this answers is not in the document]
c633: Rename to `setupTestProofs`.

merkletree/testutil.go
c633: done.

protocol/auditor/common_test.go
c633: Renamed to `hex2bin`.

crypto/testutil.go
masomel: This definitely works. Thanks!

c633:
This is ready for another review. @masomel

c633:
Merged in c9491a40924c73f26f600e945d0062ce9b3650a1.

PR #193, opened 2017-11-02
[WIP] Implement the auditor-server application layer

masomel:
Part of #151


utils/binutils/encoding.go
```diff
@@ -0,0 +1,84 @@
...
+			Error:             res.Error,
+			DirectoryResponse: response,
+		}
+	default:
+		panic("Unknown request type")
+	}
+}
+
+// MarshalSTRToFile serializes the given STR to the given path.
+func MarshalSTRToFile(str *protocol.DirSTR, path string) {
```
c633: Related: https://github.com/coniks-sys/coniks-go/pull/137#issuecomment-264236460

masomel: Definitely. This particular function is just for the test implementation, but in implementing the auditor, having a persistent storage backend is becoming more and more necessary. We should get back to implementing that soon.

utils/binutils/logger.go
```diff
@@ -0,0 +1,134 @@
+package binutils
```
c633: We already have `application/logger.go`.

masomel: Ugh, thanks for catching that. I thought I removed the `binutils` package when I rebased.

application/encoding.go
```diff
@@ -77,8 +77,12 @@ func UnmarshalResponse(t int, msg []byte) *protocol.Response {
 			Error: res.Error,
 		}
 		if err := response.Validate(); err != nil {
-			return &protocol.Response{
-				Error: protocol.ErrMalformedMessage,
+			// we don't want to return an ErrMalformedMessage
```
c633: I think this is covered in https://github.com/coniks-sys/coniks-go/blob/master/protocol/message.go#L272

masomel: Yes, but the problem is that *this* function was still returning an `ErrMalformedMessage` even if the error is in `errors`. In other words, because `Validate()` in message.go returns `msg.Error`, which gives `err == nil` after `Validate()` returns, the `return &protocol.Response{Error: protocol.ErrMalformedMessage }` statement was always being returned, even when the `msg.Error` was in `errors`.

c633: I see. I gave it a try in https://github.com/coniks-sys/coniks-go/compare/unmarshalling. If it's ok, feel free to merge to this branch.

masomel: Those changes looks good to me. Thank you! I also like that you re-wrote the test cases. I'll merge this branch when I get a chance.

Issue #194, opened 2017-11-09
go install of CLIs results in `cli` binary, not respective binary name


masomel:
I've set `GOBIN=$GOPATH/bin`. If I run `go install github.com/coniks-sys/coniks-go/coniksserver/cli`, I end up with the `cli` binary, which will conflict with the other CLIs. If I run `go install github.com/coniks-sys/coniks-go/coniksserver/cli/coniksserver.go`, I get a "No such file or directory" error unless I run the command from `$GOPATH/src`, which isn't ideal.

We should either rename the `coniks*/cli` packages to reflect their binary names, or refactor the `coniks*` packages so we can use the `cli` sub-packages as top-level packages.

c633:
Another inelegant solution is to go with `git clone ... && make`.

masomel:
Right, I considered that, but like you said, that's an inelegant solution :)

c633:
It looks great!

c633:
It makes sense, but I think we could do further refactoring:
- Move `cli` packages to `/cli/{auditor,server,bot,client}`, we still can `go install github.com/coniks-sys/coniks-go/cli/{auditor, server, bot, client}` with expected name.
- Move `coniks*` packages (e.g., `/coniksserver/{server.go, listener.go}` to `impl/{auditor, server, bot, client}` so we can reuse that code to build the C library or whatever. (I borrowed the name `impl` from KT).

masomel:
This sounds good to me. I think it makes sense to solve this issue before I finish #193, I can give this a shot since I've already started refactoring a bit.

PR #195, opened 2017-11-13
Refactor CLI packages

masomel:
Addresses #194 

This PR moves the cli binaries to `/cli` and refactors some common functionality used to implement various CONIKS executables or client applications in the `/application` package. 

- [X] refactor encoding
- [x] refactor common server functionality
- [x] refactor server listener
- [x] move logger from utils to application
- [x] refactor bot
- [x] refactor common CLI cmds
- [X]  refactor configs
- [x] refactor CLI tools


Liamsi:
Closing and reopening to trigger travis ...

application/config.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/BurntSushi/toml"
+	"github.com/coniks-sys/coniks-go/crypto/sign"
+	"github.com/coniks-sys/coniks-go/utils"
+)
+
+// AppConfig is the generic type used to specify the configuration of
+// any kind of CONIKS application-level executable (e.g. key server,
+// client etc.).
+type AppConfig interface {
+	InitConfig(file string) error
```
c633: Could be just `Init`, or `Load`. I think the latter is more precise.

application/bots/twitterbot.go
```diff
@@ -66,16 +43,15 @@ type TwitterOAuth struct {
 // tuple. Otherwise, it returns a TwitterBot struct
 // with the appropriate values obtained during the setup.
 func NewTwitterBot(path string) (Bot, error) {
-	var conf TwitterConfig
-	if _, err := toml.DecodeFile(path, &conf); err != nil {
-		return nil, fmt.Errorf("Failed to load config: %v", err)
+	var conf *TwitterConfig = &TwitterConfig{}
+	if err := conf.InitConfig(path); err != nil {
```
c633: Please see https://github.com/coniks-sys/coniks-go/pull/195/files#r154509882. Then this could be
```
var config *TwitterConfig = &TwitterConfig{}
config.Load(path)
```

application/config.go
masomel: Hmm, I thought of this. There already is a `Load` function in `application/config`, so I didn't want to have double-naming to reduce confusion. I also decided to include the `Config` for readability of the code: `server.LoadConfig()` vs `server.Load()`; I find that the former is a bit more specific. What do you think?

masomel:
@c633 When you get a chance, can you please review?

application/server/server.go
```diff
@@ -0,0 +1,174 @@
...
+
+	return protocol.NewErrorResponse(protocol.ErrMalformedMessage)
+}
+
+// Run implements the main functionality of the key server.
+// It listens for all declared connections with corresponding
+// permissions.
+func (server *ConiksServer) Run(addrs []*Address) {
+	server.WaitStopAdd()
+	go server.EpochUpdate()
```
c633: I think we could do
```
server.WaitStopAdd()
go func() {
  server.EpochUpdate()
  server.WaitStopDone()
}
```
It looks more idiomatic Go. 

That being said, I would love to find another way to refactor these APIs.

c633:
I love what you did with `cli` packages. I have only a concern regarding the `ServerBase` exported APIs.

application/server/server.go
masomel: I agree, I'm not 100% happy with these APIs. If you have an alternative suggestion, please let me know! We might find a better way to express this API when implementing the auditor.

masomel: According to places like https://tour.golang.org/concurrency/1, `go server.EpochUpdate()` is just as idiomatic in Go. Is there a reason why these goroutines should be in anonymous functions?

c633: Calling for help @Liamsi  @arlolra ...

PR #203, opened 2018-01-14
Fix application.LoadConfig bug

### LINKED ISSUE #200: application.LoadConfig() doesn't work
Steps to Reproduce:
1. `coniksserver init`
2. `coniksserver run`

Error: `Failed to load config: toml: unsupported type application.AppConfig`

Note that `toml` only supports empty interfaces.

masomel:
Fixes #200

### DIFF
--- a/application/serverbase.go
+++ b/application/serverbase.go
@@ -19,9 +19,19 @@ import (
 // which are read at initialization time from
 // a TOML format configuration file.
 type ServerBaseConfig struct {
-	Logger         *LoggerConfig      `toml:"logger"`
-	ConfigFilePath string             `toml:"config_file_path"`
+	*CommonConfig
 	EpochDeadline  protocol.Timestamp `toml:"epoch_deadline"`
+}
+
+// NewServerBaseConfig initializes a server's config file path,
+// its loader for the given config encoding, the logger configuration,
+// and its epoch deadline.
+func NewServerBaseConfig(file, encoding string, logger *LoggerConfig,
+	epDeadline protocol.Timestamp) *ServerBaseConfig {
+	return &ServerBaseConfig{
+		CommonConfig: NewCommonConfig(file, encoding, logger),
+		EpochDeadline: epDeadline,
+	}
 }
 
 // EpochTimer consists of a `time.Timer` and the epoch deadline value.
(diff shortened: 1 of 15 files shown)


c633:
I gave this a try in https://github.com/coniks-sys/coniks-go/compare/gh200?expand=1, separated into 2 commits. Feel free to merge `gh200` if you think it's okay.

masomel:
What you did definitely works, but it seems a bit more complex than it needs to be, IMO. I definitely like the idea of having a common configuration abstraction in `ConfigService`, but I don't really like the seemingly cyclical use of the use of `AppConfig`. What I mean is the fact that the `ConfigService` struct includes an `AppConfig field`, and then each of the component `Config` structs "extends" a `ConfigService` *and* also implements the `AppConfig` interface. I'll see how I can include your changes, but I still think we don't need the `AppConfig` interface anymore.

c633:
OK, it's a fair point.

application/serverbase.go
```diff
@@ -30,7 +30,7 @@ type Server interface {
 // a TOML format configuration file.
 type ServerBaseConfig struct {
 	Logger         *LoggerConfig `toml:"logger"`
-	ConfigFilePath string
+	ConfigFilePath string        `toml:"path"`
```
c633: This could be renamed to `Path` for short.

masomel:
@huy, I ended up re-introducing the `AppConfg` interface as you suggested, but I created a separate `ConfigLoader` interface that allows us to support different config encodings. Please let me know what you think.

c633:
For now, I think introducing a new `ConfigLoader` interface makes things more complicated than necessary, since we've already hardcoded `toml` encoding everywhere and nothing else. It seems like `ServerBaseConfig` is used nowhere, maybe it can be safely removed?

masomel:
> `ConfigLoader` interface makes things more complicated than necessary, ...

I don't think the `ConfigLoader` really makes things much more complicated for developers, but I do agree that it's not really necessary at the moment, especially since we don't currently support other config encodings. But I think it's great to have for future extensibility. 

> ...since we've already hardcoded `toml` encoding everywhere and nothing else.

I realized this, too. I was wondering if those hardcoded struct tags are really necessary? Much like we did for our message encoding in the protocol layer where we removed the JSON tags, we could remove the toml tags in the config encodings?

> It seems like `ServerBaseConfig` is used nowhere, maybe it can be safely removed?

Yep, especially now that we want to remove `EpochDeadline` from `SeverBaseConfig` we can remove it.

c633:
Great! LGTM!

Issue #196, opened 2017-11-25
error when trying to register the name-key pair


chesnokovilya:
I am running server from master branch on local machine(CentOs 7):
```
[ilya@localhost]~/coniks% ~/go/bin/coniksserver run -p
2017-11-25T10:02:17.072+0700	INFO	Accepting registrations	{"address": "tcp://127.0.0.1:3000"}
```
 
When I try to register name-key pair I got the following error:

```[ilya@localhost]~/coniks-client% ~/go/bin/coniksclient run              
coniks-client> register alice MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQCqGKukO1De7z
hZj6+H0qtjTkVxwTCpvKe4eCZ0
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x70b6b2]

goroutine 1 [running]:
github.com/coniks-sys/coniks-go/protocol/auditor.(*AudState).CheckSTRAgainstVerified(0xc4200cec20, 0xc4200f8720, 0x5778d9b2ba91b117, 0x570000c4200fc840)
	/home/ilya/go/src/github.com/coniks-sys/coniks-go/protocol/auditor/auditor.go:99 +0x32
github.com/coniks-sys/coniks-go/protocol/auditor.(*AudState).AuditDirectory(0xc4200cec20, 0xc4200f79c8, 0x1, 0x1, 0x7f47658a8000, 0x0)
	/home/ilya/go/src/github.com/coniks-sys/coniks-go/protocol/auditor/auditor.go:152 +0x4a
github.com/coniks-sys/coniks-go/protocol/client.(*ConsistencyChecks).updateSTR(0xc4200cec40, 0x0, 0xc4200fc840, 0xc4200150e0, 0x50)
	/home/ilya/go/src/github.com/coniks-sys/coniks-go/protocol/client/consistencychecks.go:137 +0xa1
github.com/coniks-sys/coniks-go/protocol/client.(*ConsistencyChecks).HandleResponse(0xc4200cec40, 0x0, 0xc4200fc840, 0xc420070489, 0x5, 0xc4200150e0, 0x4c, 0x50, 0x0, 0x0)
	/home/ilya/go/src/github.com/coniks-sys/coniks-go/protocol/client/consistencychecks.go:115 +0x91
[...]

PR #204, opened 2018-01-15
Pin key server initial STR, load in client

masomel:
Addresses #196 

Mechanism:
* Add STR serialization/de-serialization functions
* Serialize the key server's initial STR to disk upon startup
* Add initial STR to client config
* Client is initialized with initial STR specified in config

~~Blocked by #203~~

### DIFF
--- a/application/config.go
+++ b/application/config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 
+	"github.com/BurntSushi/toml"
 	"github.com/coniks-sys/coniks-go/crypto/sign"
 	"github.com/coniks-sys/coniks-go/protocol"
 	"github.com/coniks-sys/coniks-go/utils"
--- a/application/config.go
+++ b/application/config.go
@@ -2,11 +2,13 @@ package application
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 
 	"github.com/BurntSushi/toml"
 	"github.com/coniks-sys/coniks-go/crypto/sign"
+	"github.com/coniks-sys/coniks-go/protocol"
 	"github.com/coniks-sys/coniks-go/utils"
 )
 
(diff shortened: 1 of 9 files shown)


application/config.go
```diff
@@ -1,10 +1,13 @@
 package application
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
```
c633: missing "github.com/BurntSushi/toml"

application/server/server.go
```diff
@@ -64,6 +65,11 @@ func NewConiksServer(conf *Config) *ConiksServer {
 		epochTimer: application.NewEpochTimer(conf.EpochDeadline),
 	}
 
+	// save the initial STR to be used for initializing auditors
+	initSTRPath := utils.ResolvePath(conf.InitSTRPath,
+		conf.ConfigFilePath)
```
c633: error: `conf.ConfigFilePath undefined`

cli/coniksserver/internal/cmd/init.go
```diff
@@ -66,7 +66,9 @@ func mkConfig(dir string) {
 		SignKeyPath:   "sign.priv",
 	}
 
-	conf := server.NewConfig(file, "toml", addrs, logger, 1000000, policies)
+	conf := server.NewConfig(file, "toml", addrs, logger, 1000000, policies,
+		"init.str")
```
c633: Could we put `init.str` to the config file?

cli/coniksclient/internal/cmd/run.go
```diff
@@ -42,7 +42,10 @@ func init() {
 func run(cmd *cobra.Command, args []string) {
 	isDebugging, _ := strconv.ParseBool(cmd.Flag("debug").Value.String())
 	conf := loadConfigOrExit(cmd)
-	cc := client.New(nil, true, conf.SigningPubKey)
+
+	// FIXME: right now we're passing the initSTR, but we should really
```
c633: Instead of putting a FIXME here, I would like to put it in https://github.com/coniks-sys/coniks-go/blob/master/protocol/directory/directory.go#L70, where we should also store the tree etc...

application/server/server.go
c633: I would like to put a FIXME here: this saving should happen in `protocol/` level (i.e., when the server starts and updates)

c633:
Please make travis happy. I will try to rewrite the `config_encoding.go` after this gets merged.

masomel:
> Please make travis happy

Oops, I only saw the issue with zap in Go 1.8 (see https://travis-ci.org/coniks-sys/coniks-go/jobs/401579735#L577).

PR #206, opened 2018-02-03
Run dep ensure

corbym:
You can run dep ensure in travis by running `go get -u github.com/golang/dep/cmd/dep` before `dep ensure`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ go:
   - 1.9
   - tip
 
-# TODO: Run `dep ensure` in Travis
-# see: https://github.com/coniks-sys/coniks-go/pull/201
 before_install:
   - go get github.com/mattn/goveralls
   - go get golang.org/x/tools/cmd/cover  
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,9 @@ go:
 # see: https://github.com/coniks-sys/coniks-go/pull/201
 before_install:
   - go get github.com/mattn/goveralls
-  - go get golang.org/x/tools/cmd/cover
+  - go get golang.org/x/tools/cmd/cover  
+  - go get -u github.com/golang/dep/cmd/dep
+  - dep ensure
 
 env:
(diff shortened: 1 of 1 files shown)


corbym:
Looks like running dep ensure might be causing some tests to fail in travis on the Go 1.8 build? 

Works on 1,9 and tip

c633:
Thanks @corbym for this PR!

masomel:
Based on the travis logs, `go get` fails for a few external dependencies.

corbym:
It's possible you may need to branch for 1.8. it may be that the dependencies are no longer supporting 1.8.

The Twitter dependency is on master, for example. You won't be able to support master and a 1.8 compatible version of the dep.

c633:
So I tried to run `dep ensure` locally with go 1.8, and there was no issue as reported in Travis. You can try yourself with Docker:
```
# Dockerfile
FROM golang:1.8
WORKDIR /go/src/app
COPY . .
RUN go get -u github.com/golang/dep/cmd/dep
RUN dep ensure
```
I would wait for the "official" `dep` support of Travis.

mattcorbyeaglen:
... but it's your build that's failing, not deep ensure.

masomel:
@corbym The odd thing is that our build does not fail in Travis for go1.8 on any of our other branches, so it would seem that our build failure *is* somehow related to introducing `dep` into the Travis build flow. 

You're definitely right, though, we should be running the build locally after running `dep ensure`. Will investigate further.

corbym:
@masomel if I get time I will try to build your project under 1.8 in a docker container and run dep ensure in the same manner as the build tries to.

Sorry about the account switch above btw, I have two github accounts and sometimes I forget which is logged in!

corbym:
@c633 @masomel looks like you have mystery on your hands.. I'll leave it to you guys if that's ok?

masomel:
Thanks again @corbym for bringing this to our attention :)

